PR #14, opened 2014-12-02
Added XXH128 and XXH256

JCash:
Both XXH128/XXH256 are implemented in the spirit of XXH64, with minor tweaks to constants and the final mixin (to pass smhasher tests). They both pass all smhasher tests. Speedwise, they have the same performance as XXH64 for longer keys, and for shorter keys, they are slightly slower, due to more instructions for the final mix.

### DIFF
--- a/xxhash.c
+++ b/xxhash.c
@@ -72,11 +72,11 @@ You can contact the author at :
 #  define FORCE_INLINE static __forceinline
 #else
 #  if defined (__STDC_VERSION__) && __STDC_VERSION__ >= 199901L   /* C99 */
-#  ifdef __GNUC__
-#    define FORCE_INLINE static inline __attribute__((always_inline))
-#  else
-#    define FORCE_INLINE static inline
-#  endif
+#    ifdef __GNUC__
+#      define FORCE_INLINE static inline __attribute__((always_inline))
+#    else
+#      define FORCE_INLINE static inline
+#    endif
 #  else
 #    define FORCE_INLINE static
 #  endif /* __STDC_VERSION__ */
@@ -99,6 +99,7 @@ static void* XXH_memcpy(void* dest, const void* src, size_t size)
     return memcpy(dest,src,size);
 }
 
+
 /**************************************
 *  Basic Types
 ***************************************/
@@ -116,7 +117,6 @@ typedef unsigned int       U32;
 typedef   signed int       S32;
 typedef unsigned long long U64;
 #endif
-
 
 #if defined(__GNUC__)  && !defined(XXH_USE_UNALIGNED_ACCESS)
 #  define _PACKED __attribute__ ((packed))
@@ -190,6 +190,7 @@ static U64 XXH_swap64 (U64 x)
 }
 #endif
 
+
 /**************************************
 *  Constants
 ***************************************/
@@ -252,6 +253,7 @@ FORCE_INLINE U64 XXH_readLE64(const void* ptr, XXH_endianess endian)
 {
     return XXH_readLE64_align(ptr, endian, XXH_unaligned);
 }
+
 
 /****************************
 *  Simple Hash Functions
@@ -916,6 +918,7 @@ XXH_errorcode XXH64_freeState(XXH64_state_t* statePtr)
 {
     XXH_free(statePtr);
     return XXH_OK;
+}
 
 XXH128_state_t* XXH128_createState(void)
 {
--- a/xxhash.c
+++ b/xxhash.c
@@ -923,6 +923,16 @@ XXH_errorcode XXH128_freeState(XXH128_state_t* statePtr)
     return XXH_OK;
 }
 
+XXH256_state_t* XXH256_createState(void)
+{
+    XXH_STATIC_ASSERT(sizeof(XXH256_state_t) >= sizeof(XXH_istate256_t));   // A compilation error here means XXH256_state_t is not large enough
+    return (XXH256_state_t*)XXH_malloc(sizeof(XXH256_state_t));
+}
+XXH_errorcode XXH256_freeState(XXH256_state_t* statePtr)
+{
+    XXH_free(statePtr);
+    return XXH_OK;
+}
 
 /*** Hash feed ***/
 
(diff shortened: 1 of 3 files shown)


Cyan4973:
Hi @JCash 

Currently working on a 128/256 bits version of xxHash.
I'm using your own release as a baseline, and benchmarked it.
It seems to run as fast as xxh64, which is expected since the inner working is similar.
Such speed is btw fine, so it's a reasonable proposition.

As suggested in an earlier discussion,
primary selling point of xxh64 is its speed, as far as I know (albeit on 64-bits systems).
Therefore, I'm currently tempted to create faster versions for 128 & 256 bits.

Since one of the primary speed limitations is related to bus width,
I guess running faster is going to require > 64 bits memory accesses.
Hence the inevitable conclusion that vector code is the only remaining way to go (SSE , AVX)

Since availability is a scarce resource these days,
I will aim directly at 256 bits version using AVX2 intrinsic.

The objectives are : 
  - create a version much faster than regular 64-bits
  - keep the excellent dispersion/randomness qualities of xxh64
  - have a software backup for non-AVX2 CPU
     + which nonetheless perform reasonably well

This is a difficult set of properties, so the outcome is all but  known.
AVX2 instruction set remains relatively poor, in comparison with ALU standard opcodes, and as a consequence, quality of the hash is at stake. The requirement to be compatible with non-AVX2 instruction set limit the number of authorized instructions even more.

OK, let's start the investigation...

Cyan4973:
OK, let's start with a deeper understanding of what AVX2 can or cannot do.

To begin with, some operations are no brainer :

  - XOR, OR, AND : register wide
  - Add, Sub : available for signed integer 8, 16, 32 & 64 bits
  - Shift : available for integers 16, 32 & 64 bits

Which leaves us with 2 operations which are critical for xxHash distribution :

  - Rotate
  - Multiplication
  

### Rotate

I'll start with the simpler one.
`rotl` (_rotate left_) doesn't exist within AVX2.

The purpose of this operation is to push high bits into low position,
so that these bits can later "spread" thanks for multiplication.

There are 2 ways to achieve this objective :

  1. Emulate `rotl` : this is achievable using 2 SHIFTs and an ADD
     In contrast with `rotl`, which is a register operation,
     the above operation will require several registers,
     and some dependent operations. So it's likely to be slower.
     But that's nonetheless a fairly good emulation.
     
  2. Replace by a right shift `>>` and a XOR `^`.
     That's one less operation than previous proposition,
     it achieves a similar objective for high bits.
     Dependency is also the same (xor needs to wait for shift result).
     I'm unsure if the effect is really equivalent, but it looks like it imho.

Conclusion : replacing `rotl` looks achievable. 

     
### Multiplication
     
This is the main weakness of AVX2.
Multiplications are supported, but with the following limitations :
  - "normal" multiplications available for `int16_t` and `int32_t` only
    + No version for `uint64_t`
  - Slow : According to Intel doc, `vpmulld` has a latency of __10__ cycles,
    which is really a lot.

These characteristics creates 2 drawbacks :

  - A "software backup" will have to work on 32-bits register,
    which is twice slower than 64-bits (for 64-bits systems).
  - Hiding the latency is going to be hard.
    10 cycles is way too large.
    Impact on speed is going to be noticeable. 
    
There is another sort of multiplication available, but it doesn't really help :

  - `vpmuldq` is "only" 5 cycles of latency
    and can multiple 2 integer 32-bits values, 
    but store the result into 64 bits register.
    So it requires some register shuffle to get things done correctly.
    The register management task itself will cost more 
    than the difference in operation latency.    
    
So it leaves 2 possibilities :

  - Work on 32-bits, try to hide latency by doing other things in parallel
     + Not that simple, it will require to load a lot of registers.
       With one register already being 32 bytes,
       2 registers is 64 bytes (a cache line)
       4 registers is 128 bytes, and so on.
       Quickly, it limits the usefulness of the algorithm to very long input only.
       With only 8 AVX registers available in 32-bits mode, 
       and 16 available in 64-bits mode,
       there is also a hard limit to the number of parallel operations possible.

  - Emulate multiplication with shifts & add
     + Advantage : 64 bits operations are possible.
     + Not trivial either.
       Of course, doing a shift + add is trivial, and fast,
       but it's not about doing it once.
       xxHash works fine because its constants mix the bits correctly.
       Creating the same constants using shift/add
       is going to cost a huge number of operations,
       much more than the latency of multiply.
       So going this route requires to find some constants
       which can be emulated in a limited number of shift+add,
       but still nonetheless produce excellent spreading properties.
       It looks tricky. It's unclear if this objective is achievable.
       
Moving forward, I'm unsure of the better solution to investigate.

Cyan4973:
Some sharing here : 

I've been attempting again vectorization-based hashing late last year.
The results were not good.

At the very least, I've ruled out possibility to achieve that objective with SSE.
The critical problem with SSE is the number of registers. They are only 8.
In order to run fast, all operations must be completed in register.
This put a strong limits to the number of operations possible before overflowing the register count.
When the threshold is reached, compiler simply save / restore data from some stack space, but then, performance plummets.

So only AVX2 seems possible now. AVX2 registers are wider (256 bits) and more numerous (16), so there's enough room to stack and parallelize a few operations.

Here, the problem is the scarcity of operations possible.
Only shift, add/sub and xor. No rotation, no multiplication.
Rotation can be emulated by 2 shifts + or/xor/add, which is surprisingly slow.
Multiplication can be emulated by a serie of shift + add, and quickly becomes overwhelming.

That can be made to work, but it's hard to preserve good quality _and_ good speed.
And if the scheme is not at least 50% faster than xxh64 (on avx2), then it's not worth its complexity.
Moreover, all operations must remain accessible on a standard execution unit, at a speed no slower than xxh64. 

Investigation is not complete yet, but the remaining options are now pretty small.

Cyan4973:
A 128-bit variant of xxHash is finally arriving.
It's currently in PR at : https://github.com/Cyan4973/xxHash/pull/174

The variant can also generate 256-bit if needed (no strong request so far).

The new variant, called `XXH3`, is a lot faster than current `XXH32` and` XXH64`, on both 32 and 64-bit platforms. It has been developed with a deep focus on short inputs, which are useful for hash tables, bloom filters, and such. But long inputs have not been forgotten, and its vector-friendly core loop makes wonders. As a consequence, it has the potential to completely supercede existing `XXH32` and `XXH64` variants.

I will close this PR now that we have a clear path towards 128+ bit hash generation.

PR #19, opened 2015-02-01
script for easy checking of files

beckermr:
I have added a script to allow for easy checking of files in a single directory.

### DIFF
--- a/ckxxhsum
+++ b/ckxxhsum
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+verify="0"
+while getopts "c:" opt; do
+  case $opt in
+    c)
+      verify="1"
+      fname=$OPTARG
+      #echo "verify mode!"
+      ;;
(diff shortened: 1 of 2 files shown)


Cyan4973:
Thanks for the commit.

For information, I do not commit directly into master
(readme-only changes being the only exception).

I would be glad to redirect your commit to "dev", but Github doesn't support it, unfortunately.
Would you mind redirecting your patch into the "dev" branch ? Thanks


Also, related question : 
why do you feel the "loading" message should be removed ?
I expect this message to be useful when the file to load/hash is very large.
It avoids having the user wondering if the program works or not.

Regards

beckermr:
I will redirect my commit. As far as I know, the loading message is not consistent with other unix-like tools of this sort (e.g., md5sum). Also, the script I added uses stdout to capture the checksum. The loading message interferes with this.

PR #21, opened 2015-02-01
script for checking

beckermr:


Cyan4973:
@beckermr 
Hi Matthew 

I'm back working on xxHash, and considering your patch.
If I do properly understand the objective, spelled in #19,
your batch script "allow for easy checking of files in a single directory".

Would you mind elaborating what it does ?

>  the script I added uses stdout to capture the checksum. The loading message interferes with this.

The loading message is sent to `stderr` only, while the result is sent to `stdout`.
This difference should ensure that any tool capturing `stdout` only get the result.

Does it nonetheless cause issues for you ?

Regards

Cyan4973:
Sorry @beckermr , I just don't understand what your PR is providing. Put on hold while waiting for explanations.

beckermr:
This pull request adds a script which runs xxHash, but behaves just like the md5sum unix utility. See the docs for it here, for example, 

http://linux.about.com/library/cmd/blcmdl1_md5sum.htm

or the man page on your local linux box. 

This means you can do things like 
```bash
ckxxhsum * > xxhsums.chk
ckxxhsum -c xxhsums.chk
```

Then if any of the files captured by the wildcard have changed, `ckxxhsum` will report an error.

Cyan4973:
Thanks for the detailed explanation.
I'll have a look at your PR.

Cyan4973:
OK, the use case is clear enough.
I believe xxhsum should be upgraded to be able to handle multiple files, 
so that `./xxhsum * > out` work correctly.

beckermr:
Should the -c option be added to xxhsum as well? Otherwise the ckxxhsum script would still be needed.

Cyan4973:
Latest version in `dev` branch allows multiple input files and wildcard support. So `xxhsum * > out` will work appropriately, similarly to md5sum.

There is still no support for `-c` command though, so `ckxxhsum` script is still useful.

beckermr:
Hi all! With the merging of #53, I am going to close this PR.

Cyan4973:
:+1:

PR #22, opened 2015-02-02
Add const to types to fix GCC warnings

noporpoise:
GCC gives multiple warnings due to casting a const pointer to a non-const type. Adding in const qualifiers in a few places fixes these warnings.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@
 
 CC     := $(CC)
 CFLAGS ?= -O3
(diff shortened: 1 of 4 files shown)


Cyan4973:
Thanks
Indeed, preserving const properties is important.

Note that branch policy requires contribution to be directed to the "dev" branch. I don't accept merge directly into master.

The majority of the proposed changes are fine, but a minority of them will require some careful analysis, just to be sure.

PR #23, opened 2015-02-03
Add const to types to fix GCC warnings

noporpoise:
GCC gives multiple warnings due to casting a const pointer to a non-const type. Adding in const qualifiers in a few places fixes these warnings.

Note: This is a new pull request to replace PR #22, which was against master rather than dev branch.

### DIFF
--- a/xxhash.c
+++ b/xxhash.c
@@ -145,8 +145,8 @@ typedef struct _U64_S
 #  pragma pack(pop)
 #endif
 
-#define A32(x) (((U32_S *)(x))->v)
-#define A64(x) (((U64_S *)(x))->v)
+#define A32(x) (((const U32_S *)(x))->v)
+#define A64(x) (((const U64_S *)(x))->v)
 
 
 /*****************************************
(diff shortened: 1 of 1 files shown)


noporpoise:
Any chance this could get pulled?

Cyan4973:
Hi Isaac

Sorry, I'm currently using my small available time on finishing lz4 r129.
xxHash is next, your pull request will be evaluated during this period.

Cyan4973:
With lz4 almost completed, I can start to converge my attention again on xxHash.

Looking at your patch, it's good.

I'm just a bit surprised.
Casting a `const` pointer to a non-const one is supposed to be an important warning. And the xxhash Makefile is no shy of warnings. Following ones are used : 
````
-Wall -Wextra -Wundef -Wshadow -Wcast-align -Wstrict-prototypes
````

So I'm surprised such important issue remained undetected up to now. Are you using another compilation flag which ensures that the warning is triggered ? If yes, this would be useful, to add to the xxHash test suite, ensuring such issue do not happen again in the future.

Note that I'll have to sync the current source code with the newest version from lz4, so it may break your patch.

Regards

noporpoise:
GCC with `-Wcast-qual` gives the warning. Clang does not warn. 

Thanks,
Isaac

Cyan4973:
Thanks Isaac. It's a very useful compilation flag.
I'm surprised it's not included within `-Wall` or at least `-Wextra`,
since preserving the `const` property of a memory segment is extremely important  imho.

Anyway, thanks to your suggestion, I could reproduce your issue and catch all warnings.
All of them are false positive, since the `const` property is never abused,
but it's nonetheless good to formally ensure it through preservation of `const` property at every stage.

I'll post a fixed version soon.

Cyan4973:
A new version has been uploaded in the "dev" branch : 3b63fc28cac451f0179145467e6410d17a926bf7

It's supposed to fix this issue.

Regards

JCash:
You can also try -Wpedantic which is the next level after -Wall and -Wextra

Cyan4973:
`-pedantic` is already part of the compilation flags.
Unfortunately, it did not catch *this* specific issue.

PR #31, opened 2015-08-16
CMakeLists.txt: support installing files

Inkane:
Those two lines will tell CMake to create an install target for the generated Makefile (or it's equivalent if another CMake generator is used). It installs the header and the compiled library.


Cyan4973:
good

PR #55, opened 2016-02-19
Add manpage and its markdown source

t-mat:
This PR adds ROFF format manpage `xxhsum.1`.  Also `xxhsum.1` will be genareted from Markdown file `xxhsum.1.md` by [ronn](https://github.com/rtomayko/ronn).
You can generate `xxhsum.1` by the following procedure

```
sudo apt-get install -y ruby-ronn
# make man generates xxhsum.1 from xxhsum.1.md #
make man

# edit xxhsum.1.md here #
# auto preview for development
make preview-man
```

To keep it simple, all `Makefile` targets are written in explicit fashion.

## Removal of `ronn`'s version

`Makefile` has the following tricky part:

```
xxhsum.1: xxhsum.1.md
        cat $^ | $(MD2ROFF) $(MD2ROFF_FLAGS) | sed -n '/^\.\\\".*/!p' > $@
```

Last `sed` command removes `ronn`'s comment lines (`ronn`'s version, author, URL).  It doesn't mean disrespect for `ronn`.  We need this removal because we must keep independent `xxhsum.1` from `ronn`'s version.

## Possible problems

 - This PR adds external dependency for `ronn`.
    - Here is a tradeoff.
    - It requires ruby and rubygems environment.  So I don't say it's lightweight dependency.
 - Should we contain `xxhsum.1` in the repository?
    - Since we can generate it from `xxhsum.1.md`, we can remove it from repository.
    - But some users who can't run `ronn` can't generate `xxhsum.1`.
 - I should verify `sed` syntax for BSD families.
    - If we remove `xxhsum.1` from repository, we can also remove this `sed` filter.
    - Since `ronn` requires ruby, there is possibility to replace `sed` with `ruby` one-liner.

I think this is an experimental PR, please feel free to comment anything.

update:
I've checked `sed` syntax on Mac OSX 10.11.3 and FreeBSD 10.2-RELEASE-p12.  It works fine.

PR #56, opened 2016-02-22
Add -w, --version, --help and fix XXHSUM_VERSION

t-mat:
This PR is follow up of PR #55.  All commits are trivial.

 - Fix XXHSUM_VERSION in Makefile
 - Add the following options to mimic `md5sum`
    - `-w` : alias of `--warn`
    - `--version` : alias of `-V`
    - `--help` : alias of `-h`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ LIBVER := $(LIBVER_MAJOR).$(LIBVER_MINOR).$(LIBVER_PATCH)
 CFLAGS ?= -O3
 CFLAGS += -std=c99 -Wall -Wextra -Wshadow -Wcast-qual -Wcast-align -Wstrict-prototypes -Wstrict-aliasing=1 -Wswitch-enum -Wundef -pedantic 
 FLAGS  := $(CFLAGS) $(CPPFLAGS) $(LDFLAGS) $(MOREFLAGS)
-XXHSUM_VERSION=0.5.1
+XXHSUM_VERSION=$(LIBVER)
 MD2ROFF  =ronn
 MD2ROFF_FLAGS  = --roff --warnings --manual="User Commands" --organization="xxhsum $(XXHSUM_VERSION)"
 
--- a/xxhsum.1
+++ b/xxhsum.1
@@ -1,5 +1,5 @@
 .
-.TH "XXHSUM" "1" "February 2016" "xxhsum 0.5.1" "User Commands"
+.TH "XXHSUM" "1" "February 2016" "xxhsum 0.5.0" "User Commands"
 .
 .SH "NAME"
 \fBxxhsum\fR \- print or check xxHash non\-cryptographic checksums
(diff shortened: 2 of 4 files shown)


Cyan4973:
minor detail, but I was wondering if it could be beneficial to create 2 sections in the man page ?

Like the first part for "normal" (expected) commands (`-H1`, `-c`, etc.),
basically, the core service, with simple use cases,

and the second part for less standard commands (`-b`, `-i1`, etc.),
basically, for advanced/power users.

What do you think ?

t-mat:
> What do you think ?

It's nice and useful.  I think man-pages are mainly designed for two purposes: forward and reverse reference *book* (in 1970's).  I mean "forward reference" as reading man-pages to create/write something new.  Also "reverse reference" as reading man-pages to read/modify existent command options.
Addition of new section (heading) such as "ADVANCED OPTIONS" is good for forward reference.  However [man-pages(7)](http://man7.org/linux/man-pages/man7/man-pages.7.html#DESCRIPTION) doesn't recommend it:

>  Where a traditional heading would apply, please use it; this kind of
>  consistency can make the information easier to understand.  If you
>  must, you can create your own headings if they make things easier to
>  understand (this can be especially useful for pages in Sections 4 and
>  5).  However, before doing this, consider whether you could use the
>  traditional headings, with some subsections (.SS) within those
>  sections.

Even swiss army knife style commands such as [curl(1)](https://curl.haxx.se/docs/manpage.html) doesn't split "OPTIONS", it seems still good to follow this tradition for reverse reference purpose.

Instead of additional section (heading), I propose the following modified structure

 - Say `xxhsum -b` is special in SYNOPSIS and DESCRIPTION.
 - (optional) Virtually split OPTIONS by the message: "The following options are useful only benchmark purpose"
 - Add more example for EXAMPLE.

----

SYNOPSIS
--------

`xxhsum` [_OPTION_] ... [_FILE_] ...
`xxhsum` -b [_OPTION_] ...

DESCRIPTION
-----------

... descriptions ...

`xxhsum -b` invokes benchmark mode.  See OPTIONS and EXAMPLE for details.


OPTIONS
-------

... "normal" options ...

**The following options are useful only benchmark purpose**

* `-b`:
  Benchmark mode.  See EXAMPLE for details.

* `-B`_BLOCKSIZE_:
  Only useful for benchmark mode (`-b`).  See EXAMPLE for details.  _BLOCKSIZE_ specifies benchmark mode's test data block size in bytes.  Default value is 102400

* `-i`_ITERATIONS_:
  Only useful for benchmark mode (`-b`).  See EXAMPLE for details.  _ITERATIONS_ specifies number of iterations in benchmark.  Single iteration takes at least 2500 milliseconds.  Default value is 3


EXAMPLE
-------

... examples ...

Benchmark xxHash algorithm for 16384 bytes data in 10 times. `xxhsum` benchmarks xxHash algorithm for 32-bit and 64-bit and output results to standard output.  First column means algorithm, second column is source data size in bytes, last column means hash generation speed in mega-bytes per seconds.

    $ xxhsum -b -i10 -B16384

----

You can see same kind of structure in [ronn(1)](http://rtomayko.github.io/ronn/ronn.1.html#OPTIONS).

Cyan4973:
:+1: 
This is a good suggestion @t-mat

PR #60, opened 2016-03-24
changes to increase xxhash' configurability when inlining it

cota:
The first commit is trivial.

The other two would be nice to have; my use case is a portable high-performance application that inlines the hashing calls. With these two changes, it can get top performance on all platforms (e.g. big-endian), and it can save a branch on every hash call (on, say, ARM) since the app is guaranteed to pass aligned input data.

Can you please take a look?

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+xxh32sum
+xxh64sum
+xxhsum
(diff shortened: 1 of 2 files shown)


Cyan4973:
These are pretty good suggestions @cota .

I'll be glad to integrate them.
The only minor issue is that all PR must go through "dev" branch first (no direct PR to master).
I you could redirect your PR there, I'll accept it.

PR #61, opened 2016-03-25
[v2] changes to increase xxhash' configurability when inlining it

cota:
Supersedes pull request #60; this PR targets the dev branch instead of master.

### DIFF
--- a/xxhash.c
+++ b/xxhash.c
@@ -76,13 +76,18 @@ You can contact the author at :
 #define XXH_FORCE_NATIVE_FORMAT 0
 #endif
 
-/*!XXH_USELESS_ALIGN_BRANCH :
+/*!XXH_FORCE_ALIGN_CHECK :
  * This is a minor performance trick, only useful with lots of very small keys.
- * It means : don't check for aligned/unaligned input, because performance will be the same.
- * It saves one initial branch per hash.
+ * It means : check for aligned/unaligned input.
+ * The check costs one initial branch per hash; set to 0 when the input data
+ * is guaranteed to be aligned.
  */
+#ifndef XXH_FORCE_ALIGN_CHECK /* can be defined externally */
 #if defined(__i386) || defined(_M_IX86) || defined(__x86_64__) || defined(_M_X64)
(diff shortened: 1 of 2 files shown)


xxhash.c
```diff
@@ -76,13 +76,18 @@ You can contact the author at :
 #define XXH_FORCE_NATIVE_FORMAT 0
 #endif
 
-/*!XXH_USELESS_ALIGN_BRANCH :
+/*!XXH_FORCE_ALIGN_CHECK :
```
Cyan4973: I like the new name. Its intention is much clearer.

xxhash.c
```diff
@@ -391,15 +396,16 @@ XXH_PUBLIC_API unsigned int XXH32 (const void* input, size_t len, unsigned int s
 #else
     XXH_endianess endian_detected = (XXH_endianess)XXH_CPU_LITTLE_ENDIAN;
 
-#  if !defined(XXH_USELESS_ALIGN_BRANCH)
-    if ((((size_t)input) & 3) == 0)   /* Input is 4-bytes aligned, leverage the speed benefit */
+    if (XXH_FORCE_ALIGN_CHECK)
```
Cyan4973: The test was initially an `#ifdef` macro.
Now it's a regular `if`, with the expectation that compiler should remove the useless code segment (which is pretty reasonable).

Any specific reason you preferred the latest version ?

cota:
No particular reason. I just did it that way because I saw if `(foo || XXH_FORCE_NATIVE_FORMAT)` also in the file, so I figured you liked it that way :)
Personally I'd prefer `#if XXX_FORCE_ALIGN_CHECK`; let me know if you want me to update the commit.

Cyan4973:
LGTM

I _tend to_ prefer regular `if` , because it will allows better integration with the compilers and debuggers.
That being said, it implies that the `macro` is necessarily defined with an integer value (typically `0` or `1`). If there is a chance it could be undefined, or just defined with no value, then it's not the right construction.

In this case, it looks to me that you made it clear the macro is necessarily defined with an integer, except if fucked up by some wrong external definition

So I agree the `if` construction looks best.

PR #82, opened 2016-08-18
Replace GCC_VERSION with XXH_GCC_VERSION

Leandros:
GCC_VERSION is quite a common macro name, and might clash with already
defined macros. It's safer to use a XXH_ prefix.

I've seen LZ4 does the same, I therefore submitted a simple pull-request to apply the same in xxHash.


Cyan4973:
Good point !
Could you please redirect your PR into the "dev" branch ?
See https://github.com/Cyan4973/xxHash#branch-policy
Thanks !

Leandros:
Oops, missed that. Sure.

PR #89, opened 2016-10-26
few typo fixes in comments or docs

reedjc:
I was doing some cleanup of some company code and noticed that I should  share this upstream. Thank you for sharing your code to the world.

(Note I wanted to fix endianess to endianness but did not. Maybe consider that also.)


Cyan4973:
Indeed, thanks !

PR #90, opened 2016-11-15
Build the xxhsum binary with cmake as well

ChrisKitching:
As well as some general tidying up, this also allows the cmake build to compile the xxhsum binary,

### DIFF
--- a/cmake_unofficial/CMakeLists.txt
+++ b/cmake_unofficial/CMakeLists.txt
@@ -6,7 +6,9 @@ project(xxhash)
 set(XXHASH_LIB_VERSION "0.42.0")
 set(XXHASH_LIB_SOVERSION "0")
 
-set(BUILD_SHARED_LIBS ON CACHE BOOL "Set to ON to build shared libraries")
+option(BUILD_SHARED_LIBS "Set to ON to build shared libraries" ON)
+option(BUILD_STATIC_LIBS "Set to ON to build static libraries" ON)
+
 if(BUILD_SHARED_LIBS)
   add_library(xxhash SHARED ../xxhash.c)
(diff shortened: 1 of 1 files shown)


t-mat:
Great work, @ChrisKitching !
@nemequ, could you review this PR?  Since you've already impressive contribution to LZ4, it'd be nice if we have consistent policy and convention for xxHash.

t-mat:
FYI : @ChrisKitching , nemequ has been maintained LZ4's CMake support.  https://github.com/lz4/lz4/pull/268/commits/151021a3a628189e072d828ff01247063b9baddc

nemequ:
> @nemequ, could you review this PR? Since you've already impressive contributed to LZ4, it'd be nice if we have consistent policy and convention for xxHash.

This is definitely an improvement; looking at this in isolation I'd say to merge it.

That said, bringing xxhash up to parity with the new cmake stuff in lz4 would mean clobbering these changes; it would be much easier to just copy the file from lz4 and adapt it for xxhash than to add the relevant functionality to xxhash's existing cmake.  If that's what you want I don't mind putting together a PR (it would only take a few minutes), but in that case it won't really matter if this is merged or not so I don't see a reason to complicate the history.

Cyan4973:
This is great improvement, so I'll merge it.
There is no hurry to create a new release, so it remainsl possible to bring even more improvements to cmake build.

I believe one of the most important improvements @nemequ brought to LZ4 cmake is the ability to extract version number from lz4.h : https://github.com/lz4/lz4/blob/dev/contrib/cmake_unofficial/CMakeLists.txt#L15

The version number is indeed useful when creating library suffixes.

PR #93, opened 2016-11-17
cmake: rewrite based on LZ4's CMake support

nemequ:
Copied over the CMake stuff from LZ4, simplified a bit (no pkg-config, etc.).

I haven't tested very thoroughly, but it's pretty simple.  @ChrisKitching, since I know you're interested in the CMake support maybe you'd like to take a look?

### DIFF
--- a/cmake_unofficial/CMakeLists.txt
+++ b/cmake_unofficial/CMakeLists.txt
@@ -1,7 +1,9 @@
-cmake_minimum_required(VERSION 2.6)
-cmake_policy(VERSION 2.6)
-
-project(xxhash)
+# To the extent possible under law, the author(s) have dedicated all
+# copyright and related and neighboring rights to this software to
+# the public domain worldwide. This software is distributed without
+# any warranty.
+#
+# For details, see <http://creativecommons.org/publicdomain/zero/1.0/>.
 
 set(XXHASH_DIR "${CMAKE_CURRENT_SOURCE_DIR}/..")
 
@@ -19,33 +21,80 @@ mark_as_advanced(XXHASH_VERSION_MAJOR XXHASH_VERSION_MINOR XXHASH_VERSION_RELEAS
 option(BUILD_XXHSUM "Build the xxhsum binary" ON)
 option(BUILD_SHARED_LIBS "Build shared library" ON)
 
-# Make CMake's RPATH handling not be insane. This suff has cmake set rpaths appropriately for
-# where things end up in the install tree. For some reason that's not the default:
-# https://cmake.org/Wiki/CMake_RPATH_handling
-SET(CMAKE_SKIP_BUILD_RPATH FALSE)
-SET(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
-
-# Where we search for shared libraries
-SET(CMAKE_INSTALL_RPATH "./lib")
-
-# add the automatically determined parts of the RPATH
-# which point to directories outside the build tree to the install RPATH
-SET(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
-
-add_library(xxhash ../xxhash.c)
-set_target_properties(xxhash PROPERTIES COMPILE_DEFINITIONS "XXHASH_EXPORT"
-                       VERSION "${XXHASH_LIB_VERSION}"
-                       SOVERSION "${XXHASH_LIB_SOVERSION}")
-
-if (BUILD_XXHSUM)
-    add_executable(xxhsum ../xxhsum.c)
-    target_link_libraries(xxhsum xxhash)
+if("${CMAKE_VERSION}" VERSION_LESS "3.0")
+  project(XXHASH C)
+else()
+  cmake_policy (SET CMP0048 NEW)
+  project(XXHASH
+    VERSION ${XXHASH_VERSION_STRING}
+    LANGUAGES C)
 endif()
 
-INSTALL(FILES ../xxhash.h DESTINATION include)
-INSTALL(
-    TARGETS xxhash xxhsum
-    RUNTIME DESTINATION bin
-    ARCHIVE DESTINATION lib
-    LIBRARY DESTINATION lib
-)
+cmake_minimum_required (VERSION 2.8.12)
+
+# If XXHASH is being bundled in another project, we don't want to
+# install anything.  However, we want to let people override this, so
+# we'll use the XXHASH_BUNDLED_MODE variable to let them do that; just
+# set it to OFF in your project before you add_subdirectory(xxhash/contrib/cmake_unofficial).
+if(CMAKE_CURRENT_SOURCE_DIR STREQUAL "${CMAKE_SOURCE_DIR}")
+  # Bundled mode hasn't been set one way or the other, set the default
+  # depending on whether or not we are the top-level project.
+  if("${XXHASH_PARENT_DIRECTORY}" STREQUAL "")
+    set(XXHASH_BUNDLED_MODE OFF)
+  else()
+    set(XXHASH_BUNDLED_MODE ON)
+  endif()
+endif()
+mark_as_advanced(XXHASH_BUNDLED_MODE)
+
+# Allow people to choose whether to build shared or static libraries
+# via the BUILD_SHARED_LIBS option unless we are in bundled mode, in
+# which case we always use static libraries.
+include(CMakeDependentOption)
+CMAKE_DEPENDENT_OPTION(BUILD_SHARED_LIBS "Build shared libraries" ON "NOT XXHASH_BUNDLED_MODE" OFF)
+
+include_directories("${XXHASH_DIR}")
+
+# libxxhash
+add_library(xxhash "${XXHASH_DIR}/xxhash.c")
+set_target_properties(xxhash PROPERTIES
+  SOVERSION "${XXHASH_VERSION_STRING}"
+  VERSION "${XXHASH_VERSION_STRING}")
+
+# xxhsum
+add_executable(xxhsum "${XXHASH_DIR}/xxhsum.c")
+target_link_libraries(xxhsum xxhash)
+
+# Extra warning flags
+include (CheckCCompilerFlag)
+foreach (flag
+    -Wall -Wextra -Wcast-qual -Wcast-align -Wshadow
+    -Wstrict-aliasing=1 -Wswitch-enum -Wdeclaration-after-statement
+    -Wstrict-prototypes -Wundef)
+  # Because https://gcc.gnu.org/wiki/FAQ#wnowarning
+  string(REGEX REPLACE "\\-Wno\\-(.+)" "-W\\1" flag_to_test "${flag}")
+  string(REGEX REPLACE "[^a-zA-Z0-9]+" "_" test_name "CFLAG_${flag_to_test}")
+
+  check_c_compiler_flag("${ADD_COMPILER_FLAGS_PREPEND} ${flag_to_test}" ${test_name})
+
+  if(${test_name})
+    set(CMAKE_C_FLAGS "${flag} ${CMAKE_C_FLAGS}")
+  endif()
+
+  unset(test_name)
+  unset(flag_to_test)
+endforeach (flag)
+
+if(NOT XXHASH_BUNDLED_MODE)
+  include(GNUInstallDirs)
+
+  install(TARGETS xxhsum
+    RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}")
+  install(TARGETS xxhash
+    LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
+    ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}")
+  install(FILES "${XXHASH_DIR}/xxhash.h"
+    DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
+  install(FILES "${XXHASH_DIR}/xxhsum.1"
+    DESTINATION "${CMAKE_INSTALL_MANDIR}/man1")
+endif(NOT XXHASH_BUNDLED_MODE)


Cyan4973:
Looks good to me

cmake_unofficial/CMakeLists.txt
```diff
@@ -1,40 +1,108 @@
...
-    add_executable(xxhsum ../xxhsum.c)
-    target_link_libraries(xxhsum xxhash)
+# Whether to use position independent code for the static library.  If
+# we're building a shared library this is ignored and PIC is always
+# used.
+option(XXHASH_POSITION_INDEPENDENT_LIB "Use position independent code for static library (if applicable)" ON)
+if(XXHASH_POSITION_INDEPENDENT_LIB OR BUILD_SHARED_LIBS)
+  set(XXHASH_POSITION_INDEPENDENT_CODE TRUE)
+else()
+  set(XXHASH_POSITION_INDEPENDENT_CODE FALSE)
```
ChrisKitching: Consider using the standard cmake variable for this:
https://cmake.org/cmake/help/v3.0/variable/CMAKE_POSITION_INDEPENDENT_CODE.html#variable:CMAKE_POSITION_INDEPENDENT_CODE

In general, using standard cmake variables instead of inventing your own works better when other people want to do strange things with your build system (such as what I have done: convince it to cross compile by just passing extra cmake options externally)

cmake_unofficial/CMakeLists.txt
```diff
@@ -1,40 +1,108 @@
...
-    RUNTIME DESTINATION bin
-    ARCHIVE DESTINATION lib
-    LIBRARY DESTINATION lib
-)
+# libxxhash
+add_library(xxhash "${XXHASH_TOP_SOURCE_DIR}/xxhash.c")
+set_target_properties(xxhash PROPERTIES
+  SOVERSION "${XXHASH_VERSION_STRING}"
+  VERSION "${XXHASH_VERSION_STRING}"
+  POSITION_INDEPENDENT_CODE ${XXHASH_POSITION_INDEPENDENT_CODE})
```
ChrisKitching: Had you set `CMAKE_POSITION_INDEPENDENT_CODE` above, you wouldn't need to list it here explicitly.

cmake_unofficial/CMakeLists.txt
```diff
@@ -1,40 +1,108 @@
...
+
+# xxhsum
+add_executable(xxhsum "${XXHASH_TOP_SOURCE_DIR}/xxhsum.c")
+target_link_libraries(xxhsum xxhash)
+
+# Extra warning flags
+include (CheckCCompilerFlag)
+foreach (flag
+    -Wall -Wextra -Wcast-qual -Wcast-align -Wshadow
```
ChrisKitching: You forgot `-Wpedantic` ;)  (gcc only, though. Quite irritating)

cmake_unofficial/CMakeLists.txt
```diff
@@ -1,40 +1,108 @@
...
-SET(CMAKE_SKIP_BUILD_RPATH FALSE)
-SET(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
+cmake_minimum_required (VERSION 2.8.6)
 
-# Where we search for shared libraries
-SET(CMAKE_INSTALL_RPATH "./lib")
+# If XXHASH is being bundled in another project, we don't want to
+# install anything.  However, we want to let people override this, so
+# we'll use the XXHASH_BUNDLED_MODE variable to let them do that; just
+# set it to OFF in your project before you add_subdirectory(xxhash/contrib/cmake_unofficial).
```
ChrisKitching: I should also add that cmake effectively provides a standard way to do this by modifying CMAKE_INSTALL_PREFIX to some random temporary directory.

Bulat-Ziganshin:
>Why not have cmake generate the version header

it will make harder to build without cmake

cmake_unofficial/CMakeLists.txt
```diff
@@ -1,40 +1,108 @@
...
+set(XXHASH_TOP_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/..")
 
-set(XXHASH_LIB_VERSION "0.42.0")
-set(XXHASH_LIB_SOVERSION "0")
+# Parse version information
+file(STRINGS "${XXHASH_TOP_SOURCE_DIR}/xxhash.h" XXHASH_VERSION_MAJOR REGEX "^#define XXH_VERSION_MAJOR +([0-9]+) *$")
+string(REGEX REPLACE "^#define XXH_VERSION_MAJOR +([0-9]+) *$" "\\1" XXHASH_VERSION_MAJOR "${XXHASH_VERSION_MAJOR}")
+file(STRINGS "${XXHASH_TOP_SOURCE_DIR}/xxhash.h" XXHASH_VERSION_MINOR REGEX "^#define XXH_VERSION_MINOR +([0-9]+) *$")
+string(REGEX REPLACE "^#define XXH_VERSION_MINOR +([0-9]+) *$" "\\1" XXHASH_VERSION_MINOR "${XXHASH_VERSION_MINOR}")
+file(STRINGS "${XXHASH_TOP_SOURCE_DIR}/xxhash.h" XXHASH_VERSION_RELEASE REGEX "^#define XXH_VERSION_RELEASE +([0-9]+) *$")
```
Bulat-Ziganshin: >Why not have cmake generate the version header

it will make harder to build without cmake

t-mat: Thanks for reviewing @ChrisKitching !

But in this case, CMake must consume the header.

We must avoid to rely on CMake for fundamental purpose.  Since vesion number is one of the vital part of xxHash, we've been keeping it as simple as possible : plain code.
And it seems it's good decision.  Because there are a lot of projects which just include/copy xxHash code and simply it works.

Please note that so far we're also avoiding other (good but) complex and environment dependent things such as `autoconf` to keep plain code working well.

Cyan4973:
The comments on default cmake variables look valid to me.

For the other proposal, I believe the source of truth must remain `xxhash.h` .
Have a "single source of truth" helps a lot, and it can only be the source code, considering build systems are many.

cmake_unofficial/CMakeLists.txt
Orphis: ExternalProject is more of a hack than a proper integration. It won't propagate any compilation flag or anything, which is a requirement when embedding a library like this.

The proper way to ignore the install() calls is through the option `EXCLUDE_FROM_ALL` in `add_subdirectory()` (not sure how far it goes, but that's what you're supposed to do in modern CMake, not use another hack with a temp install folder).

cmake_unofficial/CMakeLists.txt
Orphis: This isn't available in CMake 2.8.6 the project targets anyway.

cmake_unofficial/CMakeLists.txt
Orphis: Not available in 2.8.6. I would recommend raising the minimum version higher.

cmake_unofficial/CMakeLists.txt
Orphis: I agree on CMake consuming the header. There should only be one source of truth.
Also, consider setting the folder property `CMAKE_CONFIGURE_DEPENDS` to the header file to make sure the generate step depends on it and will rerun automatically.

Orphis:
For proper CMake support, you would also use `target_include_directories()` on the libxxhash target to ensure projects embedding the library have all the headers available. It is a very good practice and having to manually propagate the include directories in the projects using xxhash is quite error prone.

Also, this requires CMake 2.8.12 if I remember correctly. You can either raise the minimum version of have conditionals on the CMake version.

Cyan4973:
This PR hasn't moved in while.
Presumably, the first round of needlessly harsh comments did not helped.
Not sure if @nemequ is still interested in moving it forward ?

Orphis:
Classic Github behaviour were work stops when the PR is submitted!
I could maybe comeback and do something about it when I'm done cleaning up the CMake scripts in other projects. I just happened to have a minimal CMake script for this library in those and I was curious to know if you had a proper CMake script available.

nemequ:
Sorry, I actually had to step away from open source stuff for a few months due to time constants, and only came back over the weekend.  I'm working my way through my backlog now, but this is literally the last item on my list…  I figure I've already disappointed the people at that end, so what's a few more days?

The first round of comments didn't have any effect; I actually still haven't read them.  I'm pretty thick-skinned, though, so unless they're really bad it's probably not an issue ;)

ChrisKitching:
> The first round of comments didn't have any effect; I actually still haven't read them. I'm pretty thick-skinned, though, so unless they're really bad it's probably not an issue ;)

Didn't mean to be "harsh". That said, I probably shouldn't review random people's patches when frazzled from short deadlines myself. Sorry about that.

Glad to hear it didn't discourage you.

nemequ:
Okay, now I'm curious…

I had no idea about CMAKE_POSITION_INDEPENDENT_CODE.  Agree that should be used.  Disagree about ExternalProject; it has some annoying drawbacks.

Nope, there was nothing in there I had any problem with.  I'll reply in detail soon; I want to play with the CMAKE_PIC thing a bit.

Thanks for the review :)

cmake_unofficial/CMakeLists.txt
nemequ: LOL.  I stole these from [the Makefile](https://github.com/Cyan4973/xxHash/blob/dev/Makefile#L34), I don't have a strong opinion on what to put in there, that's up to Yann.  There are a bunch of [possibilities in Squash](https://github.com/quixdb/squash/blob/master/cmake/ExtraWarningFlags.cmake), though.  -Wpedantic isn't one of them (it's quite irritating, and IMHO not worth the noise).

cmake_unofficial/CMakeLists.txt
nemequ: Totally agree about it being a "hack"… it's kinda like executing a command from a C program instead of calling into a library; it has its place, but if you have a choice a library is generally a better way to go.

ExternalProject basically isolates the child project, and in order to carry things over you have to interact with its build system's interface.  If you apply something to a parent project (set compiler or linker flags, choose a compiler, add a define, etc.) it doesn't get carried through to the child.

Things also don't go the other way… you can't interact with targets automatically, you need to figure out where files were generated and recreate targets in the parent project, including paths, any required flags (i.e., whatever `pkg-config --cflags child` would spit out), etc.

If that's how someone wants to interact with xxHash's build system, that's still an option.  Bundled mode is basically just an additional option that provides more control and easier integration.

That said, I agree it's a bit strange.  I don't see it all that often in the wild, but I think it's a much nicer way to go.  Since it's not hurting anything, I don't see a reason not to have it.

nemequ:
Okay, so CMake versions…

Solaris 11 is still stuck at 2.8.6 (or was a few months ago when I last checked).  Ubuntu 12.04 (LTS, still supported) uses 2.8.7.  Ubuntu 12.04 is pretty important because a lot of hosted CI services still use it; Travis, Drone, and I think some others.  Personally, I'd love to bump it to something recent; a lot of newer features make CMake *vastly* less unpleasant, but IMHO the world isn't ready yet :(.

Orphis:
You can have a newer version on Solaris, but you have to build it yourself (or get your package from somewhere else). CMake is highly portable, so that isn't a problem.
Ubuntu 12.04 EOL is in a few months (5 years total), so you should target 14.04 instead (CMake 2.8.12) or newer. TravisCI provides trusty too, which is what most people care about. Even Debian Stable has version 3.0, and 3.6 in backports and stretch the current testing at 3.7.2 (soon moved to stable).

Also, people don't have CMake scripts right now, so if they want to use the library on their old system, they have working solutions already. So you don't have to care about compatibility and could aim for a much more recent version.

The main issue with CMake is that people target 2.6 or 2.8 all the time which severely limits how cleanly you can write your scripts. Even tarting 2.8 is wrong as each 2.8.x is quite different from the others. People usually target newer versions, like 2.8.12 but just write "2.8" only.

nemequ:
> Also, people don't have CMake scripts right now, so if they want to use the library on their old system, they have working solutions already. So you don't have to care about compatibility and could aim for a much more recent version.

This paragraph convinced me.  Assuming Yann agrees, I guess I'll update the PR (in a week or so…)

 * Bump CMake dep to 2.8.12 (and try using it to verify it works)
 * Use CMAKE_POSITION_INDEPENDENT_CODE

Am I missing anything?

Orphis:
Remove include_directories and use target_include_directories instead. 
Use the folder property I mentioned to watch the header file for changes too. 
I can continue to review when that is done.

nemequ:
Finally got back to this…

 * Bumped CMake dependency to 2.8.12
 * Dropped the `XXHASH_POSITION_INDEPENDENT_CODE` stuff
 * Used `target_include_directories`

Rebased, squashed.

Cyan4973:
great news !
Do you feel it's ready for a merge ? Or do you want to work on it a bit more ?

nemequ:
>  Do you feel it's ready for a merge ? Or do you want to work on it a bit more ?

I'm comfortable with it, but I was good with the original too.  I'm not in a hurry, let's see what @Orphis and @ChrisKitching say.  Once everyone is happy I'll file a PR for LZ4 with similar changes.

cmake_unofficial/CMakeLists.txt
```diff
@@ -1,40 +1,99 @@
...
+
+  if(${test_name})
+    set(CMAKE_C_FLAGS "${flag} ${CMAKE_C_FLAGS}")
+  endif()
+
+  unset(test_name)
+  unset(flag_to_test)
+endforeach (flag)
+
+if(NOT XXHASH_BUNDLED_MODE)
```
Orphis: If users don't want to install targets, they could just do `add_subdirectory(xxhash EXCLUDE_FROM_ALL)` and you don't need that extra logic then. I don't mind it staying that way though as it saves some confusion from users.

cmake_unofficial/CMakeLists.txt
```diff
@@ -1,40 +1,99 @@
...
+cmake_minimum_required (VERSION 2.8.12)
 
-# add the automatically determined parts of the RPATH
-# which point to directories outside the build tree to the install RPATH
-SET(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+# If XXHASH is being bundled in another project, we don't want to
+# install anything.  However, we want to let people override this, so
+# we'll use the XXHASH_BUNDLED_MODE variable to let them do that; just
+# set it to OFF in your project before you add_subdirectory(xxhash/contrib/cmake_unofficial).
+get_directory_property(XXHASH_PARENT_DIRECTORY PARENT_DIRECTORY)
```
Orphis: Can't you just check: `if(CMAKE_CURRENT_SOURCE_DIR STREQUAL "${CMAKE_SOURCE_DIR}")` instead?

nemequ: I seem to remember there was a reason that wouldn't work, but TBH I can't remember what it was.  That part has been copied around a lot, the reason may or may not apply to xxhash.

AFAIK it's not hurting anything, so I'd be more comfortable leaving it as-is, but if you feel strongly about it I'd be willing to try changing it (it could always be changed back if there are problems)…

cmake_unofficial/CMakeLists.txt
nemequ: EXCLUDE_FROM_ALL is a bit different.  It will actually prevent things from being built by default, so you don't get `xxhash`/`xxhsum`.  If any examples or tests are added in the future you wouldn't get those, either.  Bundled mode is more about making sure those targets aren't *installed*.

I prefer to always build utilities, examples, and tests; they don't take long (if this were a big C++ project I might feel different), provide a good sanity check, and can be nice to have around for developers.  People who disagree can always just use EXCLUDE_FROM_ALL.

Orphis: It will also apply to the install() rules, and running "make install" from the top level context won't install those automatically. It's possible to install them by running it from that folder though.

The end goal is usually to prevent installation of those targets from the top level. This solution is non intrusive, so it's usually better than having special code everywhere to check for bundled mode before all the install commands. Then, EXCLUDE_FROM_ALL is not a bad thing on those targets anyway. If nothing depends on them, there's no point building them when embedding it using the solution I suggested.

I'm not saying that you shouldn't check for bundled mode at all, I believe it's fine before test targets for example.

cmake_unofficial/CMakeLists.txt
Orphis: If you don't remember a reason and none has been documented, then do the change as it is much simpler to read and correct.

CMake is having a bad reputation because of things like that "I don't know why, I don't understand any of it, but it had to be done that way". Then people copy-paste that all the time and bad patterns propagate.

Cyan4973:
@nemequ : if you believe this patch is good to go, I'm willing to merge it.
The last issue is that, due to delay since last update, a conflict appeared within `CMakeLists.txt`.

Cyan4973:
Is this PR still alive ?

Cyan4973:
I don't feel in great position to tell if this patch is still valid or not, due to limited experience of `cmake` build system.
Therefore, without further comment from author, I guess my only remaining option will be to drop it.

nemequ:
Is anyone actually interested in it? I can take some time later this week to revisit it if there is interest, but I'm not actually using xxHash directly for anything; just thought I'd try to bring the build system up to parity with lz4 to make maintenance a bit easier…

Cyan4973:
I'm considering a new xxHash release soon, so if you plan on updating this patch, I can delay a bit to include it.

I know there are a few `cmake` users out there, since the `cmake` script regularly receives a few comments and patches, it's just I'm not one of them.

nemequ:
Okay, updated.

PR #98, opened 2017-01-04
Remove unused update on local variable p32

wuxb45:
The last p32++ is unnecessary. It's a minor fix.


Cyan4973:
You are right @wuxb45 .
It was just kept for symmetry and easy copy/pasting, but it's not necessary.

Note that the compiler already figures that this instruction is not useful : 
when compiling with or without your patch, the resulting checksum of the binary is the same.

PR #105, opened 2017-06-14
fix memory leak involving realloc()

faragon:


Cyan4973:
OK, so if `realloc()` fails, the initial memory block is still there and must be kept track of, in order to be able to `free()` it.

This is in essence what happens on line 700+ :
https://github.com/Cyan4973/xxHash/blob/dev/xxhsum.c#L699

The reason why it's not set up the same on line 682+ seems to be that
the initial condition to enter this section is that
`*lineBuf == NULL`.
Therefore, there is nothing to track. Could as well be a `malloc()`.
This would have been clearer.

But there is also a second condition, with `*lineMax<1`.
The intention seems to be a `free()`.
Here also, it would have been clearer to express it this way.

Moreover, `*lineMax` is an `int`. So it _could_ be negative.
And since `realloc()` expects a `size_t`, there is a silent cast that could spell trouble.
(Note : maybe that's impossible due to code construction, but that requires more thorough investigation).

Anyway, I think I'd better rewrite that part, for improved readability.

Cyan4973:
In d9e2c7bfd7237f83188b6517c62aafc52bb028aa,
I modified this part of the code, in a way which should make it safer to allocation errors.

faragon:
Thank you.

PR #108, opened 2017-07-29
Makefile: fix man symlinks

1ace:


Cyan4973:
Wow, a welcomed fix !

PR #121, opened 2017-12-26
fix a trivial typo

sjnam:
Hello,
I just fixed a very trivial typo, {32,64}-bits => {32,64}-bit.
And I made another luajit version of xxHash, https://github.com/sjnam/luajit-xxHash .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,7 +124,7 @@ test-mem: xxhsum
 
 .PHONY: test32
 test32: clean xxhsum32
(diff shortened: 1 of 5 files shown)


Cyan4973:
Thanks for notification and fix @sjnam !

sjnam:
Thank you for providing a wonderful hash algorithm.

My luajit implementation of xxHash, https://github.com/sjnam/luajit-xxHash, is not only XXH64 but also XXH32. :-)

PR #123, opened 2018-02-08
intel: do not use __attribute__((packed)) on Windows

mathstuf:
On Windows, the Intel compiler is closer to MSVC rather than GCC and
does not support the GCC attribute syntax.

---
See lz4/lz4#468.


pichi:
I don't get it, I don't see `XXH_FORCE_MEMORY_ACCESS` used anywhere in the codebase except definition.

mathstuf:
@pichi See [these lines](https://github.com/Cyan4973/xxHash/blob/dev/xxhash.c#L155-L179).

pichi:
@mathstuf: Thanks. I see. The problem was in way how I read Github's search result: https://github.com/Cyan4973/xxHash/search?utf8=✓&q=XXH_FORCE_MEMORY_ACCESS

PR #126, opened 2018-03-15
Fix reversed links for NVBIO and Factor

miike:
Previously the links for factorcode and NVBIO were switched around. This fixes that and neatens up the indentation.


Cyan4973:
Thanks @miike !

PR #128, opened 2018-03-18
Remove trailing whitespace with a check/test.

shlomif:
Their presence caused my tests to fail. I hereby put all my changes
under PD/CC0/Expat licence.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -194,11 +194,16 @@ preview-man: clean-man man
 	man ./xxhsum.1
 
 test-all: clean all namespaceTest test test32 test-xxhsum-c clean-xxhsum-c \
-	armtest clangtest gpptest c90test test-mem usan staticAnalyze
+	armtest clangtest gpptest c90test test-mem usan staticAnalyze \
+	trailingWhitespace
 
 .PHONY: listL120
 listL120:  # extract lines >= 120 characters in *.{c,h}, by Takayuki Matsuoka (note : $$, for Makefile compatibility)
 	find . -type f -name '*.c' -o -name '*.h' | while read -r filename; do awk 'length > 120 {print FILENAME "(" FNR "): " $$0}' $$filename; done
+
+.PHONY: trailingWhitespace
+trailingWhitespace:
+	! grep -E "`printf '[ \\t]$$'`" *.1 *.c *.h *.md LICENSE Makefile cmake_unofficial/*
 
 .PHONY: clean
 clean: clean-xxhsum-c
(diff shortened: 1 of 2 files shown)


Cyan4973:
If I do understand properly,
`trailingWhitespace` is just a test,
it will detect any trailing white space in the list of files passed in argument.

I'm wondering if `cmake_unofficial/*` is too wide.
This directory will also store cmake generated files,
including binaries like `libxxhash`.
I suspect that the `grep` script will not like it.

shlomif:
On Sun, 18 Mar 2018 18:39:58 +0000 (UTC)
Yann Collet <notifications@github.com> wrote:

> I'm wondering if `cmake_unofficial/*` is too wide.
> This directory will also store cmake generated files,
> including binaries like `libxxhash`.
> I suspect that the `grep` script will not like it.
> 

That should not be a concern because the recommended cmake practice is to do
out-of-tree builds (e.g: "cmake ../path-to-src/") so the directory will remain
clean of generated files.





-- 
-----------------------------------------------------------------
Shlomi Fish       http://www.shlomifish.org/
My Aphorisms - http://www.shlomifish.org/humour.html

Except for boastfulness, rashness is my only defect!
    — Based on http://www.shlomifish.org/humour/TheEnemy/ .

Please reply to list if it's a mailing list post - http://shlom.in/reply .

Cyan4973:
All right, I'll adapt the list

Cyan4973:
https://github.com/Cyan4973/xxHash#branch-policy

Cyan4973:
And, by the way : 
"Markdown uses two or more spaces at the end of a line to signify a line break."

So these white spaces are actually present on purpose.

shlomif:
On Mon, 19 Mar 2018 00:15:02 +0000 (UTC)
Yann Collet <notifications@github.com> wrote:

> And, by the way : 
> "Markdown uses two or more spaces at the end of a line to signify a line
> break."
> 
> So these white spaces are actually present on purpose.
> 

I see. That sounds like one huge misfeature of markdown.

-- 
-----------------------------------------------------------------
Shlomi Fish       http://www.shlomifish.org/
https://youtu.be/n6KAGqjdmsk - “Hurt Me Tomorrow”

If you put 7 of the most talented FOSS developers in a room for a week
and asked them to fix a bug in a spreadsheet program, in 1 week
you'd have 2 new mail readers and a text-based web browser.
    — via Bram Moolenaar

Please reply to list if it's a mailing list post - http://shlom.in/reply .

PR #152, opened 2018-10-04
Regards footnote.

sachin-kumar-singh:


Cyan4973:
Thanks for the suggestion.
Let's keep it for another repository...

PR #166, opened 2019-02-12
 Added export of public symbols on Windows. 

LambdAurora:
When I wanted to build xxHash on Windows with MSVC (Visual Studio 15 2017) and CMake, I got some linkage errors about xxhsum could not find the file xxhash.lib.

And it was normal because when MSVC build a shared library, it doesn't set every symbols to public like GCC. So the export/import declaration was missing for MSVC and it couldn't build the file xxhash.lib because for it there wasn't anything to export.

So, I added the missing export/import declaration for MSVC in `XXH_PUBLIC_API` and, for compilation in `shared` mode, the definition of `XXH_EXPORT` to tell xxHash to export the symbols or to import them when using the library.

### DIFF
--- a/cmake_unofficial/CMakeLists.txt
+++ b/cmake_unofficial/CMakeLists.txt
@@ -57,6 +57,7 @@ include_directories("${XXHASH_DIR}")
 
 # libxxhash
 add_library(xxhash "${XXHASH_DIR}/xxhash.c")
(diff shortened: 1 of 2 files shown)


Cyan4973:
Thanks @LambdAurora, it's a welcomed patch for Windows users !

PR #168, opened 2019-03-03
Improve aarch64 performance

easyaspi314:
Decided to move to a new PR.

OK, huge performance improvements found for aarch64.

Apparently, aarch64 **hates** shuffling with a passion and a pride. Even more than armv7a which wasn't so fond of it in the first place.

Using `vshrn`/`vmovn` is almost always faster than `vzip`/`vld2`. It also doesn't freak out GCC 6 so I can remove the warning.

```
./xxhsum 0.6.6 (64-bits little endian), by Yann Collet
Sample of 100 KB...
XXH32               :     102400 ->    38082 it/s ( 3718.9 MB/s)
XXH32 unaligned     :     102400 ->    37847 it/s ( 3696.0 MB/s)
XXH64               :     102400 ->    25979 it/s ( 2537.0 MB/s)
XXH64 unaligned     :     102400 ->    25971 it/s ( 2536.2 MB/s)
XXH3_64bits         :     102400 ->    66295 it/s ( 6474.2 MB/s)
XXH3_64b unaligned  :     102400 ->    65166 it/s ( 6363.8 MB/s)
```

Also, by the way, scalar isn't matching.
xxhash.c:
fe41ec4075ea7be1 -> vectorized
293cfafd94b5ab94 -> scalar

scalar does this for scramble:
```c
    for (i=0; i < (int)ACC_NB; i++) {
        int const left = 2*i;
        int const right= 2*i + 1;
        xacc[i] ^= xacc[i] >> 47;
        xacc[i] ^= PRIME64_5;

        {   U64 p1 = (xacc[i] >> 32) * xkey[left];
            U64 p2 = (xacc[i] & 0xFFFFFFFF) * xkey[right];
            xacc[i] = p1 ^ p2;
    }   }
```
while vector effectively does this
```c
    for (i=0; i < (int)ACC_NB; i++) {
        int const left = 2*i;
        int const right= 2*i + 1;
        xacc[i] ^= xacc[i] >> 47;
        xacc[i] ^= PRIME64_5;

        {   U64 p1 = (xacc[i] & 0xFFFFFFFF) * xkey[left];
            U64 p2 = (xacc[i] >> 32) * xkey[right];
            xacc[i] = p1 ^ p2;
    }   }
```

Which did you intend?

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -49,6 +49,11 @@
 #    define XXH_VECTOR XXH_AVX2
 #  elif defined(__SSE2__)
 #    define XXH_VECTOR XXH_SSE2
+/* GCC < 7 for aarch64 generates unreasonably slow code for the NEON
+ * implementation. We fall back to the scalar version and emit a warning. */
+#  elif defined(__aarch64__) && !defined(__clang__) && defined(__GNUC__) && __GNUC__ < 7
+#    warning Your GCC version has broken NEON support. Please use GCC 7+ or Clang.
+#    define XXH_VECTOR XXH_SCALAR
 /* msvc support maybe later */
 #  elif defined(__GNUC__) && (defined(__ARM_NEON__) || defined(__ARM_NEON))
 #    define XXH_VECTOR XXH_NEON
--- a/xxh3.h
+++ b/xxh3.h
@@ -49,11 +49,6 @@
 #    define XXH_VECTOR XXH_AVX2
 #  elif defined(__SSE2__)
 #    define XXH_VECTOR XXH_SSE2
-/* GCC < 7 for aarch64 generates unreasonably slow code for the NEON
- * implementation. We fall back to the scalar version and emit a warning. */
-#  elif defined(__aarch64__) && !defined(__clang__) && defined(__GNUC__) && __GNUC__ < 7
-#    warning Your GCC version has broken NEON support. Please use GCC 7+ or Clang.
-#    define XXH_VECTOR XXH_SCALAR
 /* msvc support maybe later */
 #  elif defined(__GNUC__) && (defined(__ARM_NEON__) || defined(__ARM_NEON))
 #    define XXH_VECTOR XXH_NEON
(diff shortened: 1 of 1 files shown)


easyaspi314:
I don't believe the test failure is my fault.

```
C:\projects\xxhash\xxhsum.exe 0.6.6 (64-bits little endian), by Yann Collet 
Sample of 100 KB...        
XXH32               :     102400 ->    65362 it/s ( 6383.0 MB/s) 
XXH32 unaligned     :     102400 ->    65362 it/s ( 6383.0 MB/s) 
XXH64               :     102400 ->   192000 it/s (18750.0 MB/s) 
XXH64 unaligned     :     102400 ->   192000 it/s (18750.0 MB/s) 
Assertion failed!

Program: C:\projects\xxhash\xxhsum.exe
File: xxhsum.c, Line 299

Expression: fastestH > 1./2000000000
make: *** [Makefile:134: check] Error 3
```
3fast5me? 🤔

Cyan4973:
Thanks @easyaspi314 !
The new patch works great !

> by the way, scalar isn't matching.

Thanks for the check !
Indeed, the intended version is the one which matches vector code.

> I don't believe the test failure is my fault.

You are right.
The issue seems to be a potential division by zero, which is prevented by an `assert()`.
This code should be updated to account for the possibility that initial trial runs faster than timer resolution.

Cyan4973:
I merged a bit too fast.

After testing performance of the new version on a Samsung S7 phone powered by Qualcomm ARM chip, I measured a pretty serious drop in performance. Not sure why that is, just it's quite pronounced on this chip.

```
XXH32             : 2.0 GB/s
XXH64             : 3.8 GB/s
XXH3 - Scalar     : 4.3 GB/s (-DXXH_VECTOR=0)
XXH3 - NEON (old) : 5.9 GB/s
XXH3 - NEON (new) : 4.4 GB/s
```

The new patch also produces an alignment warning.
I've reverted the patch for the time being (#169).

I suspect it can be hard to debug.
We may be in a situation where performance vary a lot depending on exact chip, despite belonging to the same ARM family. In which case, you don't experience the same on your side.

PR #171, opened 2019-03-07
better uint128_t multiply for 32-bit, MSVC multiply intrinsics, and fix XXH32 SSE4 bug

easyaspi314:
Inline assembly is the only thing that works on Clang to stop it from vectorizing XXH32 because it is literally impossible to do so. This lets us (well, not me…) use the AVX2 version of XXH3 without worrying about XXH32, and lets us remove the Makefile hack.

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -57,7 +57,14 @@
 #  endif
 #endif
 
-
+/* U64 XXH_mult32to64(U32 a, U64 b) { return (U64)a * (U64)b; } */
+#ifdef _MSC_VER
+#   include <intrin.h>
+    /* MSVC doesn't do a good job with the mull detection. */
+#   define XXH_mult32to64 __emulu
+#else
+#   define XXH_mult32to64(x, y) ((U64)((x) & 0xFFFFFFFF) * (U64)((y) & 0xFFFFFFFF))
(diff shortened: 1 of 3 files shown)


xxh3.h
```diff
@@ -84,12 +91,15 @@ ALIGN(64) static const U32 kKey[KEYSET_DEFAULT_SIZE] = {
...
+#if defined(__GNUC__) && defined(__i386__)
+/* GCC is stupid and tries to vectorize this.
+ * This tells GCC that it is wrong. */
+__attribute__((__target__("no-sse")))
+#endif
+static U64
 XXH3_mul128(U64 ll1, U64 ll2)
 {
-#if defined(__SIZEOF_INT128__) || (defined(_INTEGRAL_MAX_BITS) && _INTEGRAL_MAX_BITS >= 128)
+#if 0 && defined(__SIZEOF_INT128__) || (defined(_INTEGRAL_MAX_BITS) && _INTEGRAL_MAX_BITS >= 128)
```
Cyan4973: I assume the `0 &&` here was there for tests ?

easyaspi314: I have no idea what you are talking about.

PR #172, opened 2019-03-09
Add unroll pragma for Clang in XXH3_accumulate.

easyaspi314:
Clang doesn't unroll the XXH3_accumulate loop for some reason. Using `#pragma clang loop unroll(enable)` to hint to Clang that it should unroll results in a huge 1.4-1.5x speedup.

Before: 
```
XXH3_64bits         :     102400 ->   162910 it/s (15909.2 MB/s)
XXH3_64b unaligned  :     102400 ->   140762 it/s (13746.3 MB/s)
```
After:
```
XXH3_64bits         :     102400 ->   215992 it/s (21093.0 MB/s)
XXH3_64b unaligned  :     102400 ->   188791 it/s (18436.6 MB/s)
```

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -503,6 +503,11 @@ static void XXH3_scrambleAcc(void* acc, const void* key)
 static void XXH3_accumulate(U64* acc, const void* restrict data, const U32* restrict key, size_t nbStripes)
 {
     size_t n;
+
+/* Clang doesn't unroll this loop without the pragma. Unrolling results in code that is about 1.4x faster. */
+#if defined(__clang__) && !defined(__OPTIMIZE_SIZE__)
+#  pragma clang loop unroll(enable)
+#endif
     for (n = 0; n < nbStripes; n++ ) {
         XXH3_accumulate_512(acc, (const BYTE*)data + n*STRIPE_LEN, key);
         key += 2;
--- a/xxh3.h
+++ b/xxh3.h
@@ -197,9 +197,9 @@ XXH3_mul128(U64 ll1, U64 ll2)
     U32 const l2 = (U32)ll2;
 
     U64 const llh  = XXH_mult32to64(h1, h2);
-    U64 const llm1 = XXH_mult32to64(l1, h2;
-    U64 const llm2 = XXH_mult32to64(h1, l2;
-    U64 const lll  = XXH_mult32to64(l1, l2;
+    U64 const llm1 = XXH_mult32to64(l1, h2);
+    U64 const llm2 = XXH_mult32to64(h1, l2);
+    U64 const lll  = XXH_mult32to64(l1, l2);
 
     U64 const t = lll + (llm1 << 32);
     U64 const carry1 = t < lll;


Cyan4973:
Wow ! That's a great speed up ! Congratulation @easyaspi314 !

Also : note that I had to change the formula a little bit today.
Further analysis shows that UMAC is flawed for checksumming purpose,
but thankfully, it takes only a little modification to overcome this deficiency.

The new code is <5% slower, due to additional operation in the critical loop.
I haven't updated the ARM code, I know this part much less.

easyaspi314:
Okay…well that'll make things a lot more difficult on ARM. I would have to reshuffle, because ARM does this:
```c
uint32_t xdata[4] = { A, B, C, D };

uint32x2x2_t d = { { A, C }, { B, D } };
```

At this point, ARM is stuck with `d` shuffled.

I guess I could do `vshrn`/`vmovn` instead, as that might be the better option at this point, but as you saw in my last PR, that slows things down on non-Cortex-A57s.

Cyan4973:
Just FYI : 
the `#pragma clang loop unroll` slows things down on my ARM setup (Qualcom chip, clang 7.0, termux).
Non-turbo speed goes down from 5.4 GB/s to 5.1 GB/s.
Not a big issue, and not sure why.
I have no doubt it increases speed for you.
It just shows once again how hard it is to find optimizations that work positively for all ARM variants.

easyaspi314:
Hmm. Well, at this point, ARM is back to the drawing board. I haven't tested the unroll on ARM yet.

Side note: Scalar code looks like it will break if there is a carry.

Cyan4973:
oh, you mean, the `#pragma` speeds up `clang` for x64 ?

I haven't noticed any difference on my setup,
but then, `clang` on my laptop and desktop is really good, 
so I don't think there is any performance left on the table.
note: I'm using `clang 7.0`

I'm sorry the modification make the ARM code tricky.
This addition is necessary to guarantee that all input constituents will be present in the final mix.
Without it, there is a probability that a 4-byte segment get just ignored (it's nullified).
Even though it happens on average once every 16 GB, which may seem rare, that's still way too much for a 64+ bits hash, and for the purpose of checksumming.

Adding the input was the simplest additional operation I could find.
In spite of this, even on x86/x64, it can result in a visible drop, depending on CPU and compiler, since it's right in the critical loop. For comparison, I get up to 20% less bandwidth on my desktop, while the laptop is unaffected, and the ARM phone is <5% slower.

easyaspi314:
Well how does this do?
```c
_mm_add_epi32(xacc[i], _mm_shuffle_epi32(d, _MM_SHUFFLE(2, 3, 0, 1)));
```

Because I could definitely work with that, as I can just combine and use `umull2` on NEON. 

If not, I can just do `vshrn`/`vmovn`, but that means 6-7 more cycles in the main loop.

Cyan4973:
I tried something like that initially (though I xored, instead of add).
It would reduce performance by >30%, which was too impactful.

easyaspi314:
Before on my phone (it's not firing on all cores)
```
XXH3_64bits         :     102400 ->    80447 it/s ( 7856.2 MB/s)
XXH3_64b unaligned  :     102400 ->    80244 it/s ( 7836.4 MB/s)
```
After:
```
XXH3_64bits         :     102400 ->    59310 it/s ( 5792.0 MB/s)
XXH3_64b unaligned  :     102400 ->    59674 it/s ( 5827.5 MB/s)
```

easyaspi314:
I can get it a little faster on my phone (6.2 GB/s) with a different shuffling method and by adding beforehand, however, it becomes incredibly slow on the A57:
```
./xxhsum 0.6.6 (64-bits little endian), by Yann Collet
Sample of 100 KB...
XXH32               :     102400 ->    37991 it/s ( 3710.1 MB/s)
XXH32 unaligned     :     102400 ->    37819 it/s ( 3693.3 MB/s)
XXH64               :     102400 ->    25980 it/s ( 2537.1 MB/s)
XXH64 unaligned     :     102400 ->    25969 it/s ( 2536.0 MB/s)
XXH3_64bits         :     102400 ->    39652 it/s ( 3872.3 MB/s)
XXH3_64b unaligned  :     102400 ->    38851 it/s ( 3794.0 MB/s)
```

Fortunately, it just looks like terrible codegen from Clang 3.8. I am going to try cross compiling with newer versions of Clang and GCC.

easyaspi314:
Cross compiled in a Manjaro VM with both Clang 7.0.1 and aarch64-linux-gnu-gcc 8.2.0
```
~$ ./xxhsum.clang7 -b
./xxhsum.clang7 0.6.6 (64-bits little endian), by Yann Collet
Sample of 100 KB...
XXH32               :     102400 ->    38211 it/s ( 3731.5 MB/s)
XXH32 unaligned     :     102400 ->    37852 it/s ( 3696.5 MB/s)
XXH64               :     102400 ->    25979 it/s ( 2537.0 MB/s)
XXH64 unaligned     :     102400 ->    25974 it/s ( 2536.6 MB/s)
XXH3_64bits         :     102400 ->    50487 it/s ( 4930.4 MB/s)
XXH3_64b unaligned  :     102400 ->    49648 it/s ( 4848.4 MB/s)
~$ ./xxhsum.gcc8 -b
./xxhsum.gcc8 0.6.6 (64-bits little endian), by Yann Collet
Sample of 100 KB...
XXH32               :     102400 ->    38221 it/s ( 3732.5 MB/s)
XXH32 unaligned     :     102400 ->    37857 it/s ( 3696.9 MB/s)
XXH64               :     102400 ->    25994 it/s ( 2538.5 MB/s)
XXH64 unaligned     :     102400 ->    25986 it/s ( 2537.7 MB/s)
XXH3_64bits         :     102400 ->    39240 it/s ( 3832.1 MB/s)
XXH3_64b unaligned  :     102400 ->    39031 it/s ( 3811.6 MB/s)
```

5GB/s is much better.

Using `vshrn`/`vmovn` is faster like I saw before, but considering what happened before, that might not be the same in your case:
```
~$ ./xxhsum.clang7.shrn -b
./xxhsum.clang7.shrn 0.6.6 (64-bits little endian), by Yann Collet
Sample of 100 KB...
XXH32               :     102400 ->    38223 it/s ( 3732.7 MB/s)
XXH32 unaligned     :     102400 ->    37880 it/s ( 3699.2 MB/s)
XXH64               :     102400 ->    25981 it/s ( 2537.2 MB/s)
XXH64 unaligned     :     102400 ->    25976 it/s ( 2536.7 MB/s)
XXH3_64bits         :     102400 ->    57480 it/s ( 5613.2 MB/s)
XXH3_64b unaligned  :     102400 ->    55917 it/s ( 5460.6 MB/s)
~$ ./xxhsum.gcc8.shrn -b
./xxhsum.gcc8.shrn 0.6.6 (64-bits little endian), by Yann Collet
Sample of 100 KB...
XXH32               :     102400 ->    38197 it/s ( 3730.1 MB/s)
XXH32 unaligned     :     102400 ->    37838 it/s ( 3695.1 MB/s)
XXH64               :     102400 ->    25987 it/s ( 2537.8 MB/s)
XXH64 unaligned     :     102400 ->    25982 it/s ( 2537.3 MB/s)
XXH3_64bits         :     102400 ->    48220 it/s ( 4709.0 MB/s)
XXH3_64b unaligned  :     102400 ->    47879 it/s ( 4675.7 MB/s)
```

Also, I just realized I was doing a terrible life algorithm again.
Smart me: Use the vbox shared folder to copy my ssh key over and scp that from the vm.
Actual me: Build on manjaro, copy to the vbox shared folder on my Mac, and scp from there. 🤦‍♂️ 

I put the changes in the branch called `xxh3-aarch64-test`. If you define `ALT`, it will use the `vzip` method, otherwise it will use the `vshrn`/`vmovn` method. Could you please test it on your S7?

Cyan4973:
Here are some results from S7 with `clang7` under `termux` : 
```
XXH3_64bits() : 
before : 5.7 GB/s
after, without ALT : 4.8 GB/s
after, with -DALT : 4.8 GB/s
```
I couldn't measure a significant difference between the two variants.

For comparison, `XXH64` is at `3.8 GB/s` on this system.

easyaspi314:
I checked the assembly output on ARMv7 from Clang and I noticed that was making copies before the vzip. I modify it in place with an inline assembly string and I get ~7.2. 

Me:

enough

Cyan4973:
I've started adding some self-tests within `xxhsum`.
They are triggered every time the benchmark command `-b` is invoked.

These hash values are validated vs the `SSE2` path, which is the most important one.
They are supposed to help us find potential discrepancies between targets and code paths (AVX2, NEON and scalar).

easyaspi314:
Side note: The scalar version bus errors on ARMv7. Alignment memes are still an issue on ARM, mainly with the `ldrd` instruction. I switch to `XXH_read32/64` and it works

easyaspi314:
This is definitely the best code path for ARMv7 Clang. GCC for ARM hates everything though.

```c

#elif (XXH_VECTOR == XXH_NEON)

    assert(((size_t)acc) & 15 == 0);
    {                 uint64x2_t* const xacc  = (uint64x2_t *)acc;
                  const uint32_t* const xdata = (const uint32_t *)data;
        ALIGN(16) const uint32_t* const xkey  = (const uint32_t *)key;

        size_t i;
        for (i=0; i < STRIPE_LEN / sizeof(uint64x2_t); i++) {
            uint32x4_t d = vld1q_u32(xdata + i * 4);
            uint32x4_t const k = vld1q_u32(xkey + i * 4);
            uint32x4_t dk = vaddq_u32(d, k);   /* uint32 dk[4]  = {d0+k0, d1+k1, d2+k2, d3+k3} */
#if !defined(__aarch64__) && !defined(__arm64__)
            /* vzip on ARMv7 Clang generates a lot of vmovs without this. */
            __asm__("vzip.32 %e[dk], %f[dk]" : [dk] "+w" (dk));
            xacc[i] = vaddq_u64(xacc[i], vreinterpretq_u64_u32(d));
            xacc[i] = vmlal_u32(xacc[i], vget_low_u32(dk), vget_high_u32(dk));                /* uint64 res[2] = {dk0*dk1,dk2*dk3} */
#else
            uint32x2x2_t zipped = vzip_u32(vget_low_u32(dk), vget_high_u32(dk));
            xacc[i] = vaddq_u64(xacc[i], vreinterpretq_u64_u32(d));
            /* xacc must be aligned on 16 bytes boundaries */
            xacc[i] = vmlal_u32(xacc[i], zipped.val[0], zipped.val[1]);                /* uint64 res[2] = {dk0*dk1,dk2*dk3} */
#endif
        }
    }


#elif (XXH_VECTOR == XXH_NEON)

    assert(((size_t)acc) & 15 == 0);
    {       uint64x2_t* const xacc =       (uint64x2_t*) acc;
        const uint32_t* const xkey  = (const uint32_t *) key;
        size_t i;
        for (i=0; i < STRIPE_LEN/sizeof(uint64x2_t); i++) {
            uint64x2_t data = xacc[i];
            uint64x2_t const shifted = vshrq_n_u64(data, 47);
            data = veorq_u64(data, shifted);

            {
                uint32x2x2_t const k = vld2_u32 (xkey+i*4);              /* load and swap */
                /* shuffle: 0, 1, 2, 3 -> 0, 2, 1, 3 */
                uint32x2x2_t const d =
                    vzip_u32(
                        vget_low_u32(vreinterpretq_u32_u64(data)),
                        vget_high_u32(vreinterpretq_u32_u64(data))
                    );
                uint64x2_t const dk  = vmull_u32(d.val[0],k.val[0]);     /* U64 dk[2]  = {d0 * k0, d2 * k2} */
                uint64x2_t const dk2 = vmull_u32(d.val[1],k.val[1]);     /* U64 dk2[2] = {d1 * k1, d3 * k3} */
                xacc[i] = veorq_u64(dk, dk2);                            /* xacc[i] = dk ^ dk2;             */
        }   }
    }

```

easyaspi314:
On my phone, it gets 8.8 GB/s with Clang and 3.9 GB/s with GCC.

Like seriously, GCC and Clang for the x86 family are very competitive, but GCC for the ARM family is useless.

Cyan4973:
Even on `x64`, the differences between `gcc` and `clang` on vectorial code are present. Especially on the `AVX2` path, the performance `gcc` pales in comparison to `clang`.
So I guess `clang` is simply way better at this game.

As a side effect, `clang` vectorizing `XXH32` is actually a bad choice (on `x86`/`x64`). `clang` seems to trust its vectorization capabilities a bit too much at times.

Also : I know the current code is not yet correct regarding alignment and endianess. The next stage is to improve automated testings in this regard, so that the flaws get exposed, and then corrected. I expect `qemu` to be useful for this objective.

easyaspi314:
I have access to mips64, ppc64le, ppc64, sun4u, and sparc64 in the GCC compiler farm. You might want to make an account because they'll let you use it for testing and building as long as it is open source software. https://cfarm.tetaneutral.net/

Additionally, while SSE slows down XXH32 and XXH64 on x86_64. with NEON intrinsics, my G3 loves it and is about 2 times faster (2.2 GB/s->4.1 GB/s) on XXH32 and about 3 times faster (0.8 GB/s->2.4 GB/s) on XXH64. The A57 doesn't get a huge difference though, mainly because of the native (albeit slow) 64-bit integers and the higher NEON latency.

Cyan4973:
I've added ARM dedicated tests on `travis CI`.
They currently fail, as expected.
https://travis-ci.org/Cyan4973/xxHash/jobs/505405171

Since these tests are now part of the continuous integration plan,
this will make it possible to detect bugs, check they are fixed and ensure they remain fixed.

I can take of the warnings on the `scalar` code path,
and I will likely need your help to fix the `NEON` code path.

easyaspi314:
OK, so the ARM32 NEON build is failing because arm-none-eabi-gcc defaults to the ARMv4t chip, aka the ARM7TDMI from the 90s, which, needless to say, does not support NEON. It doesn't even have an FPU. In order to enable NEON, use this:
```
-march=armv7-a   # target ARMv7a
-mfloat-abi=hard # hardware floating point instead of software
-mfpu=neon       # NEON FPU
```
I am about to push a fixed xxh3 once I finish testing and commenting.

Cyan4973:
FYI, `PowerPC` tests have been added too,
it made it possible to fix remaining issues with regards to endianess.

Now time to focus on ARM...

Cyan4973:
OK, some progresses,

using the 3 additional flags you mentioned, compilation runs farther with `arm-linux-gnueabi-gcc`. It reaches link stage, where it fails with : 
> error: xxhsum.o uses VFP register arguments, xxhsum does not

Compilation with `NEON` is not a problem for `aarch64`, `aarch64-linux-gnueabi-gcc` does not need any additional flags to properly compile and run `xxhsum`. Internal tests then fail, with wrong hash value, but this one is expected.

easyaspi314:
Try `-mfloat-abi=softfp` too, that might work.

Also, I am looking at the XXH3 code and it looks like the seed is returned verbatim on a zero length hash. Isn't that kinda bad?

Cyan4973:
> the seed is returned verbatim on a zero length hash. 

This is intentional.
With the default `seed == 0`, it returns `0` on `len == 0`.
This used to be a requested property for `XXH32`/`XXH64` (which cannot be retrofitted to fit that request).

That being said, you could argue that a user providing a non-zero `seed` would prefer to receive some mixed bits, rather than plain `seed`. That would be fair. I have not deeply thought about this, and merely added `seed` so that the result would be non-constant with different seeds. This kind of mimic what happens with a `crc` when presented with a `seed` (effectively an accumulator) and a zero-length input : it just returns the `seed`.

I'm completely opened here. `seed` could be replaced by an `avalanche(seed)` instead, mixing the bits.

PR #173, opened 2019-03-13
Update ARM NEON code

easyaspi314:
The NEON algorithms have now been updated to match the SSE2 algorithm.

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -208,17 +208,32 @@ XXH3_mul128(U64 ll1, U64 ll2)
     U32 v[2] = { (U32)(ll2 >> 32), (U32)ll2 };
     U32 k;
 
+    /* U64 t = (U64)u[1] * (U64)v[1];
+     * w[3] = t & 0xFFFFFFFF;
+     * k = t >> 32; */
     __asm__("umull %0, %1, %2, %3"
             : "=r" (w[3]), "=r" (k)
             : "r" (u[1]), "r" (v[1]));
+
+    /* t = (U64)u[0] * (U64)v[1] + w[2] + k;
+     * w[2] = t & 0xFFFFFFFF;
+     * k = t >> 32; */
     __asm__("umaal %0, %1, %2, %3"
             : "+r" (w[2]), "+r" (k)
             : "r" (u[0]), "r" (v[1]));
     w[1] = k;
     k = 0;
+
+    /* t = (U64)u[1] * (U64)v[0] + w[2] + k;
+     * w[2] = t & 0xFFFFFFFF;
+     * k = t >> 32; */
     __asm__("umaal %0, %1, %2, %3"
             : "+r" (w[2]), "+r" (k)
             : "r" (u[1]), "r" (v[0]));
+
+    /* t = (U64)u[0] * (U64)v[0] + w[1] + k;
+     * w[1] = t & 0xFFFFFFFF;
+     * k = t >> 32; */
     __asm__("umaal %0, %1, %2, %3"
             : "+r" (w[1]), "+r" (k)
             : "r" (u[0]), "r" (v[0]));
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks for this excellent `NEON` support @easyaspi314 !

Cyan4973:
As a "continuation quest", it would be great if it was possible to enable `NEON` automated tests on Travis CI. So far my attempts [have not been successful](https://travis-ci.org/Cyan4973/xxHash/jobs/505548115). I'm especially surprised for `aarch64`, which works fine on my VM, but not on Travis ...

easyaspi314:
OK, I see the issue. It has nothing to do with the code, it has to do with Travis's GCC version which is old af.

**GCC 4.8 disables NEON support by default because it is buggy.**

```
$ aarch64-linux-gnu-gcc-4.8 -E -xc /dev/null -dM | grep NEON
$
```

One of these should work.
```
aarch64-linux-gnu-gcc -D__ARM_NEON
clang --target=aarch64-linux-gnu -march=armv8.1-a # (probably preferred because Clang 3.8 is decent)
```

Cyan4973:
For information, benchmark results on my S7+termux+clang7 platform,

without "turbo" mode,
I got 5 GB/s with this patch on `XXH3_64bits()` using `NEON` code path.
This compares favorably to 3.8 GB/s for `XXH64()`,
and 3.7 GB/s for `XXH3` using the scalar code path.

With "turbo" mode, `XXH3_64bits()` with `NEON` reaches 6.1 GB/s, 
it compares to 4.6 GB/s for `XXH64()`,
and 4.7 GB/s for `XXH3` using the scalar code path.

It also passes all consistency tests.

Cyan4973:
Thanks for diagnosis @easyaspi314 ,

on Travis, I updated the `yml` script for ARM tests to use the newer `Xenial` VM.
It solves the issue for `aarch64`, which now [compiles and run correctly the `NEON` code path](https://travis-ci.org/Cyan4973/xxHash/jobs/505893894#L470).

Now, only the regular ARM 32-bit `NEON` test remains to be solved.

Cyan4973:
Also, for information :
The plan now is to release a new version of xxHash, `v0.7.0`,
including this new hash,
still labelled "experimental", and protected behind a macro `XXH_STATIC_LINKING_ONLY`, because it's not stable yet.
The scope is also limited to one-short functions, though [it includes both 64 and 128 bits](https://github.com/Cyan4973/xxHash/blob/xxh3/xxhash.h#L328) variants.

This will give time to collect feedback from interested users.
There are many open questions, and I will try to list them now in code comments.

A future version will complete the scope, and move `XXH3` into stable.
Complete scope must include at least streaming hash and canonical representation.

Optional future features are custom secret keys, and secret key generation.

Cyan4973:
I haven't found a good solution for 32-bit ARM and `NEON`.

I'm currently suspecting an issue regarding a mix of `hardfp` and `softfp`, maybe due to dependencies on C runtime. But that does not give me a solution.

For the time being, I will therefore disable the ARM 32-bit + `NEON` test.
`NEON` code path is validated through `aarch64`, so we are not completely blind regarding its conformance.

easyaspi314:
Ok. 
Also, I did some testing with Clang's ARMv7a assembly output, and if I use the normal path without the inline assembly and remove the excess `vmov` instructions by hand, I get 8.9 GB/s. The inline assembly has an unfortunate side effect of messing with Clang's interleaving, which is why the second `vzip` didn't benefit from the asm, as the serialization wasn't good for pipelining.

Cyan4973:
I'm going to prepare a blog article, to present `XXH3` when it's released.
I believe it would be appropriate to mention your contribution, at least for the `NEON` code path.
I can certainly use your pseudo `@easyaspi314`, but I could also use any other name of your preference if you wish, be it your real name or a different pseudo of your choice.

easyaspi314:
I usually go by easyaspi314, but my full name is Devin Hussey. Probably put both.

PR #174, opened 2019-03-13
XXH3

Cyan4973:
Introduces a new hash algorithm, named `XXH3`,
able to generate 64 and 128-bit outputs
(and, if needed, 256-bit in the future).

`XXH3` is also a lot faster than `XXH32` and `XXH64`, both for long and (especially) short inputs.

The current version is labelled "experimental", in order to collect feedback from interested users.
It will be stabilized in a future version.


Bramzor:
Finally! Thanks for the work. Now it will take some time before we are able to use it with different languages. But in my case, I hope it will reach javascript soon.

easyaspi314:
Good luck with JavaScript. There is a lot of 64-bit arithmetic here, and a lot of the performance comes from SIMD. You might be fine with WASM, though.

I am trying to write a cleaned up reference implementation. I might create a separate repo of these reference implementations. 

Next, I think I'll work on porting the non-XXH32a/XXH64a changes from my project (such as wording fixes in xxhsum, macro endian checking, improved alignment handling) into the XXH3 branch. 

Last, I think I am probably going to waste a few hours of my life making a joke Bash implementation to match [my other joke implementations](https://github.com/easyaspi314/xxbash). Honestly, it wouldn't be that bad if it weren't for Bash's complete lack of a logical shift right.

Cyan4973:
Something I should probably mention somewhere :
`XXH3` is actually friendly to 32-bit platforms.

It's true that most modern hashes work on 64-bit registers, because it makes data ingestion twice faster.
Then, 64-bit multiplications are pretty powerful to mix bits
Emulating all these instructions on 32-bit platforms works but is damn slow.
Hence, most of these hashes pay it with a huge performance penalty on 32-bit.

`XXH3` loses much less performance when moving to 32-bit. In my tests, it's actually even competitive with `XXH32`, so far the "king of the hill" for speed on 32-bit platforms.

It's not longer necessary to sacrifice one platform to fare better in the other one. That's a nice property, especially for checksumming, as the nb of platforms that are still 32-bit is surprisingly large ! `WASM` is indeed one of them.

easyaspi314:
**Why XXH3 is very good for most 32-bit machines:** 

Unlike XXH64 which took full advantage of a 64-bit instruction set, XXH3 is much more conservative with the 64-bit instructions, mostly relying on these instructions which only take one or two instructions on 32-bit:
- 32->64 long unsigned multiply: `mul` (1 op, dword), `umull` 
- add: `add` + `adc`, `umlal` on ARM when combined with above
- xor: 2x `xor`/`eor`, with some xors only being required once.
- `x << 32`, `x >> 32`, `x & 0xFFFFFFFF`, `x | ((U64)y << 32)`: *literally free*
- `x ^ (x >> 47)`: `srl`+`xor`, `lsr`+`eor`:

Note that this:
```c
   x ^ (x >> 47)
```
is easy to calculate despite being a shift (which is normally expensive), because since the shift is larger than 32 bits, the lower bits of x will be zero in the shift. Because of this, instead of shifting by 47, it only has to shift the high bits of x by 15 (47 - 32). Changing it to, say, 31 where the low bits are used makes it significantly more complicated.

This is in comparison to XXH64 which did the following:
- add: `add` + `adc`
- 64->64 multiply: x86: `imul`, `mov`, `imul`, `mov`, `add`, `mul`, `add`, ARM: `umull`, `mla`, `mla`
- XXH_rotl64(31): x86: `shld` * 2, ARM: `lsl`, `orr`, `lsl`, `orr` 

As a result, the XXH64 main loop without SIMD requires 120 instructions for x86 and 82 instructions for ARMv7, both with abundant swapping to the stack.

While SSE2 and NEON can help with XXH64 on 32-bit, the implementation is littered with literally two entirely different multiply routines on NEON, a bunch of hacks on x86, and all that works makes it only about half the speed of XXH32. (There is a SSE4.1 method which is faster than XXH32 on the Nehalem and Sandy/Ivy Bridge families only, but the slower performance of `pmulld` on literally every other Intel chip combined with the increased requirement of SSE4.1 makes it very niche).

Not only are all of these instructions very simple for 32-bit processors, they are all readily available in the widely supported SSE2 and NEON instruction sets, which are supported by basically all Android and iOS devices, Raspberry Pi 2+, and all x86 processors since the Pentium 4.

Cyan4973:
Note : I noticed a very large difference in performance between `clang` and `gcc` when compiling in 32-bit mode.

Without any `SSE2` instruction, I can see `xxh3` running faster than `XXH32` on `clang`.
The result is not so good with `gcc`.

easyaspi314:
Well apparently, from what I am seeing, just like on ARM Thumb, GCC seems to be going ham with the converting multiplies to shifts+adds, resulting in a really pathetic XXH32:
```asm
XXH32_round:
    lea    ecx, [edx + edx * 4]
    lea    ecx, [ecx + ecx * 4]
    lea    ecx, [ecx + ecx * 4]
    lea    ecx, [ecx + ecx * 4]
    lea    ecx, [ecx + ecx * 4]
    lea    ecx, [edx + ecx * 8]
    add    ecx, ecx
    add    ecx, edx
    add    ecx, ecx
    add    ecx, edx
    lea    ecx, [ecx + ecx * 4]
    lea    ecx, [edx + ecx * 4]
    lea    ecx, [edx + ecx * 4]
    add    ecx, ecx
    add    ecx, edx
    shl    ecx, 0x4
    add    ecx, edx
    lea    edx, [edx + ecx * 8]
    sub    eax, edx
    rol    eax, 0xd
    mov    edx, eax
    lea    eax, [eax + eax * 4]
    lea    eax, [eax + eax * 4]
    lea    eax, [eax + eax * 4]
    add    eax, eax
    add    eax, edx
    lea    eax, [eax + eax * 8]
    lea    eax, [eax + eax * 8]
    lea    eax, [edx + eax * 8]
    lea    ecx, [edx + eax * 4]
    mov    eax, ecx
    shl    eax, 0x8
    sub    eax, ecx
    shl    eax, 0x4
    add    eax, edx
    ret   
```
Yes, that is XXH32_round with `-fno-inline`

```
./xxhsum 0.7.0 (32-bits little endian), by Yann Collet
Sample of 100 KB...
XXH32               :     102400 ->     6118 it/s (  597.5 MB/s)
XXH32 unaligned     :     102400 ->     6082 it/s (  593.9 MB/s)
XXH64               :     102400 ->    16055 it/s ( 1567.8 MB/s)
XXH64 unaligned     :     102400 ->    15954 it/s ( 1558.0 MB/s)
XXH3_64bits         :     102400 ->    14544 it/s ( 1420.3 MB/s)
XXH3_64b unaligned  :     102400 ->    12461 it/s ( 1216.9 MB/s)
```

easyaspi314:
Benchmarks for targeting `-march=i686 -mtune=generic -m32 -mno-sse3 -msse2`:
gcc:
```
./xxhsum 0.7.0 (32-bits little endian), by Yann Collet
Sample of 100 KB...
XXH32               :     102400 ->    52336 it/s ( 5110.9 MB/s)
XXH32 unaligned     :     102400 ->    52266 it/s ( 5104.1 MB/s)
XXH64               :     102400 ->    16734 it/s ( 1634.2 MB/s)
XXH64 unaligned     :     102400 ->    16529 it/s ( 1614.1 MB/s)
XXH3_64bits         :     102400 ->   117921 it/s (11515.7 MB/s)
XXH3_64b unaligned  :     102400 ->    95508 it/s ( 9326.9 MB/s)
```
clang:
```
./xxhsum 0.7.0 (32-bits little endian), by Yann Collet
Sample of 100 KB...
XXH32               :     102400 ->    37148 it/s ( 3627.7 MB/s)
XXH32 unaligned     :     102400 ->    36675 it/s ( 3581.5 MB/s)
XXH64               :     102400 ->    22006 it/s ( 2149.0 MB/s)
XXH64 unaligned     :     102400 ->    21748 it/s ( 2123.8 MB/s)
XXH3_64bits         :     102400 ->   122640 it/s (11976.5 MB/s)
XXH3_64b unaligned  :     102400 ->   111069 it/s (10846.6 MB/s)
```
Clang spills the primes for some reason, resulting in the slowdown.

easyaspi314:
I made a PR that will improve the output messages in xxhsum.

Cyan4973:
`XXH3` is now released as part of xxHash v0.7.0.

cup:
@easyaspi314 logical shift right is only different with negative numbers - is that a problem in this case?

easyaspi314:
But bash always uses signed arithmetic, and when an "unsigned" number is greater than 0x7FFFFFFFFFFFFFFF, a.k.a it has the sign bit set, it causes issues.

cup:
the number you gave is max for signed 64-bit integers

why would bash need to go higher than that? do you have a practical use case?

easyaspi314:
Yeah, when writing a s****y XXH64 implementation in Bash which actually relies on overflowing arithmetic. :joy:

Nearly every other language has some form of a logical shift right, even when there are no unsigned integers.

Examples include Java and JavaScript, which supply a `>>>` operator. Bash is one of the only languages I have seen that doesn't have one.

It can be added in 22 lines, and I think I am going to submit a patch for the yucks, because why not.

PR #176, opened 2019-03-15
Improve xxhsum output message quality

easyaspi314:
- xxhsum now prints more professional-looking error messages:

Before:
```
    Pb opening foo
```
After:
```
    Error: Could not open 'foo': No such file or directory.
```

- xxhsum will now attempt to display the architecture and the compiler version in the benchmark WELCOME_MESSAGE.
  It detects the following compilers:
    - Clang
    - GCC
    - Intel Compiler
    - MSVC
    - tcc  

  and it should detect the following architectures:
    - x86 (+SSE2/AVX/AVX2)
    - x86_64 (+SSE2/AVX/AVX2)
    - ARM (+NEON)
    - aarch64
    - PowerPC 64
    - PowerPC
    - AVR
    - MIPS 64
    - MIPS

Before:
```
    ./xxhsum 0.7.0 (64-bits little endian), by Yann Collet
```
After:
```
    ./xxhsum 0.7.0 (64-bits x86_64 + SSE2 little endian), GCC 8.3.0, by Yann Collet
```
- Sanity checks are consistent now and give better warning messages:

Before:
```
    ERROR : Test  1 : 0x12345678 <> 0x02CC5D05   !!!!!
```
```
    ERROR : Test  1 : 64-bit values non equals   !!!!!
     0x1234567890ABCDEFULL != 0xEF46DB3751D8E999ULL
```
After:
```
    Error: 32-bit hash test 1: Internal sanity check failed!
    Got 0x12345678, expected 0x02CC5D05.
    or temporarily comment out the tests in BMK_sanityCheck.
```
...and the 64-bit and 128-bit messages now match. I eventually want to name the
tests instead of just using the test number, but this is still better than before.

- xxhsum now displays `stdin` instead of `-` when reading from stdin.

### DIFF
--- a/xxhsum.c
+++ b/xxhsum.c
@@ -44,7 +44,6 @@
 #  define _LARGEFILE64_SOURCE
 #endif
 
-
 /* ************************************
  *  Includes
  **************************************/
@@ -55,6 +54,7 @@
 #include <sys/stat.h>   /* stat, stat64, _stat64 */
 #include <time.h>       /* clock_t, clock, CLOCKS_PER_SEC */
 #include <assert.h>     /* assert */
+#include <errno.h>      /* errno */
 
 #define XXH_STATIC_LINKING_ONLY   /* *_state_t */
 #include "xxhash.h"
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks @easyaspi314 !
It's a great and helpful patch !

easyaspi314:
Oops, I literally just pushed another commit to this to detect Emscripten, wasm, and asm.js. Could you also add that?

Cyan4973:
arf, sorry !

`xxHash v0.7.0` has just been released, so it won't be part of the release. 
But it can still be merged.
I can take your PR into `dev` directly now, it's no longer necessary to separate `XXH3` into is own feature branch.

Issue #179, opened 2019-03-16
XXH3 doesn't compile with MSVC


mgieseki:
First of all thanks for adding XXH3. It's an awesome addition to xxHash. While the recent sources build fine with MinGW, they failed to compile with MSVC (version19.16.27027, Visual Studio 2017 15.9.9) due to some unsupported `#pragma` and `__attribute__` constructs.
Here's the compiler output (sorry, I've only the German locale installed at the moment):
```
xxh3.h(148): error C2156: Pragma muss außerhalb der Funktion liegen
xxh3.h(616): error C2143: Syntaxfehler: Es fehlt ")" vor "("
xxh3.h(616): error C2059: Syntaxfehler: ")"
xxh3.h(617): error C2143: Syntaxfehler: Es fehlt ")" vor "Typ"
xxh3.h(617): error C2091: Funktionsergebnis ist eine Funktion
xxh3.h(618): error C2085: "XXH3_hashLong_64b": Nicht in der formalen Parameterliste enthalten
xxh3.h(618): error C2143: Syntaxfehler: Es fehlt ";" vor "{"
xxh3.h(619): error C2065: "seed": nichtdeklarierter Bezeichner
xxh3.h(621): error C2065: "data": nichtdeklarierter Bezeichner
xxh3.h(621): warning C4022: "XXH3_hashLong": Zeigertyp für übergebenen Parameter 2 stimmt nicht mit dem deklarierten Zeigertyp überein
xxh3.h(621): error C2065: "len": nichtdeklarierter Bezeichner
xxh3.h(625): error C2065: "len": nichtdeklarierter Bezeichner
xxh3.h(625): warning C4047: "return": Anzahl der Dereferenzierungen bei "int (__cdecl *)()" und "XXH64_hash_t" unterschiedlich
xxh3.h(741): error C2143: Syntaxfehler: Es fehlt ")" vor "("
xxh3.h(741): error C2059: Syntaxfehler: ")"
xxh3.h(742): error C2143: Syntaxfehler: Es fehlt ")" vor "Typ"
xxh3.h(742): error C2091: Funktionsergebnis ist eine Funktion
xxh3.h(742): error C2084: Funktion "int (__cdecl *__attribute__())()" hat bereits einen Funktionsrumpf
xxh3.h(616): note: Siehe vorherige Definition von "__attribute__"
xxh3.h(743): error C2085: "XXH3_hashLong_128b": Nicht in der formalen Parameterliste enthalten
[...]

PR #181, opened 2019-03-16
Fixed the 'dllimport' build issue with MSVC.

LambdAurora:
Possible fix for #179 !


Cyan4973:
What about adding a `XXH_IMPORT` statement instead ?
This would require the library to set `XXH_IMPORT` or `XXH_EXPORT` in order to publish its symbols.
But a user importing the file directly into its project would not need to set anything,
which seems the wanted behavior.

LambdAurora:
Made the changes for that wanted behavior, and forks fine with MSVC!

`dllimport` is not defined by default now.

PR #185, opened 2019-03-19
dead code fixed, further improvements proposed

czMaxx:
tested on MSVC and GCC, x86 and x86_64

test it on ARM if you want/can

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -295,7 +295,7 @@ XXH3_len_1to3_64b(const void* data, size_t len, const void* keyPtr, XXH64_hash_t
         BYTE const c3 = ((const BYTE*)data)[len - 1];
         U32  const l1 = (U32)(c1) + ((U32)(c2) << 8);
         U32  const l2 = (U32)(len) + ((U32)(c3) << 2);
-        U64  const ll11 = XXH_mult32to64((l1 + seed + key32[0]), (l2 + key32[1]));
+        U64  const ll11 = XXH_mult32to64((l1 + seed + key32[0]), (l2 + key32[1])); // fixme: MSVC warns that seed is truncated to 32bits
         return XXH3_avalanche(ll11);
     }
 }
@@ -540,7 +540,7 @@ static void XXH3_scrambleAcc(void* acc, const void* key)
         xacc[i] ^= xacc[i] >> 47;
 
         {   U64 const p1 = XXH_mult32to64(xacc[i] & 0xFFFFFFFF, xkey[left]);
-            U64 const p2 = XXH_mult32to64(xacc[i] >> 32,        xkey[right]);
+            U64 const p2 = XXH_mult32to64(xacc[i] >> 32,        xkey[right]); // warning C5045, Spectre related - may decrease the speed with some circumstances (the /Qspectre option may be turned on by default and a fence instruction may appear = UD on older processors)
             xacc[i] = p1 ^ p2;
     }   }
 
@@ -619,7 +619,7 @@ static XXH64_hash_t XXH3_mergeAccs(const U64* acc, const U32* key, U64 start)
 XXH_NO_INLINE XXH64_hash_t    /* It's important for performance that XXH3_hashLong is not inlined. Not sure why (uop cache maybe ?), but difference is large and easily measurable */
 XXH3_hashLong_64b(const void* data, size_t len, XXH64_hash_t seed)
 {
-    ALIGN(64) U64 acc[ACC_NB] = { seed, PRIME64_1, PRIME64_2, PRIME64_3, PRIME64_4, PRIME64_5, (U64)0 - seed, 0 };
+    ALIGN(64) U64 acc[ACC_NB] = { seed, PRIME64_1, PRIME64_2, PRIME64_3, PRIME64_4, PRIME64_5, (U64)0 - seed, 0 }; // fixme: MSVC warns that there's a non-constant initializer
 
     XXH3_hashLong(acc, data, len);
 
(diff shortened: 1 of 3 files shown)


xxh3.h
```diff
@@ -295,7 +295,7 @@ XXH3_len_1to3_64b(const void* data, size_t len, const void* keyPtr, XXH64_hash_t
         BYTE const c3 = ((const BYTE*)data)[len - 1];
         U32  const l1 = (U32)(c1) + ((U32)(c2) << 8);
         U32  const l2 = (U32)(len) + ((U32)(c3) << 2);
-        U64  const ll11 = XXH_mult32to64((l1 + seed + key32[0]), (l2 + key32[1]));
+        U64  const ll11 = XXH_mult32to64((l1 + seed + key32[0]), (l2 + key32[1])); // fixme: MSVC warns that seed is truncated to 32bits
```
easyaspi314: We should change XXH_mult32to64 to this then:
```c

/* U64 XXH_mult32to64(U32 a, U32 b) { return (U64)a * (U64)b; } */
#ifdef _MSC_VER
#   include <intrin.h>
    /* MSVC doesn't do a good job with the mull detection. */
#   define XXH_mult32to64(x, y) __emulu((U32)(x), (U32)(y))
#else
#   define XXH_mult32to64(x, y) ((U64)((x) & 0xFFFFFFFF) * (U64)((y) & 0xFFFFFFFF))
#endif
```

xxh3.h
```diff
@@ -540,7 +540,7 @@ static void XXH3_scrambleAcc(void* acc, const void* key)
         xacc[i] ^= xacc[i] >> 47;
 
         {   U64 const p1 = XXH_mult32to64(xacc[i] & 0xFFFFFFFF, xkey[left]);
-            U64 const p2 = XXH_mult32to64(xacc[i] >> 32,        xkey[right]);
+            U64 const p2 = XXH_mult32to64(xacc[i] >> 32,        xkey[right]); // warning C5045, Spectre related - may decrease the speed with some circumstances (the /Qspectre option may be turned on by default and a fence instruction may appear = UD on older processors)
```
easyaspi314: Not sure if we can do much about that.

xxh3.h
```diff
@@ -619,7 +619,7 @@ static XXH64_hash_t XXH3_mergeAccs(const U64* acc, const U32* key, U64 start)
 XXH_NO_INLINE XXH64_hash_t    /* It's important for performance that XXH3_hashLong is not inlined. Not sure why (uop cache maybe ?), but difference is large and easily measurable */
 XXH3_hashLong_64b(const void* data, size_t len, XXH64_hash_t seed)
 {
-    ALIGN(64) U64 acc[ACC_NB] = { seed, PRIME64_1, PRIME64_2, PRIME64_3, PRIME64_4, PRIME64_5, (U64)0 - seed, 0 };
+    ALIGN(64) U64 acc[ACC_NB] = { seed, PRIME64_1, PRIME64_2, PRIME64_3, PRIME64_4, PRIME64_5, (U64)0 - seed, 0 }; // fixme: MSVC warns that there's a non-constant initializer
```
easyaspi314: Yeah, I was about to address that 
```c
    ALIGN(64) U64 acc[ACC_NB] = { 0, PRIME64_1, PRIME64_2, PRIME64_3, PRIME64_4, PRIME64_5, 0, 0 };
    acc[0] = seed;
    acc[6] = (U64)0 - seed;
```

xxhash.h
```diff
@@ -291,7 +291,7 @@ struct XXH64_state_s {
    uint64_t v4;
    uint64_t mem64[4];
    uint32_t memsize;
-   uint32_t reserved[2];   /* never read nor write, might be removed in a future version */
+   uint32_t reserved;   /* never read nor write, might be removed in a future version */
```
easyaspi314: Why don't we just remove reserved entirely? It *is* a future version…

xxh3.h
Cyan4973: OK, one issue is that the code source is under change currently.
Latest version is actually in `xxh3` branch.
This specific issue is fixed there.
There are likely other changes which would be of interest for this patch.

I suggest :
- either base your patch on `xxh3`, and PR into `cyan4973/xxh3`
- or wait for `xxh3` to be merged into `dev`. This will hopefully happen later in the week. Might be very soon if I break the changes into 2 parts (`_64bits` first, then `_128bits` later).

czMaxx: If the truncation of the upper 32 bits is desired, then yes

xxh3.h
czMaxx: there may be some defines to avoid the Spectre fix, however I did not dig into it yet

xxhash.h
Cyan4973: I don't want to introduce risks of ABI break while also introducing `xxh3`.

I believe `uint32_t reserved[2];` can or even should be replaced by `uint64_t reserved;`. It doesn't change anything regarding the structure size, it would just be "proper".

`reserved` might be removed in a future version, or it might remain there forever.
The tiny memory savings pale in comparison of risks of ABI break.

xxh3.h
Cyan4973: I wonder what kind of Spectre issue could be raised at this line.
I don't see any branch nor speculation that could be used to trigger a side effect ...

xxhash.h
czMaxx: the question is "who's responsible for not crashing inconsistent headers and libs?", because it has some historical footprint

xxh3.h
Cyan4973: It's more than truncation, the new algorithm in the `xxh3` branch is different in non-trivial ways.

czMaxx: Ok, I can wait for the merge and then either propagate it to my branch or fork it again from the most recent version and create new request.

xxh3.h
czMaxx: I will check the particular asm code to investigate. There's a loop with the arithmetics (so at least one branch is there) and you never know what the compiler decides to do to speed it up.

xxh3.h
czMaxx: Any update on the right timing to fork the repo and fix the few glitches?

Cyan4973:
`XXH3_64bit()` has been updated,
so if you wish you can test and improve it.

The `XXH128()` variant is still in development.

PR #190, opened 2019-03-30
Add PPC64 VSX support, CPU dispatching, unroll fixes, unify SSE2/AVX2, Windows fixes, and better document SIMD code

easyaspi314:
Sorry about the dump.

* The `XXH3_hashLong` `accumulate512` main loop is manually inlined and unrolled twice. On x86 and PPC chips, I force Clang to fully unroll the loop, but on ARM I don't, as only two unrolls seems fastest. 
* SIMD code is heavily documented to improve readability and maintainability.
* The SSE2 and AVX2 code now share the same code block, they use macros and typedefs to use the correct intrinsics.
* VSX is now supported, for both big and little endian. Once full test values are released, I will triple check to verify that all tests pass.
   * When compiled with Clang 8 it is exactly 30 GB/s, even with endian swapping on a POWER9. XXH64 is 15 GB/s, and XXH32 is 5-6. 
* XXH3 will now dispatch to either the SSE2, scalar, or AVX2 functions at runtime when built with  `make MULTI_TARGET=1`. The current method recompiles an isolated version of hashLong three times.
* I changed the dispatch code I initially used to use function pointers.
* I added a way to change/get the CPU mode, and it now verifies whether the CPU actually supports the mode.
* The dispatcher is tested in `./xxhsum -b` when XXH_MULTI_TARGET is defined.
* The welcome message will now show if the CPU has PPC64 VSX.


easyaspi314:
@Cyan4973 you there?

Cyan4973:
Hi @easyaspi314 

I'll get back to `xxh3` later this week,
I've spent a bit too much time on it, and need to catch up other priorities.
I should be in better shape later this week, and your patch is my first item in priority list.

easyaspi314:
I've spent too much as well lol. Midterms have been biting my ass.

easyaspi314:
Ok about to dump an updated CMakeLists.txt which uses a static library, as well as some Windows fixes.

Windows fixes:

- Fixed MSVC truncation warning
- Don't use __emulu on x86_64
- clang-cl has a different cost model or something, and XXHNN_round and XXH_readNN needed XXH_FORCE_INLINE.
- Use restrict as __restrict in C89/C++ mode for MSVC and GCC which slightly speeds things up

MSVC's dispatcher seems to be working, but the XXH3 code is so slow it ends up being slower than both XXH64 and XXH32, and even the scalar version is slightly faster (but not even close to the other results I saw) than the SSE2 version on my 2.8 GHz Core 2 Duo E7400 Wolfdale.

MSVC:
XXH32: 5.1 GB/s
XXH64: 5.1 GB/s
XXH3 SSE2: 4.3 GB/s (could tune to a fragile 4.6)
XXH3 Scalar: 4.7 GB/s

Clang with inline fix
XXH32: 4.9 GB/s < roll bug where clang drops extra movs
XXH64: 5.1 GB/s
XXH3 SSE2: 12 GB/s
XXH3 Scalar: 7.4 GB/s

I'll try tomorrow to get some improvements put in and then I will try to upload.

easyaspi314:
Update: I've been using my desktop (2.8 GHz Core 2 Duo E7400 Wolfdale) on Windows and I have a working albeit hacky CMakeLists.txt which uses static libraries.

```cmake
option(MULTI_TARGET "Enable CPU dispatching if supported" OFF)
...
if (MULTI_TARGET)
	string (TOUPPER "${CMAKE_SYSTEM_PROCESSOR}" processor_name)
	string (REGEX MATCH ".*(AMD64|X86_64|I386|I686|X86).*"  is_x86 "${processor_name}")
	if (is_x86)

		string (REGEX MATCH "(AppleClang|Clang|GNU|Intel)"  is_gcc_family "${CMAKE_C_COMPILER_ID}")
		if (is_gcc_family OR MSVC)
			set (CMAKE_C_FLAGS "-DXXH_MULTI_TARGET=1  ${CMAKE_C_FLAGS} ")
			add_library(xxh3-scalar STATIC "${XXHASH_DIR}/xxh3-target.c")
			add_library(xxh3-sse2 STATIC "${XXHASH_DIR}/xxh3-target.c")
			add_library(xxh3-avx2 STATIC "${XXHASH_DIR}/xxh3-target.c")
			target_link_libraries(xxhash xxh3-scalar xxh3-sse2 xxh3-avx2)
			target_compile_definitions(xxh3-scalar PRIVATE XXH_VECTOR=0)
			target_compile_definitions(xxh3-sse2 PRIVATE XXH_VECTOR=1)
			target_compile_definitions(xxh3-avx2 PRIVATE XXH_VECTOR=2)
			if (MSVC)
				if (NOT ("${is_x86}" STREQUAL "AMD64" OR "${is_x86}" STREQUAL "X86_64"))
					target_compile_options(xxh3-sse2 PRIVATE "-arch:SSE2")
				endif()
				target_compile_options(xxh3-avx2 PRIVATE "-arch:AVX2")
			else()
				target_compile_options(xxh3-scalar PRIVATE "-mno-sse2")
				target_compile_options(xxh3-sse2 PRIVATE "-msse2")
				target_compile_options(xxh3-avx2 PRIVATE "-mavx2")
			endif()
			message(STATUS "Multi target code enabled.")
		endif()				
	endif()
	unset(processor_name)
	unset(is_x86)
endif()
```
MSVC's XXH3 codegen is so bad that the scalar code is faster.
```
xxhsum.exe 0.7.0 (64-bits x86_64 little endian), MSVC 19.20.27508.01, by Yann Collet
Sample of 100 KB...
XXH32               :     102400 ->    52242 it/s ( 5101.7 MB/s)
XXH32 unaligned     :     102400 ->    39896 it/s ( 3896.1 MB/s)
XXH64               :     102400 ->    52805 it/s ( 5156.8 MB/s)
XXH64 unaligned     :     102400 ->    37463 it/s ( 3658.5 MB/s)
XXH3_64bits         :     102400 ->    44579 it/s ( 4353.4 MB/s)
XXH3_64 unaligned   :     102400 ->    33367 it/s ( 3258.5 MB/s)
XXH3_64 SSE2        :     102400 ->    44865 it/s ( 4381.4 MB/s)
XXH3_64b SSE2 unali :     102400 ->    33485 it/s ( 3270.0 MB/s)
XXH3_64 Scalar      :     102400 ->    49748 it/s ( 4858.2 MB/s)
XXH3_64b Scalar una :     102400 ->    26635 it/s ( 2601.1 MB/s)
```

Clang (in clang-cl mode) acts a little weird and apparently it doesn't inline most of the XXH32 and XXH64 helpers.
```
xxhsum.exe 0.7.0 (64-bits x86_64 + SSE2 little endian), Clang 8.0.0 (tags/RELEASE_800/final), by Yann Collet
Sample of 100 KB...
XXH32               :     102400 ->    10509 it/s ( 1026.2 MB/s)
XXH32 unaligned     :     102400 ->     9940 it/s (  970.7 MB/s)
XXH64               :     102400 ->    20617 it/s ( 2013.4 MB/s)
XXH64 unaligned     :     102400 ->    19567 it/s ( 1910.8 MB/s)
XXH3_64bits         :     102400 ->   118933 it/s (11614.5 MB/s)
XXH3_64 unaligned   :     102400 ->   122880 it/s (12000.0 MB/s)
XXH3_64 SSE2        :     102400 ->   121428 it/s (11858.2 MB/s)
XXH3_64b SSE2 unali :     102400 ->   118867 it/s (11608.1 MB/s)
XXH3_64 Scalar      :     102400 ->    34517 it/s ( 3370.8 MB/s)
XXH3_64b Scalar una :     102400 ->    22657 it/s ( 2212.6 MB/s)
```

I fixed it up by FORCE_INLINEing the read and round functions, changing __forceinline to __attribute__  (clang-cl defined _MSC_VER) and it's much faster now.
```
xxhsum.exe 0.7.0 (64-bits x86_64 + SSE2 little endian), Clang 8.0.0 (tags/RELEASE_800/final), by Yann Collet
Sample of 100 KB...
XXH32               :     102400 ->    48947 it/s ( 4780.0 MB/s)
XXH32 unaligned     :     102400 ->    39896 it/s ( 3896.1 MB/s)
XXH64               :     102400 ->    52966 it/s ( 5172.4 MB/s)
XXH64 unaligned     :     102400 ->    36141 it/s ( 3529.4 MB/s)
XXH3_64bits         :     102400 ->   123623 it/s (12072.5 MB/s)
XXH3_64 unaligned   :     102400 ->   119946 it/s (11713.5 MB/s)
XXH3_64 SSE2        :     102400 ->   122880 it/s (12000.0 MB/s)
XXH3_64b SSE2 unali :     102400 ->   119729 it/s (11692.3 MB/s)
XXH3_64 Scalar      :     102400 ->    73143 it/s ( 7142.9 MB/s)
XXH3_64b Scalar una :     102400 ->    71873 it/s ( 7018.8 MB/s)
```

Cyan4973:
nice find @easyaspi314 !

easyaspi314:
Eyyyyyy!!!! I figured it out!

We need to make a copy of acc for MSVC. While the performance is still relatively bad, I'll take it over what I had before.

```
xxhsum.exe 0.7.0 (64-bits x86_64 little endian), MSVC 19.20.27508.01, by Yann Collet
Sample of 100 KB...
XXH32               :     102400 ->    53179 it/s ( 5193.2 MB/s)
XXH32 unaligned     :     102400 ->    40421 it/s ( 3947.4 MB/s)
XXH64               :     102400 ->    53667 it/s ( 5241.0 MB/s)
XXH64 unaligned     :     102400 ->    38944 it/s ( 3803.1 MB/s)
XXH3_64bits         :     102400 ->    86898 it/s ( 8486.1 MB/s)
XXH3_64 unaligned   :     102400 ->    78769 it/s ( 7692.3 MB/s)
XXH3_64 SSE2        :     102400 ->    87771 it/s ( 8571.4 MB/s)
XXH3_64b SSE2 unali :     102400 ->    78422 it/s ( 7658.4 MB/s)
XXH3_64 Scalar      :     102400 ->    49698 it/s ( 4853.3 MB/s)
XXH3_64b Scalar una :     102400 ->    36141 it/s ( 3529.4 MB/s)
```

It also seems to help Clang a tiny bit.
```
xxhsum.exe 0.7.0 (64-bits x86_64 + SSE2 little endian), Clang 8.0.0 (tags/RELEASE_800/final), by Yann Collet
Sample of 100 KB...
XXH32               :     102400 ->    52382 it/s ( 5115.5 MB/s)
XXH32 unaligned     :     102400 ->    40960 it/s ( 4000.0 MB/s)
XXH64               :     102400 ->    53895 it/s ( 5263.2 MB/s)
XXH64 unaligned     :     102400 ->    37012 it/s ( 3614.5 MB/s)
XXH3_64bits         :     102400 ->   128000 it/s (12500.0 MB/s)
XXH3_64 unaligned   :     102400 ->   122880 it/s (12000.0 MB/s)
XXH3_64 SSE2        :     102400 ->   125133 it/s (12220.1 MB/s)
XXH3_64b SSE2 unali :     102400 ->   121104 it/s (11826.5 MB/s)
XXH3_64 Scalar      :     102400 ->    77188 it/s ( 7537.9 MB/s)
XXH3_64b Scalar una :     102400 ->    76800 it/s ( 7500.0 MB/s)
```

I'm gonna try GCC in a few to make sure we don't regress, and then I will push.

easyaspi314:
OK I fixed quite a few Windows bugs, which I put in the commit message.

easyaspi314:
Hmm. Testing around and the Windows clang-cl code is significantly slower than the WSL and MinGW/Cygwin builds.

I was able to get clang-cl to 13ish, but MinGW builds were able to reach 19-20 GB/s which scales much better. Also, XXH32 was 6.4, XXH64 spikes at 9.6, and XXH3 Scalar varies from 8-9. 

I wonder if it is the CMake flags changing the codegen or something. Either way, it is very confusing.

easyaspi314:
@Cyan4973 Mind if I reroll XXH64_finalize for 32-bit? It seems to be faster because of cache.

Cyan4973:
Sounds like a nice improvement @easyaspi314 !

Try to separate your PR "per topic".
From a review perspective, it's much nicer to have a multiple PR, each one targeting a precise objective.

easyaspi314:
Ok I'll separate it when I get home. Got carried away again.

easyaspi314:
OK, so this is weird. The problem with the slower hashes with clang-cl appear to be from xxhsum.

Cyan4973:
It seems this patch depends on the dispatcher patch ?

Cyan4973:
There are a number of improvements which look very interesting in this patch.

PPC64 VSX support, unroll fixes, unify SSE2/AVX2, Windows fixes,
all these look pretty good and desirable.

Is there any way to decouple them from the dynamic dispatcher ?

easyaspi314:
Yeah as xxh3-target.c is basically just extracted from xxh3.h.

For a summary on the changes,
 - Dispatching (although I prefer the newer version)
 - Support for building the dispatch code in both the Makefile and the CMakeLists.txt
 - VSX support
 - Unified SSE2 and AVX2 code
 - Improved documentation of the SIMD code 
 - Fixed NEON code
 - Removed the aarch64 multiply code, aarch64 would have already taken the __uint128_t path by then.
 - Use 64-bit types in the scalar code to improve 64-bit performance without any effect on 32-bit
 - Only use _emulu on x86, as x86_64 is best with the GCC style of mask and multiply.
 - Fix the truncation warning on MSVC (already addressed)
 - Make an aligned copy of acc to improve MSVC performance for no apparent reason
 - Use the extension __restrict on most C/C++ compilers
 - Flush stderr to fix MinTTY issues
 - ...?

I am not sure on whether you want the dispatch code or not.

The code is relatively easy to decouple, however, managing PRs for the dispatch version and the original version separately will be merge conflict hell, so if you merge the multitarget version it will be easier to add the other changes. It all depends on the structure we agree on.

Cyan4973:
I guess we should probably settle on the dynamic dispatcher first, so that you get a sense on how to organize PR dependencies.

Cyan4973:
Does this PR still serve a purpose ?

easyaspi314:
Um not really, but I think we still need the multiple file setup.

arm_neon.h can't even be included with neon disabled. It isn't like the Intel intrinsic headers which only error when you use the functions.

And in order to be compatible with the Play Store, we need to dispatch NEON on ARMv7a.

Cyan4973:
I suppose multi-file is acceptable if it can be optional.
But I wouldn't rush on this feature, there are other aspects which are more important to stabilize.

Branch `xxh128` features a new design for the 128-bits hash.
It's significantly different, so you might want to have a look.

After validating it, I intend on improving the API, by adding streaming capability.

easyaspi314:
Ok, my initial NEON version barely gets 3.2 GB/s on my G3, which is slower than the NEON XXH32 implementation (4.1 GB/s). 

NEON has to do four shuffles instead of 1.

Cyan4973:
If you feel there are some simplifications which are accessible, please feel free to suggest.
It's hard to keep the quality high and the computing cost low, but there are still a few choices left which may end up being equivalent (though only quality tests can tell).

easyaspi314:
For the shuffle, how do 1230, 1032, 2301, and 3012 work? Those can be done in one instruction unlike 3210 which requires 2.

Cyan4973:
We want to transfer input entropy between channel.
This pretty much restricts possible shuffles 
from 0123 to 3210 or 2301.

So maybe 2301.
But I must check. 3210 had the advantage to invert lower and higher parts, which might be more fair in term of multiplication spread. That being said, only tests can tell.

easyaspi314:
We can also do a byte rotate with `_mm_slli_si128` and `vextq_u8`, however, that would probably slow down scalar a lot.

Cyan4973:
For the current scalar code in 64-bit mode, probably not.
Because right now, it uses `>> 32` and `<< 32`, so it would pretty similar to do `>> 24` and `<< 40` instead.

However : 
- `_mm_slli_si128` only does the shift left part. The complementing shift right part must also be done for a rotate. This adds 2 instructions (shift right + recombination) on the vectorial path.
- The 32-bit mode would be slightly more complex, since 32-bit shift is pretty much free for it.

I'm more concerned by the first point, as it's likely to impact negatively SSE/AVX speed.

Cyan4973:
Good news : I've started the "short" (<1h) quality tests for the variant using "shuffle 2301", and it seems to pass all tests successfully.
So, if you feel this could help the `NEON` code path, this becomes a good candidate.

easyaspi314:
Oops. Shuffling isn't the bottleneck, it's the second multiply. That makes more sense.

vmlal.u32 is 5-7 cycles depending on the chip.

Cyan4973:
So what do you think ?
Changing the shuffling is useless ? or is it "a tiny bit" helpful ?

Given that both variants path the same quality tests, I'm opened to both possibilities.

Cyan4973:
Changing the shuffle makes it possible to improve speed of the 64-bit _scalar_ variant.
On my laptop, it went up from 6.3 GB/s to 7.0 GB/s.
This is inline with `XXH3_64bit()` scalar speed with `gcc`, at 7.2 GB/s.
(`clang` is able to auto-vectorize the scalar variant, so it's not a "fair" comparison).

So this seems another argument in favor of changing the shuffle.

PR #192, opened 2019-04-18
Isolate multitarget code, add ARM Linux dispatching

easyaspi314:
- This is the multitargeting code, just the multitargeting code, and nothing but the multitargeting code
- I also added support for dispatching on ARM Linux (incl. Android) which checks /proc/cpuinfo.
- Switches are removed because it triggered -Wswitch-enum.

### DIFF
--- a/xxh3-dispatch.h
+++ b/xxh3-dispatch.h
@@ -1,0 +1,277 @@
+/*
+   xxHash - Extremely Fast Hash algorithm
+   Development source file for `xxh3`
+   Copyright (C) 2019-present, Yann Collet.
+
+   BSD 2-Clause License (http://www.opensource.org/licenses/bsd-license.php)
+
+   Redistribution and use in source and binary forms, with or without
+   modification, are permitted provided that the following conditions are
+   met:
+
+       * Redistributions of source code must retain the above copyright
+   notice, this list of conditions and the following disclaimer.
+       * Redistributions in binary form must reproduce the above
+   copyright notice, this list of conditions and the following disclaimer
+   in the documentation and/or other materials provided with the
+   distribution.
+
+   THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+   "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+   LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+   A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+   OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+   SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+   LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+   DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+   THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+   (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+   OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+   You can contact the author at :
+   - xxHash source repository : https://github.com/Cyan4973/xxHash
+*/
+
+/* This is a private header which implements CPU dispatch code. */
+#ifndef XXH3_DISPATCH_H
+#define XXH3_DISPATCH_H
+
+#define XXH_TARGET_MODE_NONE 0
+#define XXH_TARGET_MODE_X86 1
+#define XXH_TARGET_MODE_ARM 2
+
+/* Figure out the best way to get a cpuid. */
+#if defined(_MSC_VER) && (defined(_M_IX86) || defined(_M_AMD64) || defined(_M_X64))
+ /* MSVC has intrinsics for this. */
+#   include <intrin.h>
+#   define XXH_CPUID __cpuid
+#   define XXH_CPUIDEX __cpuidex
+#   define XXH_TARGET_MODE XXH_TARGET_MODE_X86
+#elif defined(__GNUC__) && (defined(__i386__) || defined(__x86_64__))
+
+/* The GCC family can easily use inline assembly. */
+static void XXH_CPUIDEX(int* cpuInfo, int function_id, int function_ext)
+{
+    int eax, ecx;
+    eax = function_id;
+    ecx = function_ext;
+    __asm__ __volatile__(
+#ifdef (__x86_64__) /* EBX is required for PIC so save it */
+# define EBX "q %%rbx\n"
+#else
+# define EBX "l %%ebx\n"
+#endif
+        "push" EBX
+        "cpuid\n"
+        "movl %%ebx, %1\n"
+        "pop" EBX
+      : "+a" (eax), "=r" (cpuInfo[1]), "+c" (ecx), "=d" (cpuInfo[3]));
+#undef EBX
+    cpuInfo[0] = eax;
+    cpuInfo[2] = ecx;
+}
+static void XXH_CPUID(int *cpuInfo, int function_id)
+{
+    XXH_CPUIDEX(cpuInfo, function_id, 0);
+}
+#  define XXH_TARGET_MODE XXH_TARGET_MODE_X86
+/* ARM Linux */
+#elif defined(__linux__) && (defined(__arm__) || defined(__thumb__) || defined(__thumb2__))
+#  include <stdio.h> /* ARM needs fopen, fclose, and fgets */
+#  define XXH_TARGET_MODE XXH_TARGET_MODE_ARM
+#else
+#  define XXH_TARGET_MODE XXH_TARGET_MODE_NONE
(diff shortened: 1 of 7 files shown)


easyaspi314:
@Cyan4973 Are you ok with this structure? xxh3.h is still able to be combined.

Alternatively, we could also keep the scalar code inside xxh3.h and have the multitargeting code be default.

That might be better, so do something like xxhash-vec.h.

final directory structure:
```
 |- xxhash.c <- required
 |- xxhash.h <-required
 |- xxh3-target.c <- optional
```
I'm thinking that if we have `__has_include`, we can automatically include it if it exists, otherwise define `HAS_XXH3_TARGET_C=1`.

Cyan4973:
Apologies @easyaspi314 ,
I'm sick since yesterday, a condition which has dramatically shrinked my output capacity.
I still don't feel great, but it's only a matter of days before being better.
xxHash is first on my priority list, be sure I'll review your excellent patch as soon as I feel capable enough.

easyaspi314:
Ok, hope you feel better. 

I think I will stick with the add-on idea. XXH3 is perfectly fine in its scalar form: Unlike Highway or Meow, XXH3 doesn't fall flat on its face when SIMD is unavailable. So someone who doesn't want to deal with build system complications can still just copy xxhash.c and xxhash.h.

easyaspi314:
I decided to make both dispatching code and multitarget code optional with the following structure:
|- xxhash.c <- required
|- xxhash.h <- required
|- xxh3-target.c <- optional for dispatching or SIMD
|- xxh3-dispatch.h <- optional for dispatching

easyaspi314:
Should I fix the arm code here or should I wait

Cyan4973:
OK, I'm back to work on xxhash.
This will be my priority for the next 2 weeks, so I'll be a lot more present.
It must conclude with a new release, with most remaining issue addressed.

I think I left with a patch for `XXH3_64bits()` which was "almost complete", but not completely.
Let me get back to that, and finish it properly.
We'll add your multi-targeting code on top of that.

xxhash.h
```diff
@@ -406,6 +406,14 @@ struct XXH64_state_s {
 #  define XXH3_64bits_withSeed XXH_NAME2(XXH_NAMESPACE, XXH3_64bits_withSeed)
 #  define XXH3_128bits XXH_NAME2(XXH_NAMESPACE, XXH3_128bits)
 #  define XXH3_128bits_withSeed XXH_NAME2(XXH_NAMESPACE, XXH3_128bits_withSeed)
+#  define XXH3_setCpuMode XXH_NAME2(XXH_NAMESPACE, XXH3_setCpuMode)
+#  define XXH3_getCpuMode XXH_NAME2(XXH_NAMESPACE, XXH3_getCpuMode)
+/* Private API, but it still needs a namespace because the mulit target code always
```
Cyan4973: Can't these symbols be `static`, hence not present on public side ?

xxh3-dispatch.h
```diff
@@ -0,0 +1,277 @@
...
+        XXH3_hashLong = &_XXH3_hashLong_Scalar;
+        return XXH_CPU_MODE_SCALAR;
+    }
+    cpu_mode = XXH_CPU_MODE_AUTO;
+    XXH3_hashLong = &XXH3_dispatcher;
+    return XXH_CPU_MODE_AUTO;
+#undef TRY_SET_MODE
+}
+
+/* Should we keep this? */
```
Cyan4973: That's a good question.
I definitely understand the function while in debugging mode.

Now, outside of this situation, this seems less essential, but why not.
The xxhash API has a category protected by macro guard `XXH_STATIC_LINKING_ONLY`.
I believe it could be present there.

Cyan4973:
From an overview perspective, this patch is much more involved than I expected.
I'm still trying to understand if, at target, it's able to be included within source code, so that only `xxhash.h` and `xxhash.c` remain. I suspect it can, but still need to look into details to be sure.

The second point is the relatively high complexity this patch introduces. This will be costly, from a maintenance perspective. It's more than a matter of "being lazy". Evolutions will be slower, more difficult, more risky, bugs will be more likely, and harder to hunt, overtime it can become such a tax that the software cannot evolve anymore.

So complexity must be worth it.
To be fair, the ability to trigger AVX2 automatically without user intervention seems to be worth it.

But I'm surprised by the amount of code required for this.
That's because we have deployed a _similar_ system in `zstd`, though more limited in number of architectures dispatched, and it did not felt so expensive in term of lines of code or file structure.

Last issue : `AVX2` is known for impacting CPU speed. This is explained [here](https://en.wikichip.org/wiki/intel/frequency_behavior) for example. It may seem a curiosity, and of course the performance of the hot loop remains pretty good in spite of the clock frequency drop, but the problems is more about all the other processes running on the same server that can also be negatively impacted. In datacenter, servers can have > 20 cores with hundreds of active processes, so the gains from hashing something with `AVX2` can be more than offset by all the processes losses due to `AVX2` frequency downclocking.

What I mean is, enabling `AVX2` is not a free lunch. It might be worth it, but it depends on the specific workload of the server, and enabling it might end up being detrimental _globally_.
(side note : there is no such issue with `SSE2`).

So, with this context in mind, we must make sure that users can control this behavior.
`XXH_VECTOR` was supposed to offer this control.
This patch also changes this behavior, so it needs to clarify / document how the user can opt-in / opt-out from the dynamic selection mechanism.

Cyan4973:
Another thing I was thinking : 
pushing a dynamic dispatcher into the library is one thing, which is relatively complex to do properly, since it will impact every single program linking to the `xxhash` library.

But in the CLI (`xxhsum`), it looks easier. On top of that, there are less requirements on this side regarding the number of files and their complexity.
It could be a "proving ground", where the mechanisms can be tested and perfected, without impacting the library's integration into 3rd-party applications.

What do you think ?

easyaspi314:
Well if we remove the setCpuMode code, it will be simpler, but that would remove the option to toggle it which I believe is a useful feature.

Most of the code is simply verification to avoid issues like what happens on Windows where the program will silently fail.

The CPUID/cpuinfo check will save the highest supported version on startup as well as switching to the automatic CPU mode. Then, when XXH_setCpuMode is called, it will compare the requested mode with the supported mode and fall back to the proper mode. I preferred it as a switch statement, but it made Clang complain about -Wswitch-enum (y u no like default)


Basically, in switch form and without the NEON test, it hopefully becomes clearer what is going on:
```c
XXH3_cpu_mode_t
XXH_setCpuMode(XXH_cpu_mode_t requested)
{
    run_feature_test_if_needed();
    if (requested == XXH_CPU_MODE_AUTO)
        requested = supported;
    switch(requested) {
        case XXH_CPU_MODE_AVX2:
            if (supported >= XXH_CPU_MODE_AVX2) {
                XXH3_hashLong = &_XXH3_hashLong_AVX2;
                return (cpu_mode = XXH_CPU_MODE_AVX2);
            }
        // Fallthrough
        case XXH_CPU_MODE_SSE2:
            if (supported >= XXH_CPU_MODE_SSE2) {
                XXH3_hashLong = &_XXH3_hashLong_SSE2;
                return (cpu_mode = XXH_CPU_MODE_SSE2);
            }
        // Fallthrough
        default:
            XXH3_hashLong = &_XXH3_hashLong_Scalar;
            return (cpu_mode = XXH_CPU_MODE_SCALAR);
    }
}

```

easyaspi314:
And the other thing is I had to do some weird shit in order to make XXH3 modular. 

One of your requests was the ability to keep the standalone implementation, so you only have to pop in xxhash.c and xxhash.h. Since there is no way to support multiple targets at once on MSVC AFAIK, this is the easiest option. 

Unless you are fine with having MSVC use the SSE2 path exclusively, but I don't want to make it incompatible because who knows, maybe one day, MSVC will become a decent compiler…

```
$ cd msvc-src
$ git rm -rf ./*
$ git pull https://github.com/llvm-mirror/clang
$ git commit -m "Significant performance, compliance, and codegen improvents"
```
if only

I plan to move the scalar code completely into xxh3.h though, so that xxh3-target.c only has SIMD code. That removes the normal duplication issue. 

The dependency chain is like this. xxh3-dispatch.h is a one way optional dependency for xxh3-target.c which is a one way optional dependency for xxhash.c (currently xxh3.h):



As for maintainability, I don't think it is that much more complicated. It does have a bit of weirdness, but otherwise it is fairly clear that the code goes there.

I might rename the files to xxh3-simd.c and xxh3-multi.h to make their intent clearer.

Cyan4973:
Generally speaking I feel your plan makes sense, and it's well presented.
(+1 for the suggested file name change, I believe it helps clarifying roles).

However, there is something which is hard to measure and transmit, which is about "how much change I'm comfortable introducing in a new release".

`XXH3` is, in itself, a pretty big change, I'm planning to expand its API support for next release, v0.8.
This dynamic dispatcher also introduces some important new subtleties, especially at build stage.
I don't feel comfortable to force users to "accept everything or nothing".

Making the dispatcher separate and optional is a fine move.
At least their is a way to say "you don't have to opt-in to everything".

But I don't think that `xxh3-simd.c` can be optional. I believe SIMD performance is part of the expectation from `xxh3`, so even is a scalar mode is shipped, it's not the "expected" one.

If you could find a way to make only the dispatcher optional, while SIMD remains part of the baseline source, it would be an easier sell.

> Unless you are fine with having MSVC use the SSE2 path exclusively, but I don't want to make it incompatible because who knows, maybe one day, MSVC will become a decent compiler…

I don't want to make the build / integration more complex for everybody because of MSVC.
It's fine to add complexity for MSVC, but it's not fine to make `gcc` or `clang` build more complex to support MSVC.

For reference, the dynamic dispatcher within `zstd` selected to sacrifice MSVC precisely for this reason (so it's actually only supported for `gcc` and `clang`).

easyaspi314:
Ok, so if you're cool with scrapping the MSVC dispatching support, that makes things a little easier.

I am just going to rewrite it to make things cleaner because now I don't need to emulate the MSVC intrinsic. Take three.

PR #198, opened 2019-04-22
Update XXH3-64bits

Cyan4973:
- secret key only used as mask
- seed modifies secret key


Cyan4973:
OK, so it seems we have an issue with the `aarch64` code : https://travis-ci.org/Cyan4973/xxHash/jobs/523245312#L516

Likely one of the changes was not ported to this code path.

Cyan4973:
@easyaspi314 , I think I will need your help to make the `aarch64` test pass by fixing the `NEON` scrambler.

Cyan4973:
I'm going to merge this update now.
It changes the 64-bit formula significantly, while maintaining its original speed.

I'll now work on the 128-bit variant.

Updates on the vectorial part can come any time later.

PR #199, opened 2019-04-26
ARM and scalar improvements/fixes

easyaspi314:
* XXH3_accumulate has the better documented version for neon
* XXH3_scrambleAcc works properly now
* XXH3_accumulate's scalar version uses 64-bit types for better performance on 64-bit targets
* Added the restrict keyword to acc to prevent any reloads
* Removed the aarch64 mult128 code; GCC would never reach it anyways
* Disabled forced unrolling on ARM Clang, it is detrimental to performance.


Cyan4973:
Great Improvements ! Thanks @easyaspi314 !

PR #200, opened 2019-04-26
dead code removed, performance tuning for MSVC

czMaxx:
tested in MSVC 32bit/64bit, GCC7 64bit

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -633,7 +633,7 @@ XXH_FORCE_INLINE void XXH3_initKeySeed(U32* key, U64 seed64)
 XXH_NO_INLINE XXH64_hash_t    /* It's important for performance that XXH3_hashLong is not inlined. Not sure why (uop cache maybe ?), but difference is large and easily measurable */
 XXH3_hashLong_64b(const void* data, size_t len, XXH64_hash_t seed)
 {
-    ALIGN(64) U64 acc[ACC_NB] = { seed, PRIME64_1, PRIME64_2, PRIME64_3, PRIME64_4, PRIME64_5, (U64)0 - seed, 0 };
+    ALIGN(64) U64 acc[ACC_NB] = { seed, PRIME64_1, PRIME64_2, PRIME64_3, PRIME64_4, PRIME64_5, (U64)0 - seed, 0 }; // fixme: warning C4204: nonstandard extension used: non-constant aggregate initializer
     ALIGN(64) U32 key[KEYSET_DEFAULT_SIZE];
 
     XXH3_initKeySeed(key, seed);
@@ -718,7 +718,7 @@ XXH3_len_1to3_128b(const void* data, size_t len, const void* keyPtr, XXH64_hash_
         U32  const l2 = (U32)(len) + ((U32)(c3) << 2);
         U64  const ll11 = XXH_mult32to64((unsigned int)(l1 + seed + key32[0]), (unsigned int)(l2 + key32[1]));
         U64  const ll12 = XXH_mult32to64((unsigned int)(l1 + key32[2]), (unsigned int)(l2 - seed + key32[3]));
-        XXH128_hash_t const h128 = { XXH3_avalanche(ll11), XXH3_avalanche(ll12) };
+        XXH128_hash_t const h128 = { XXH3_avalanche(ll11), XXH3_avalanche(ll12) }; // fixme: warning C4204: nonstandard extension used: non-constant aggregate initializer
         return h128;
     }
 }
--- a/xxh3.h
+++ b/xxh3.h
@@ -633,7 +633,7 @@ XXH_FORCE_INLINE void XXH3_initKeySeed(U32* key, U64 seed64)
 XXH_NO_INLINE XXH64_hash_t    /* It's important for performance that XXH3_hashLong is not inlined. Not sure why (uop cache maybe ?), but difference is large and easily measurable */
 XXH3_hashLong_64b(const void* data, size_t len, XXH64_hash_t seed)
 {
-    ALIGN(64) U64 acc[ACC_NB] = { seed, PRIME64_1, PRIME64_2, PRIME64_3, PRIME64_4, PRIME64_5, (U64)0 - seed, 0 }; // fixme: warning C4204: nonstandard extension used: non-constant aggregate initializer
+    ALIGN(64) U64 acc[ACC_NB] = { seed, PRIME64_1, PRIME64_2, PRIME64_3, PRIME64_4, PRIME64_5, (U64)0 - seed, 0 }; /* fixme: warning C4204: nonstandard extension used: non-constant aggregate initializer */
     ALIGN(64) U32 key[KEYSET_DEFAULT_SIZE];
 
     XXH3_initKeySeed(key, seed);
@@ -718,7 +718,7 @@ XXH3_len_1to3_128b(const void* data, size_t len, const void* keyPtr, XXH64_hash_
         U32  const l2 = (U32)(len) + ((U32)(c3) << 2);
         U64  const ll11 = XXH_mult32to64((unsigned int)(l1 + seed + key32[0]), (unsigned int)(l2 + key32[1]));
         U64  const ll12 = XXH_mult32to64((unsigned int)(l1 + key32[2]), (unsigned int)(l2 - seed + key32[3]));
-        XXH128_hash_t const h128 = { XXH3_avalanche(ll11), XXH3_avalanche(ll12) }; // fixme: warning C4204: nonstandard extension used: non-constant aggregate initializer
+        XXH128_hash_t const h128 = { XXH3_avalanche(ll11), XXH3_avalanche(ll12) }; /* fixme: warning C4204: nonstandard extension used: non-constant aggregate initializer */
         return h128;
     }
 }
--- a/xxhash.c
+++ b/xxhash.c
@@ -351,7 +351,7 @@ static U32 XXH32_avalanche(U32 h32)
 
 #define XXH_get32bits(p) XXH_readLE32_align(p, align)
 
-// handle switch fallthrough
+/* handle switch fallthrough */
 #if defined( __GNUC__ )
 	#define FALL_INTENDED	__attribute__((fallthrough))
 #else
@@ -717,14 +717,14 @@ typedef union _XXH_ULARGE_INTEGER
 	U64	Quad;
 	struct
 	{
-		U32	Lo;	// 32bit Win is always LE
+		U32	Lo;	/* 32bit Win is always LE */
 		U32	Hi;
 	} s;
 } XXH_ULARGE_INTEGER;
 
-// 64bit multiplication emits __aullmul on Win32, which is slow as hell
-// let's replace it with an inline multiplication based on intrinsic
-// we can gain nearly 80% on performance of XXH64 compiled for 32bit
+/* 64bit multiplication emits __aullmul on Win32, which is slow as hell
+ let's replace it with an inline multiplication based on intrinsic
+ we can gain nearly 80% on performance of XXH64 compiled for 32bit */
 XXH_FORCE_INLINE U64 XXH_Mul64x64to64( const U64 u64A, const U64 u64B )
 {
 	XXH_ULARGE_INTEGER a;
--- a/xxhash.c
+++ b/xxhash.c
@@ -351,13 +351,6 @@ static U32 XXH32_avalanche(U32 h32)
 
 #define XXH_get32bits(p) XXH_readLE32_align(p, align)
 
-/* handle switch fallthrough */
-#if defined( __GNUC__ )
-	#define FALL_INTENDED	__attribute__((fallthrough))
-#else
-	#define FALL_INTENDED
-#endif
-
 static U32
 XXH32_finalize(U32 h32, const void* ptr, size_t len, XXH_alignment align)
 
@@ -378,12 +371,12 @@ XXH32_finalize(U32 h32, const void* ptr, size_t len, XXH_alignment align)
 		case 12:
 		{
 			PROCESS4;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 8:
 		{
 			PROCESS4;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 4:
 		{
@@ -393,12 +386,12 @@ XXH32_finalize(U32 h32, const void* ptr, size_t len, XXH_alignment align)
 		case 13:
 		{
 			PROCESS4;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 9:
 		{
 			PROCESS4;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 5:
 		{
@@ -409,12 +402,12 @@ XXH32_finalize(U32 h32, const void* ptr, size_t len, XXH_alignment align)
 		case 14:
 		{
 			PROCESS4;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 10:
 		{
 			PROCESS4;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 6:
 		{
@@ -426,32 +419,32 @@ XXH32_finalize(U32 h32, const void* ptr, size_t len, XXH_alignment align)
 		case 15:
 		{
 			PROCESS4;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 11:
 		{
 			PROCESS4;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 7:
 		{
 			PROCESS4;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 3:
 		{
 			PROCESS1;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 2:
 		{
 			PROCESS1;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 1:
 		{
 			PROCESS1;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 0:
 		default:
@@ -711,36 +704,38 @@ static U64 XXH_read64(const void* memPtr)
 
 #if defined(_MSC_VER)     /* Visual Studio */
 #  define XXH_swap64 _byteswap_uint64
-#if defined(_M_IX86)
-typedef union _XXH_ULARGE_INTEGER
-{
-	U64	Quad;
-	struct
-	{
-		U32	Lo;	/* 32bit Win is always LE */
-		U32	Hi;
-	} s;
-} XXH_ULARGE_INTEGER;
 
-/* 64bit multiplication emits __aullmul on Win32, which is slow as hell
- let's replace it with an inline multiplication based on intrinsic
- we can gain nearly 80% on performance of XXH64 compiled for 32bit */
-XXH_FORCE_INLINE U64 XXH_Mul64x64to64( const U64 u64A, const U64 u64B )
-{
-	XXH_ULARGE_INTEGER a;
-	a.Quad = u64A;
-	XXH_ULARGE_INTEGER b;
-	b.Quad = u64B;
+  #if defined(_M_IX86)
+  typedef union _XXH_ULARGE_INTEGER
+  {
+  	U64	Quad;
+  	struct
+  	{
+		  U32	Lo;	/* 32bit Win is always LE */
+		  U32	Hi;
+	  } s;
+  } XXH_ULARGE_INTEGER;
 
-	XXH_ULARGE_INTEGER c;
-	c.Quad = __emulu( a.s.Lo, b.s.Lo );
+  /* 64bit multiplication emits __aullmul on Win32, which is slow as hell
+   let's replace it with an inline multiplication based on intrinsic
+   we can gain nearly 80% on performance of XXH64 compiled for 32bit */
+  XXH_FORCE_INLINE U64 XXH_Mul64x64to64( const U64 u64A, const U64 u64B )
+  {
+  	XXH_ULARGE_INTEGER a;
+  	a.Quad = u64A;
+  	XXH_ULARGE_INTEGER b;
+  	b.Quad = u64B;
 
-	c.s.Hi += (U32)(a.s.Hi * b.s.Lo);
-	c.s.Hi += (U32)(a.s.Lo * b.s.Hi);
+  	XXH_ULARGE_INTEGER c;
+  	c.Quad = __emulu( a.s.Lo, b.s.Lo );
 
-	return c.Quad;
-}
-#endif
+  	c.s.Hi += (U32)(a.s.Hi * b.s.Lo);
+  	c.s.Hi += (U32)(a.s.Lo * b.s.Hi);
+
+  	return c.Quad;
+ }
+ #endif
+
 #elif XXH_GCC_VERSION >= 403
 #  define XXH_swap64 __builtin_bswap64
 #else
@@ -758,7 +753,7 @@ static U64 XXH_swap64 (U64 x)
 #endif
 
 #if !defined( XXH_Mul64x64to64 )
-#define XXH_Mul64x64to64( a, b )	((U64)(a) * (U64)(b))
+# define XXH_Mul64x64to64( a, b )	((U64)(a) * (U64)(b))
 #endif
 
 XXH_FORCE_INLINE U64 XXH_readLE64(const void* ptr)
@@ -843,12 +838,12 @@ XXH64_finalize(U64 h64, const void* ptr, size_t len, XXH_alignment align)
 		case 24:
 		{
 			PROCESS8_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 16:
 		{
 			PROCESS8_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 8:
 		{
@@ -858,17 +853,17 @@ XXH64_finalize(U64 h64, const void* ptr, size_t len, XXH_alignment align)
 		case 28:
 		{
 			PROCESS8_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 20:
 		{
 			PROCESS8_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 12:
 		{
 			PROCESS8_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 4:
 		{
@@ -878,12 +873,12 @@ XXH64_finalize(U64 h64, const void* ptr, size_t len, XXH_alignment align)
 		case 25:
 		{
 			PROCESS8_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 17:
 		{
 			PROCESS8_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 9:
 		{
@@ -894,17 +889,17 @@ XXH64_finalize(U64 h64, const void* ptr, size_t len, XXH_alignment align)
 		case 29:
 		{
 			PROCESS8_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 21:
 		{
 			PROCESS8_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 13:
 		{
 			PROCESS8_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 5:
 		{
@@ -915,12 +910,12 @@ XXH64_finalize(U64 h64, const void* ptr, size_t len, XXH_alignment align)
 		case 26:
 		{
 			PROCESS8_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 18:
 		{
 			PROCESS8_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 10:
 		{
@@ -932,17 +927,17 @@ XXH64_finalize(U64 h64, const void* ptr, size_t len, XXH_alignment align)
 		case 30:
 		{
 			PROCESS8_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 22:
 		{
 			PROCESS8_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 14:
 		{
 			PROCESS8_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 6:
 		{
@@ -954,12 +949,12 @@ XXH64_finalize(U64 h64, const void* ptr, size_t len, XXH_alignment align)
 		case 27:
 		{
 			PROCESS8_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 19:
 		{
 			PROCESS8_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 11:
 		{
@@ -972,37 +967,37 @@ XXH64_finalize(U64 h64, const void* ptr, size_t len, XXH_alignment align)
 		case 31:
 		{
 			PROCESS8_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 23:
 		{
 			PROCESS8_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 15:
 		{
 			PROCESS8_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 7:
 		{
 			PROCESS4_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 3:
 		{
 			PROCESS1_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 2:
 		{
 			PROCESS1_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 1:
 		{
 			PROCESS1_64;
-			FALL_INTENDED;
+			/* fallthrough */
 		}
 		case 0:
 		default:
--- a/xxhash.c
+++ b/xxhash.c
@@ -106,21 +106,10 @@ static void* XXH_memcpy(void* dest, const void* src, size_t size) { return memcp
  is time-expensive for such small buffers
 
 
-XXH_FORCEINLINE void XXH_SmallMemCpy( void* pDest, const void* pSrc, const U32 u32Size )
-{
-	BYTE* const pu8Dest = (BYTE*)pDest;
-	const BYTE* const pu8Src = (const BYTE*)pSrc;
-
-	for ( uint32_t i = 0U; i < u32Size; i++ )
-	{
-		pu8Dest[ i ] = pu8Src[ i ];
-	}
-
-	return;
-}
+#define XXH_memcpy( dst, src, size ) __movsb( (unsigned char*)(dest), (unsigned char*)(src), (size) )
 
 
-the "dummy" inlined memcpy above gives better results for the small buffers */
+the define above can be wrapped with #ifdef _MSC_VER */
 
 #include <assert.h>   /* assert */
 
--- a/xxhash.c
+++ b/xxhash.c
@@ -98,14 +98,14 @@ static void  XXH_free  (void* p)  { free(p); }
 /*! and for memcpy() */
 #include <string.h>
 static void* XXH_memcpy(void* dest, const void* src, size_t size) { return memcpy(dest,src,size); }
-// there's a significant performance penalty in the stream version of XXH
-// when you use the traditional library memcpy in MSVC!!
-// we always copy a maximum of 32 bytes, which is pretty small number
-// and the overhead of the call to memcpy (which is not inlined)
-// and the decision whether it is worth using fast memcpy (SSE etc.)
-// is time-expensive for such small buffers
+/* there's a significant performance penalty in the stream version of XXH
+ when you use the traditional library memcpy in MSVC!!
+ we always copy a maximum of 32 bytes, which is pretty small number
+ and the overhead of the call to memcpy (which is not inlined)
+ and the decision whether it is worth using fast memcpy (SSE etc.)
+ is time-expensive for such small buffers
 
-/*
+
 XXH_FORCEINLINE void XXH_SmallMemCpy( void* pDest, const void* pSrc, const U32 u32Size )
 {
 	BYTE* const pu8Dest = (BYTE*)pDest;
@@ -118,9 +118,9 @@ XXH_FORCEINLINE void XXH_SmallMemCpy( void* pDest, const void* pSrc, const U32 u
 
 	return;
 }
-*/
 
-// the "dummy" inlined memcpy above gives better results for the small buffers
+
+the "dummy" inlined memcpy above gives better results for the small buffers */
 
 #include <assert.h>   /* assert */
 
--- a/xxhash.c
+++ b/xxhash.c
@@ -98,6 +98,29 @@ static void  XXH_free  (void* p)  { free(p); }
 /*! and for memcpy() */
 #include <string.h>
 static void* XXH_memcpy(void* dest, const void* src, size_t size) { return memcpy(dest,src,size); }
+// there's a significant performance penalty in the stream version of XXH
+// when you use the traditional library memcpy in MSVC!!
+// we always copy a maximum of 32 bytes, which is pretty small number
+// and the overhead of the call to memcpy (which is not inlined)
+// and the decision whether it is worth using fast memcpy (SSE etc.)
+// is time-expensive for such small buffers
+
+/*
+XXH_FORCEINLINE void XXH_SmallMemCpy( void* pDest, const void* pSrc, const U32 u32Size )
+{
+	BYTE* const pu8Dest = (BYTE*)pDest;
+	const BYTE* const pu8Src = (const BYTE*)pSrc;
+
+	for ( uint32_t i = 0U; i < u32Size; i++ )
+	{
+		pu8Dest[ i ] = pu8Src[ i ];
+	}
+
+	return;
+}
+*/
+
+// the "dummy" inlined memcpy above gives better results for the small buffers
 
 #include <assert.h>   /* assert */
 
@@ -109,7 +132,7 @@ static void* XXH_memcpy(void* dest, const void* src, size_t size) { return memcp
 *  Compiler Specific Options
 ***************************************/
 #ifdef _MSC_VER    /* Visual Studio */
-#  pragma warning(disable : 4127)      /* disable: C4127: conditional expression is constant */
+#  pragma warning(disable : 4127 5045)      /* disable: C4127: conditional expression is constant, C5045: Spectre mitigation warning */
 #  define XXH_FORCE_INLINE static __forceinline
 #  define XXH_NO_INLINE static __declspec(noinline)
 #else
@@ -328,6 +351,13 @@ static U32 XXH32_avalanche(U32 h32)
 
 #define XXH_get32bits(p) XXH_readLE32_align(p, align)
 
+// handle switch fallthrough
+#if defined( __GNUC__ )
+	#define FALL_INTENDED	__attribute__((fallthrough))
+#else
+	#define FALL_INTENDED
+#endif
+
 static U32
 XXH32_finalize(U32 h32, const void* ptr, size_t len, XXH_alignment align)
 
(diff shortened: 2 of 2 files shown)


czMaxx:
aah, the fallthrough attribute is not supported in C90. thought it will be better than the comment, but obviously not.. I will fix it next week

easyaspi314:
Eh not so sure about how I feel about this.

XXH3 was supposed to be the new alternative to XXH64 which is made to be fast on 32-bit and 64-bit, and more than 2x faster than XXH64 even on 64-bit.

XXH64 does have a problem with 32-bit, but that is expected because it does 64-bit arithmetic which is expensive.

I have tried optimizing XXH64 into SSE2, however, it is a very, very long code block which only benefits 32-bit.

An SSE4 version can be faster, but only on Nehalem, Sandy Bridge, and Ivy Bridge. Haswell nerfed `pmulld` significantly, so yeah, it would require a lot of different checks.

```c
const __m128i prime1_lo = _mm_set1_epi32((int)PRIME64_1);
const __m128i prime2_lo = _mm_set1_epi32((int)PRIME64_2);
const __m128i prime1_hi = _mm_set1_epi32((int)(PRIME64_1 >> 32));
const __m128i prime2_hi = _mm_set1_epi32((int)(PRIME64_2 >> 32));
__m128i v[2] = ...;

{
    int i;
    for (i = 0; i < 2; i++) {
        __m128i inp_lo = _mm_loadu_si128((const __m128i *)p);
        __m128i inp_hi = _mm_shuffle_epi32(inp_lo, _MM_SHUFFLE(2, 3, 0, 1));
        __m128i lo = _mm_mul_epu32(inp_lo, prime2_lo);
        __m128i lo_hi = _mm_mul_epu32(inp_lo, prime2_hi);
        __m128i hi_lo = _mm_mul_epu32(inp_hi, prime2_lo);
        __m128i hi = _mm_slli_epi64(_mm_add_epi64(lo_hi, hi_lo), 32);
        __m128i prod = _mm_add_epi64(hi, lo);
        __m128i acc = _mm_add_epi64(v[i], prod);
        __m128i temp = _mm_slli_epi64(acc, 31);
        acc = _mm_srli_epi64(acc, 33);
        acc = _mm_or_si128(acc, temp);
        inp_hi = _mm_shuffle_epi32(acc, _MM_SHUFFLE(2, 3, 0, 1));
        lo = _mm_mul_epu32(acc, prime1_lo);
        lo_hi = _mm_mul_epu32(acc, prime1_hi);
        hi_lo = _mm_mul_epu32(inp_hi, prime1_lo);
        hi = _mm_slli_epi64(_mm_add_epi64(lo_hi, hi_lo), 32);
        v[i] = _mm_add_epi64(hi, lo);
        p += 16;
    }
}
```
That is 18 LOC for *two* rounds, not four, and it is pretty hard to follow. That gets 2 GB/s approximately, and it isn't really worth it when it is still slower than XXH32.

Have you tried changing XXH_FORCE_MEMORY_ACCESS? Try defining it to 2.

xxh3.h
```diff
@@ -633,7 +633,7 @@ XXH_FORCE_INLINE void XXH3_initKeySeed(U32* key, U64 seed64)
 XXH_NO_INLINE XXH64_hash_t    /* It's important for performance that XXH3_hashLong is not inlined. Not sure why (uop cache maybe ?), but difference is large and easily measurable */
 XXH3_hashLong_64b(const void* data, size_t len, XXH64_hash_t seed)
 {
-    ALIGN(64) U64 acc[ACC_NB] = { seed, PRIME64_1, PRIME64_2, PRIME64_3, PRIME64_4, PRIME64_5, (U64)0 - seed, 0 };
+    ALIGN(64) U64 acc[ACC_NB] = { seed, PRIME64_1, PRIME64_2, PRIME64_3, PRIME64_4, PRIME64_5, (U64)0 - seed, 0 }; /* fixme: warning C4204: nonstandard extension used: non-constant aggregate initializer */
```
Cyan4973: 👍

xxhash.c
```diff
@@ -993,7 +1191,7 @@ XXH_PUBLIC_API unsigned long long XXH64_digest (const XXH64_state_t* state)
 
     h64 += (U64) state->total_len;
 
-    return XXH64_finalize(h64, state->mem64, (size_t)state->total_len, XXH_aligned);
+    return XXH64_finalize(h64, state->mem64, (size_t)state->memsize, XXH_aligned);
```
Cyan4973: There's a need for consistency between the 32 and 64 bit versions.
Either select to use the total `len` when invoking `_finalize()`, and expect `_finalize()` to do the final masking,
or select to use the `len` of the final partial stripe, and `assert()` that it's correctly set in the `_finalize()`.

The first one looks more general, because it doesn't add constraint to the caller, but the second one might saving one masking operation (which probably does not matter much).

Either way is fine with me. Consistency matters more for this point.

xxhash.c
```diff
@@ -635,6 +711,36 @@ static U64 XXH_read64(const void* memPtr)
 
 #if defined(_MSC_VER)     /* Visual Studio */
 #  define XXH_swap64 _byteswap_uint64
+#if defined(_M_IX86)
```
Cyan4973: indent macros so that we can clearly understand that this is a subsection on `_MSC_VER`.

xxh3.h
easyaspi314: I do want to mention that MSVC will not let you use static const variables as constant expressions in `/Ta` or whatever the flag is for ANSI conformance.

However, agreed that seed should be inserted manually.

Cyan4973: I need to add a test to ensure such issues are caught during CI tests.

xxhash.c
czMaxx: state->memsize is supposed to hold the number of bytes residing in the accumulator and since we refer to the accumulator, I find it more consistent to use the memsize parameter istead of the total_len (the size is anded after all and both numbers give the same result, but this looks somehow clearer, IMO).

czMaxx:
> 
> 
> Eh not so sure about how I feel about this.
> 
> XXH3 was supposed to be the new alternative to XXH64 which is made to be fast on 32-bit and 64-bit, and more than 2x faster than XXH64 even on 64-bit.
> 
> XXH64 does have a problem with 32-bit, but that is expected because it does 64-bit arithmetic which is expensive.
> 
> I have tried optimizing XXH64 into SSE2, however, it is a very, very long code block which only benefits 32-bit.
> 
> An SSE4 version can be faster, but only on Nehalem, Sandy Bridge, and Ivy Bridge. Haswell nerfed `pmulld` significantly, so yeah, it would require a lot of different checks.
> 
> ```c
> const __m128i prime1_lo = _mm_set1_epi32((int)PRIME64_1);
> const __m128i prime2_lo = _mm_set1_epi32((int)PRIME64_2);
> const __m128i prime1_hi = _mm_set1_epi32((int)(PRIME64_1 >> 32));
> const __m128i prime2_hi = _mm_set1_epi32((int)(PRIME64_2 >> 32));
> __m128i v[2] = ...;
> 
> {
>     int i;
>     for (i = 0; i < 2; i++) {
>         __m128i inp_lo = _mm_loadu_si128((const __m128i *)p);
>         __m128i inp_hi = _mm_shuffle_epi32(inp_lo, _MM_SHUFFLE(2, 3, 0, 1));
>         __m128i lo = _mm_mul_epu32(inp_lo, prime2_lo);
>         __m128i lo_hi = _mm_mul_epu32(inp_lo, prime2_hi);
>         __m128i hi_lo = _mm_mul_epu32(inp_hi, prime2_lo);
>         __m128i hi = _mm_slli_epi64(_mm_add_epi64(lo_hi, hi_lo), 32);
>         __m128i prod = _mm_add_epi64(hi, lo);
>         __m128i acc = _mm_add_epi64(v[i], prod);
>         __m128i temp = _mm_slli_epi64(acc, 31);
>         acc = _mm_srli_epi64(acc, 33);
>         acc = _mm_or_si128(acc, temp);
>         inp_hi = _mm_shuffle_epi32(acc, _MM_SHUFFLE(2, 3, 0, 1));
>         lo = _mm_mul_epu32(acc, prime1_lo);
>         lo_hi = _mm_mul_epu32(acc, prime1_hi);
>         hi_lo = _mm_mul_epu32(inp_hi, prime1_lo);
>         hi = _mm_slli_epi64(_mm_add_epi64(lo_hi, hi_lo), 32);
>         v[i] = _mm_add_epi64(hi, lo);
>         p += 16;
>     }
> }
> ```
> 
> That is 18 LOC for _two_ rounds, not four, and it is pretty hard to follow. That gets 2 GB/s approximately, and it isn't really worth it when it is still slower than XXH32.
> 
> Have you tried changing XXH_FORCE_MEMORY_ACCESS? Try defining it to 2.

There are several things that can slow down certain ways of hashing. One of them is the memcpy function in the stream version. I can't enforce the simple inlined version and MSVC always emits a call to the robust function, which is fine for large buffers, but absolutely ineffective when you hash a buffer byte by byte e.g. (there should be an intrinsic for rep movs, which would be probably the most efficient way to copy small buffers - we don't exceed 64 bytes), so there's a room for improvement.

Another one is the 64bit multiplication on 32bit platforms. No clue how GCC handles this (I compile only 64bit version with GCC), but MSVC never inlines the __aullmul function, which makes the result terribly slow. This is resolved with the custom multiplication based on intrinsics.

Vectorization (SSE 4.1) helps significantly in all 32bit builds, except 32bit XXH32 - the 3-operand imul is quite fast and probably can't be beaten. In 64bit version, the native code is faster than SSE, except stream XXH32. Btw: your SSE version relies on compilers compatible with the Intel intrinsics guide. There's an easy way to make it compilable with GCC builtin functions.

#define XXH_XMMWORD	__m128i

MSVC:
		#define XXH_sse_loada( X )		_mm_load_si128( (X) )
		#define XXH_sse_loadu( X )		_mm_loadu_si128( (X) )
		#define XXH_sse_pmulld( a, b )	_mm_mullo_epi32( (a), (b) )
		#define XXH_sse_paddd( a, b )	_mm_add_epi32( (a), (b) )
		#define XXH_sse_pslldi( a, b )	_mm_slli_epi32( (a), (b) )
		#define XXH_sse_psrldi( a, b )	_mm_srli_epi32( (a), (b) )
		#define XXH_sse_por( a, b )		_mm_or_si128( (a), (b) )
		#define XXH_sse_stora( a, b )	_mm_store_si128( (a), (b) )
		#define XXH_sse_pshufd( a, b )	_mm_shuffle_epi32( (a), (b) )
		#define XXH_sse_phaddd( a, b )	_mm_hadd_epi32( (a), (b) )
		#define XXH_sse_pmovzxdq( X )	_mm_cvtepu32_epi64( (X) )
		#define XXH_sse_pmuludq( a, b )	_mm_mul_epu32( (a), (b) )
		#define XXH_sse_paddq( a, b )	_mm_add_epi64( (a), (b) )
		#define XXH_sse_psllqi( a, b )	_mm_slli_epi64( (a), (b) )
		#define XXH_sse_psrlqi( a, b )	_mm_srli_epi64( (a), (b) )

GCC:
	#define XXH_sse_loada( X )		(XXH_XMMWORD)__builtin_ia32_loaddqu( (const char*)(X) )
	#define XXH_sse_loadu( X )		(XXH_XMMWORD)__builtin_ia32_loaddqu( (const char*)(X) )
	#define XXH_sse_pmulld( a, b )	(XXH_XMMWORD)__builtin_ia32_pmulld128( (__v4si)(a), (__v4si)(b) )
	#define XXH_sse_paddd( a, b )	(XXH_XMMWORD)__builtin_ia32_paddd128( (__v4si)(a), (__v4si)(b) )
	#define XXH_sse_pslldi( a, b )	(XXH_XMMWORD)__builtin_ia32_pslldi128( (__v4si)(a), (b) )
	#define XXH_sse_psrldi( a, b )	(XXH_XMMWORD)__builtin_ia32_psrldi128( (__v4si)(a), (b) )
	#define XXH_sse_por( a, b )		__builtin_ia32_por128( (a), (b) )
	#define XXH_sse_stora( a, b )	__builtin_ia32_storedqu( (char*)(a), (__v16qi)(b) )
	#define XXH_sse_pshufd( a, b )	(XXH_XMMWORD)__builtin_ia32_pshufd( (__v4si)(a), (b) )
	#define XXH_sse_phaddd( a, b )	(XXH_XMMWORD)__builtin_ia32_phaddd128( (__v4si)(a), (__v4si)(b) )
	#define XXH_sse_pmovzxdq( X )	__builtin_ia32_pmovzxdq128( (__v4si)(X) )
	#define XXH_sse_pmuludq( a, b )	__builtin_ia32_pmuludq128( (__v4si)(a), (__v4si)(b) )
	#define XXH_sse_paddq( a, b )	__builtin_ia32_paddq128( (a), (b) )
	#define XXH_sse_psllqi( a, b )	__builtin_ia32_psllqi128( (a), (b) )
	#define XXH_sse_psrlqi( a, b )	__builtin_ia32_psrlqi128( (a), (b) )

czMaxx:
One compiler environment is still not satisfied with the fallthrough even though I rolled it back to the comment (have to take a look again). Btw: when I've been playing with the manual vectorization of XXH, my throughput test gave me these results:

> source\repos\xxhbench\Release>xxhbench32.exe
stream hash [CF9A41C4] computed in 0.33580 s (3049 MB/s)
SSE stream hash [CF9A41C4] computed in 0.20894 s (4901 MB/s)
stream hash [90E9D42C0E9C842E] computed in 0.52255 s (1960 MB/s)
SSE stream hash [90E9D42C0E9C842E] computed in 0.29240 s (3502 MB/s)
block hash [679FF1AD] computed in 0.18799 s (5447 MB/s)
SSE block hash [679FF1AD] computed in 0.19632 s (5216 MB/s)
block hash [FED919317D73CAF4] computed in 0.44707 s (2290 MB/s)
SSE block hash [FED919317D73CAF4] computed in 0.24585 s (4165 MB/s)

> source\repos\xxhbench\x64\Release>xxhbench64.exe
stream hash [CF9A41C4] computed in 0.24038 s (4260 MB/s)
SSE stream hash [CF9A41C4] computed in 0.20999 s (4876 MB/s)
stream hash [90E9D42C0E9C842E] computed in 0.15198 s (6738 MB/s)
SSE stream hash [90E9D42C0E9C842E] computed in 0.33661 s (3042 MB/s)
block hash [679FF1AD] computed in 0.18589 s (5509 MB/s)
SSE block hash [679FF1AD] computed in 0.20714 s (4943 MB/s)
block hash [FED919317D73CAF4] computed in 0.11186 s (9154 MB/s)
SSE block hash [FED919317D73CAF4] computed in 0.25969 s (3943 MB/s)

there might be a potential to speed up the 64bit version with AVX, which incorporates the rotation and there's no need to shift it twice and or the result. However, the support of the particular AVX instruction subset is limited to new processors and SSE is suitable for more HW configurations.

easyaspi314:
With regards to using builtins, definite no.

GCC may support these, but Clang does not. There is also no benefit over the manual intrinsics, either, as the Intel API is implemented with inline wrappers or vector extensions.

> there might be a potential to speed up the 64bit version with AVX, which incorporates the rotation and there's no need to shift it twice and or the result. 

Unlikely. All processors which support it also have the high latency vpmulld. It's just not going to work, as the ten cycles is astronomical in comparison to one cycle + ILP.

NEON does seem to benefit from vectorizing XXH32, especially 32-bit, which I'm considering if we add dispatching.

If @Cyan4973 is fine with including the SSE2, SSE4, and NEON XXH64 implementations, I can include it, but, like I said, unlike the SSE2/NEON XXH3 versions which benefit both 32-bit and 64-bit, this only benefits 32-bit. It just seems like a lot of code for little benefit.

For the ref, the SSE4 multiply code looks like this (with the same SSE2 code):
```c
/* { PRIME64_2 & 0xFFFFFFFF, PRIME64_2 >> 32 } */
const __m128i prime2 = _mm_set_epi32((int)PRIME64_2, (int)(PRIME64_2 >> 32), (int)PRIME64_2, (int)(PRIME64_2 >> 32));
/* { PRIME64_2 >> 32, PRIME64_2 & 0xFFFFFFFF } */
const __m128i prime2Rev = _mm_shuffle_epi32(prime1, _MM_SHUFFLE(2,3,0,1));

/* SSE2 grade school multiply
 * top: first op
 * botHi: { second op >> 32, ignored } 
 * botLo: { second op & 0xFFFFFFFF, ignored } */
XXH_FORCE_INLINE __m128i
XXH_U64x2_mult(__m128i top, const __m128i botHi,  const __m128i botLo)
{
     __m128i topHi, topHiBotLo, topLoBotHi, topLoBotLo, ret;
     /*  topHi = top >> 32; */
     topHi = _mm_shuffle_epi32(top, _MM_SHUFFLE(2,3,0,1));
     /* three multiplies:
      *     topLo * botLo
      *     topHi * botLo
      *     topLo * botHi
      * topHi * botHi is ignored as it will be higher than 64 bits
      */
     topLoBotLo = _mm_mul_epu32(top, botLo);
     topHiBotLo = _mm_mul_epu32(topHi, botLo);
     topLoBotHi = _mm_mul_epu32(top, botHi);
     /* add higher products, then shift left */
     ret = _mm_add_epi64(topHiBotLo, topLoBotHi);
     ret = _mm_slli_epi64(ret, 32);
     /* Add the low product */
     ret = _mm_add_epi64(ret, topLoBotLo);
     return ret;
}
     
/* SSE4.1 alternative multiply
 * top: first op
 * botRev: { second op >> 32, second op & 0xFFFFFFFF }
 * botLo: { second op & 0xFFFFFFFF, ignored } */
XXH_FORCE_INLINE __m128i
XXH_U64x2_mult(__m128i top, const __m128i botRev, const __m128i botLo)
{
    __m128i cross, zero, product, low_bits, high_bits, odd, even;
    /*
     * Calculate the two cross products, using our preswapped bot.
     * We only need a 32-bit multiply.
     *
     *     U32 cross[2] = { topLo * botHi, topHi * botLo };
     *
     */
    cross = _mm_mullo_epi32(top, botRev);

    /* Pairwise add high and low bits, then shift left.
     *  What we essentially do:
     *
     *      U64 tmp = (U64)cross[0] + cross[1];
     *      high_bits64 = tmp << 32;
     *
     * However, that is inefficient. Instead, we do this:
     *
     *     U64 high_bits64 = ((cross64 & 0xFFFFFFFF00000000) + (cross64 << 32)) & 0xFFFFFFFF00000000;
     *
     * or precisely:
     *
     *    U32 odd[2] = { 0, cross[1] };
     *    U64 even64 = cross64 << 32;
     *    U32 high_bits[2] = { odd[0] + even[0] /* both zero on little endian */, odd[1] + even[1] };
     */
    zero = _mm_setzero_si128();
    odd = _mm_blend_epi16(zero, cross, 204);
    even = _mm_slli_epi64(cross, 32);
    high_bits = _mm_add_epi32(odd, even);

    /* Multiply low bits */
    low_bits = _mm_mul_epu32(top, botLo);
    /* Add together */
    product = _mm_add_epi64(high_bits, low_bits);
    return ret;
}
```

However, like I said, you will only get faster code with Nehalem, Sandy Bridge, and Ivy Bridge (also potentially Penryn). The 10 cycle lag on _mm_mullo_epi32 in Haswell or later makes the SSE2 version faster.

Cyan4973:
I agree @easyaspi314 ,
we already embed a lot of complexity in this version,
it seems not worth it adding a narrow case for the time being.

easyaspi314:
Plus, the algorithm simply wasn't designed for 32-bit. 

It is basically like optimizing the bubble sort. 

You can spend months optimizing the hell out of it, doing all of the tricks like SIMD, unrolling, branch prediction, even write the whole thing in optimized assembly which is tuned to each CPU. 

Quick sort, merge sort, timsort, insertion sort, and even shell sort are still going to beat it almost every time.

Either way we optimize it, without a 64-bit multiply, the math is not in our favor.

Let's assume for now that 1 instruction == 1 cycle, ignore ILP/out of order execution, and let's take the best cases for scalar instruction count, ARMv7-a for 32 bit and aarch64 for 64-bit, all with infinite registers.

XXH64 is very easy for 64-bit:
```armasm
    ldr inp, [p]
    madd acc, inp, PRIME64_2, acc
    ror acc, acc, #33
    mul acc, acc, PRIME64_1
```
On our ideal system, 4 cycles per 8 bytes = 0.5 cpb.
    
For our 32-bit chip, it takes a LOT more cycles.
```armasm
    ldr inp_lo, [p]
    ldr inp_hi, [p, #4]
    umull product_lo, product_hi, PRIME64_2_lo, inp_lo
    mla product_hi, PRIME64_2_hi, inp_lo, product_hi
    mla product_hi, PRIME64_2_lo, inp_hi, product_hi
    adds acc_lo, product_lo, acc_lo
    adc acc_hi, product_hi, acc_hi
    lsl temp_hi, acc_hi, #31
    orr temp_hi, temp_hi, acc_lo, lsr #1
    lsl temp_lo, acc_lo, #31
    orr temp_lo, temp_lo, acc_hi, lsr #1
    umull acc_lo, acc_hi, PRIME64_2_lo, temp_lo
    mla acc_hi, PRIME64_2_hi, temp_lo, acc_hi
    mla acc_hi, PRIME64_1_lo, temp_hi, acc_hi
```
14 cycles per 8 bytes = 1.75 cpb.

Due to this extra arithmetic, even in an ideal case, XXH64 on 32-bit is going to be at least 3.5 times slower than XXH64 on 64-bit.

Take into account the stack swapping from running out of registers (i386 especially), realistic cycle counts, and the four lanes, this is actually a lot more.

Even using simd is still going to be slower than native arithmetic due to the latency.

XXH3 is designed to be good on 32-bit and 64-bit.

One XXH3_accumulate_512 round on ARM looks like this, assuming acc_hi and acc_lo are in registers (probably loaded beforehand with `ldm`)
```armasm
    ldr key_lo, [xkey]
    ldr key_hi, [xkey, #4]
    ldr data_lo, [xdata]
    ldr data_hi, [xdata, #4]
    adds acc_lo, data_lo, acc_lo
    adc acc_hi, data_hi, acc_hi
    eor data_lo, data_lo, key_lo
    eor data_hi, data_hi, key_hi
    umlal acc_lo, acc_hi, data_lo, data_hi
```
9 cycles per 8 bytes = 1.125 cpb.

Comparable to XXH32:
```armasm
    ldr inp, [p]
    mla acc, inp, PRIME32_2, acc
    ror acc, acc, #19
    mul acc, acc, PRIME32_1
```
4 cycles per 4 bytes = 1 cpb.

As you can see, XXH3 is naturally fast on 32-bit despite being a 64-bit hash. Combine it with the SIMD friendliness of the code (loads, 64-bit adds, 32->64 multiply and xors are available on SSE2, NEON, and 64-bit VSX) and it is worth optimizing.

czMaxx:
Doing just the multiplication in SSE makes no sense, that's obvious. But when it comes to do all the arithmetics in SSE, it speeds up the processing, because you can process 4-times more data in one shot than in 32bit native asm and the benefit is bigger than the penalty caused by the latencies (and this speed up works for almost all scenarios on x86). It should be up to the caller whether he wants to use the SSE version or not, thus it should not be the integral part of XXH32/64, but rather a parallel implementation. That's why I did not include the SSE code. But regarding the replacement of _aullmul, it's based on native asm and can get rid of the (4 * push + call __aullmul) overhead. You can argue that adding more complexity is not the way or that the 32bit build is irrelevant, because everything is supposed to be replaced with XXH3. I just ask why not to improve it, when the fix is so painless (just few lines of code and one ifdef)? The decision is up to you and I can simply roll the part back and keep just the dead code fix in the pull request.

czMaxx:
Just for the record, the SSE4 code derived from the original source may look like:

static const ALIGNED_ARR( uint32_t, xmm32prime1, 4, 16 ) = { 2654435761U, 2654435761U, 2654435761U, 2654435761U };

static const ALIGNED_ARR( uint32_t, xmm32prime2, 4, 16 ) = { 2246822519U, 2246822519U, 2246822519U, 2246822519U };

uint32_t SSE4_XXH32( const void* input, const uint32_t len, const uint32_t seed )
{
	if ( !input )
	{
		return 0;
	}

	const uint8_t* p = (const uint8_t*)input;

	if ( POINTER_WRAP( p, len ) )
	{
		return 0;
	}

	uint32_t h32;

	uint32_t u32Rounds = len >> 4U;

	if ( u32Rounds )
	{
		ALIGNED_ARR( uint32_t, vect, 4, 16 );

		vect[ 0 ] = seed + PRIME32_1 + PRIME32_2;
		vect[ 1 ] = seed + PRIME32_2;
		vect[ 2 ] = seed;
		vect[ 3 ] = seed - PRIME32_1;

		const XXH_XMMWORD cprm1 = xxh_sse_loada( (XXH_XMMWORD*)xmm32prime1 );
		const XXH_XMMWORD cprm2 = xxh_sse_loada( (XXH_XMMWORD*)xmm32prime2 );
		XXH_XMMWORD seed128 = xxh_sse_loada( (XXH_XMMWORD*)vect );

		while ( u32Rounds-- )
		{
			XXH_XMMWORD input128 = xxh_sse_loadu( (XXH_XMMWORD*)p );
			input128 = xxh_sse_pmulld( input128, cprm2 );
			seed128 = xxh_sse_paddd( seed128, input128 );
			XXH_XMMWORD temp = seed128;
			seed128 = xxh_sse_pslldi( seed128, 13 );
			temp = xxh_sse_psrldi( temp, 19 );
			seed128 = xxh_sse_por( seed128, temp );
			seed128 = xxh_sse_pmulld( seed128, cprm1 );
			p += 16U;
		}

		xxh_sse_stora( (XXH_XMMWORD*)vect, seed128 );

		h32 = xxh_rol32( vect[ 0 ], 1 ) + xxh_rol32( vect[ 1 ], 7 ) + xxh_rol32( vect[ 2 ], 12 ) + xxh_rol32( vect[ 3 ], 18 );
	}
	else
	{
		h32 = seed + PRIME32_5;
	}

	h32 += len;

	return XXH32_finalize( h32, p, len );
}

XXH_errorcode SSE4_XXH32_update( XXH32_state_t* const state, const void* input, uint32_t len )
{
	if ( !input )
	{
		return XXH_ERROR;
	}

	const uint8_t* p = (const uint8_t*)input;

	if ( POINTER_WRAP( p, len ) )
	{
		return XXH_ERROR;
	}

	state->total_len += len;

	const uint32_t u32Room = 16U - state->memsize;

	if ( u32Room > len )
	{
		xxh_SmallMemCpy( (void*)(state->mem8 + state->memsize), p, len );
		state->memsize += len;

		return XXH_OK;
	}

	const XXH_XMMWORD cprm1 = xxh_sse_loada( (XXH_XMMWORD*)xmm32prime1 );
	const XXH_XMMWORD cprm2 = xxh_sse_loada( (XXH_XMMWORD*)xmm32prime2 );
	XXH_XMMWORD seed = xxh_sse_loada( (XXH_XMMWORD*)&state->v1 );

	if ( state->memsize )
	{
		xxh_SmallMemCpy( (void*)(state->mem8 + state->memsize), p, u32Room );

		XXH_XMMWORD input128 = xxh_sse_loada( (XXH_XMMWORD*)state->mem8 );
		input128 = xxh_sse_pmulld( input128, cprm2 );
		seed = xxh_sse_paddd( seed, input128 );
		XXH_XMMWORD temp = seed;
		seed = xxh_sse_pslldi( seed, 13 );
		temp = xxh_sse_psrldi( temp, 19 );
		seed = xxh_sse_por( seed, temp );
		seed =xxh_sse_pmulld( seed, cprm1 );

		p += u32Room;
		len -= u32Room;
	}

	uint32_t u32Rounds = len >> 4U;

	while ( u32Rounds-- )
	{
		XXH_XMMWORD input128 = xxh_sse_loadu( (XXH_XMMWORD*)p );
		input128 = xxh_sse_pmulld( input128, cprm2 );
		seed = xxh_sse_paddd( seed, input128 );
		XXH_XMMWORD temp = seed;
		seed = xxh_sse_pslldi( seed, 13 );
		temp = xxh_sse_psrldi( temp, 19 );
		seed = xxh_sse_por( seed, temp );
		seed =xxh_sse_pmulld( seed, cprm1 );
		p += 16U;
	}

	xxh_sse_stora( (XXH_XMMWORD*)&state->v1, seed );

	state->memsize = len & 15U;

	xxh_SmallMemCpy( (void*)state->mem8, (void*)p, state->memsize );

	return XXH_OK;
}

static const ALIGNED_ARR( uint64_t, xmm64prime1, 2, 16 ) = { 11400714785074694791ULL, 11400714785074694791ULL };
static const ALIGNED_ARR( uint64_t, xmm64prime2, 2, 16 ) = { 14029467366897019727ULL, 14029467366897019727ULL };

uint64_t SSE4_XXH64( const void* input, const uint32_t len, const uint64_t seed )
{
	if ( !input )
	{
		return 0;
	}

	const uint8_t* p = (const uint8_t*)input;

	if ( POINTER_WRAP( p, len ) )
	{
		return 0;
	}

	uint64_t h64;

	uint32_t u32Rounds = len >> 5U;

	if ( u32Rounds )
	{
		ALIGNED_ARR( uint64_t, vect, 4, 16 );

		vect[ 0 ] = seed + PRIME64_1 + PRIME64_2;
		vect[ 1 ] = seed + PRIME64_2;
		vect[ 2 ] = seed;
		vect[ 3 ] = seed - PRIME64_1;

		const XXH_XMMWORD cprm1 = xxh_sse_loada( (XXH_XMMWORD*)xmm64prime1 );
		const XXH_XMMWORD cprm2 = xxh_sse_loada( (XXH_XMMWORD*)xmm64prime2 );
		XXH_XMMWORD seed1 = xxh_sse_loada( (XXH_XMMWORD*)&vect[ 0 ] );
		XXH_XMMWORD seed2 = xxh_sse_loada( (XXH_XMMWORD*)&vect[ 2 ] );

		while ( u32Rounds-- )
		{
			XXH_XMMWORD input128 = xxh_sse_loadu( (XXH_XMMWORD*)p );
			XXH_XMMWORD temp = xxh_sse_pmulld( input128, xxh_sse_pshufd( cprm2, 0xB1 ) );
			temp = xxh_sse_phaddd( temp, temp );
			temp = xxh_sse_pmovzxdq( temp );
			temp = xxh_sse_pshufd( temp, 0xB1 );
			input128 = xxh_sse_pmuludq( input128, cprm2 );
			input128 = xxh_sse_paddq( input128, temp );
			seed1 = xxh_sse_paddq( seed1, input128 );
			temp = seed1;
			seed1 = xxh_sse_psllqi( seed1, 31 );
			temp = xxh_sse_psrlqi( temp, 33 );
			seed1 = xxh_sse_por( seed1, temp );
			temp = xxh_sse_pmulld( seed1, xxh_sse_pshufd( cprm1, 0xB1 ) );
			temp = xxh_sse_phaddd( temp, temp );
			temp = xxh_sse_pmovzxdq( temp );
			temp = xxh_sse_pshufd( temp, 0xB1 );
			seed1 = xxh_sse_pmuludq( seed1, cprm1 );
			seed1 = xxh_sse_paddq( seed1, temp );
			p += 16U;
			input128 = xxh_sse_loadu( (XXH_XMMWORD*)p );
			temp = xxh_sse_pmulld( input128, xxh_sse_pshufd( cprm2, 0xB1 ) );
			temp = xxh_sse_phaddd( temp, temp );
			temp = xxh_sse_pmovzxdq( temp );
			temp = xxh_sse_pshufd( temp, 0xB1 );
			input128 = xxh_sse_pmuludq( input128, cprm2 );
			input128 = xxh_sse_paddq( input128, temp );
			seed2 = xxh_sse_paddq( seed2, input128 );
			temp = seed2;
			seed2 = xxh_sse_psllqi( seed2, 31 );
			temp = xxh_sse_psrlqi( temp, 33 );
			seed2 = xxh_sse_por( seed2, temp );
			temp = xxh_sse_pmulld( seed2, xxh_sse_pshufd( cprm1, 0xB1 ) );
			temp = xxh_sse_phaddd( temp, temp );
			temp = xxh_sse_pmovzxdq( temp );
			temp = xxh_sse_pshufd( temp, 0xB1 );
			seed2 = xxh_sse_pmuludq( seed2, cprm1 );
			seed2 = xxh_sse_paddq( seed2, temp );
			p += 16U;
		}

		xxh_sse_stora( (XXH_XMMWORD*)&vect[ 0 ], seed1 );
		xxh_sse_stora( (XXH_XMMWORD*)&vect[ 2 ], seed2 );

		h64 = xxh_rol64( vect[ 0 ], 1 ) + xxh_rol64( vect[ 1 ], 7 ) + xxh_rol64( vect[ 2 ], 12 ) + xxh_rol64( vect[ 3 ], 18 );

		h64 = XXH64_mergeRound( h64, vect[ 0 ] );
		h64 = XXH64_mergeRound( h64, vect[ 1 ] );
		h64 = XXH64_mergeRound( h64, vect[ 2 ] );
		h64 = XXH64_mergeRound( h64, vect[ 3 ] );
	}
	else
	{
		h64 = seed + PRIME64_5;
	}

	h64 += len;

	return XXH64_finalize( h64, p, len );
}

XXH_errorcode SSE4_XXH64_update( XXH64_state_t* const state, const void* input, uint32_t len )
{
	if ( !input )
	{
		return XXH_ERROR;
	}

	const uint8_t* p = (const uint8_t*)input;

	if ( POINTER_WRAP( p, len ) )
	{
		return XXH_ERROR;
	}

	state->total_len += len;

	const uint32_t u32Room = 32U - state->memsize;

	if ( u32Room > len )
	{
		xxh_SmallMemCpy( (void*)(state->mem8 + state->memsize), p, len );
		state->memsize += len;

		return XXH_OK;
	}

	const XXH_XMMWORD cprm1 = xxh_sse_loada( (XXH_XMMWORD*)xmm64prime1 );
	const XXH_XMMWORD cprm2 = xxh_sse_loada( (XXH_XMMWORD*)xmm64prime2 );
	XXH_XMMWORD seed1 = xxh_sse_loada( (XXH_XMMWORD*)&state->v1 );
	XXH_XMMWORD seed2 = xxh_sse_loada( (XXH_XMMWORD*)&state->v3 );

	if ( state->memsize )
	{
		xxh_SmallMemCpy( (void*)(state->mem8 + state->memsize), p, u32Room );

		XXH_XMMWORD input128 = xxh_sse_loadu( (XXH_XMMWORD*)state->mem8 );
		XXH_XMMWORD temp = xxh_sse_pmulld( input128, xxh_sse_pshufd( cprm2, 0xB1 ) );
		temp = xxh_sse_phaddd( temp, temp );
		temp = xxh_sse_pmovzxdq( temp );
		temp = xxh_sse_pshufd( temp, 0xB1 );
		input128 = xxh_sse_pmuludq( input128, cprm2 );
		input128 = xxh_sse_paddq( input128, temp );
		seed1 = xxh_sse_paddq( seed1, input128 );
		temp = seed1;
		seed1 = xxh_sse_psllqi( seed1, 31 );
		temp = xxh_sse_psrlqi( temp, 33 );
		seed1 = xxh_sse_por( seed1, temp );
		temp = xxh_sse_pmulld( seed1, xxh_sse_pshufd( cprm1, 0xB1 ) );
		temp = xxh_sse_phaddd( temp, temp );
		temp = xxh_sse_pmovzxdq( temp );
		temp = xxh_sse_pshufd( temp, 0xB1 );
		seed1 = xxh_sse_pmuludq( seed1, cprm1 );
		seed1 = xxh_sse_paddq( seed1, temp );
		input128 = xxh_sse_loadu( (XXH_XMMWORD*)(state->mem8 + 16U) );
		temp = xxh_sse_pmulld( input128, xxh_sse_pshufd( cprm2, 0xB1 ) );
		temp = xxh_sse_phaddd( temp, temp );
		temp = xxh_sse_pmovzxdq( temp );
		temp = xxh_sse_pshufd( temp, 0xB1 );
		input128 = xxh_sse_pmuludq( input128, cprm2 );
		input128 = xxh_sse_paddq( input128, temp );
		seed2 = xxh_sse_paddq( seed2, input128 );
		temp = seed2;
		seed2 = xxh_sse_psllqi( seed2, 31 );
		temp = xxh_sse_psrlqi( temp, 33 );
		seed2 = xxh_sse_por( seed2, temp );
		temp = xxh_sse_pmulld( seed2, xxh_sse_pshufd( cprm1, 0xB1 ) );
		temp = xxh_sse_phaddd( temp, temp );
		temp = xxh_sse_pmovzxdq( temp );
		temp = xxh_sse_pshufd( temp, 0xB1 );
		seed2 = xxh_sse_pmuludq( seed2, cprm1 );
		seed2 = xxh_sse_paddq( seed2, temp );

		p += u32Room;
		len -= u32Room;
	}

	uint32_t u32Rounds = len >> 5U;

	while ( u32Rounds-- )
	{
		XXH_XMMWORD input128 = xxh_sse_loadu( (XXH_XMMWORD*)p );
		XXH_XMMWORD temp = xxh_sse_pmulld( input128, xxh_sse_pshufd( cprm2, 0xB1 ) );
		temp = xxh_sse_phaddd( temp, temp );
		temp = xxh_sse_pmovzxdq( temp );
		temp = xxh_sse_pshufd( temp, 0xB1 );
		input128 = xxh_sse_pmuludq( input128, cprm2 );
		input128 = xxh_sse_paddq( input128, temp );
		seed1 = xxh_sse_paddq( seed1, input128 );
		temp = seed1;
		seed1 = xxh_sse_psllqi( seed1, 31 );
		temp = xxh_sse_psrlqi( temp, 33 );
		seed1 = xxh_sse_por( seed1, temp );
		temp = xxh_sse_pmulld( seed1, xxh_sse_pshufd( cprm1, 0xB1 ) );
		temp = xxh_sse_phaddd( temp, temp );
		temp = xxh_sse_pmovzxdq( temp );
		temp = xxh_sse_pshufd( temp, 0xB1 );
		seed1 = xxh_sse_pmuludq( seed1, cprm1 );
		seed1 = xxh_sse_paddq( seed1, temp );
		p += 16U;
		input128 = xxh_sse_loadu( (XXH_XMMWORD*)p );
		temp = xxh_sse_pmulld( input128, xxh_sse_pshufd( cprm2, 0xB1 ) );
		temp = xxh_sse_phaddd( temp, temp );
		temp = xxh_sse_pmovzxdq( temp );
		temp = xxh_sse_pshufd( temp, 0xB1 );
		input128 = xxh_sse_pmuludq( input128, cprm2 );
		input128 = xxh_sse_paddq( input128, temp );
		seed2 = xxh_sse_paddq( seed2, input128 );
		temp = seed2;
		seed2 = xxh_sse_psllqi( seed2, 31 );
		temp = xxh_sse_psrlqi( temp, 33 );
		seed2 = xxh_sse_por( seed2, temp );
		temp = xxh_sse_pmulld( seed2, xxh_sse_pshufd( cprm1, 0xB1 ) );
		temp = xxh_sse_phaddd( temp, temp );
		temp = xxh_sse_pmovzxdq( temp );
		temp = xxh_sse_pshufd( temp, 0xB1 );
		seed2 = xxh_sse_pmuludq( seed2, cprm1 );
		seed2 = xxh_sse_paddq( seed2, temp );
		p += 16U;
	}

	xxh_sse_stora( (XXH_XMMWORD*)&state->v1, seed1 );
	xxh_sse_stora( (XXH_XMMWORD*)&state->v3, seed2 );

	state->memsize = len & 31U;

	xxh_SmallMemCpy( (void*)state->mem8, (void*)p, state->memsize );

	return XXH_OK;
}

where the XXH64 part is ugly, because there's no pmullq, which would make it straightforward just like the XXH32. However, the benefit of processing 128bits in one round is still bigger and the routine is faster than the native x86. AVX implementation above the 256bit registers would increase the throughput even further, but the question is whether someone actually wants to run a 32bit build on a platform with the necessary AVX version.

czMaxx:
Oh, the formatting is weird. Sorry

easyaspi314:
Yeah, but trust me. SSE does not improve XXH32. We have tried :confused:

Maybe on AVX2-128 it could make a difference, but I doubt it.

If you haven't already, read my UGLY HACK comment in XXH32_round. It just doesn't work out.

Let's take Skylake.

3 cycles for `movdqu`
12 cycles for `pmulld`
1 cycle for `paddd`
Assume movdqa is ignored
1 cycle for `pslld`
1 cycle for `pslld`
1 cycle for `por`
12 cycles for `pmulld`

31 cycles.

Now scalar:

4 cycles for `imul`
1 cycle for `add`
1 cycle for `rol`
4 cycles for `imul`

10, times 4 = 40 cycles.

However, instruction level parallelism makes this faster. 

All x86_64 chips are out-of-order processors.

If the chip recognizes that two instructions don't depend on each other, it can execute the instructions in parallel (i.e. running between the latency cycles, running on different cores, all sorts of crazy witchcraft), and by writing it in SSE4.1, you are actually scalarizing it because each instruction depends on the next.

ARM relies less on out of order (it is a relatively new feature introduced on the Cortex-A15), and `vmul.i32` is not a high latency instruction, so ILP has much less of a benefit than NEON. That is why the NEON path is faster.

czMaxx:
From my benchmarking, it actually does, but only the stream version (I still did not have time to check it with vTune to spot the real bottleneck - sometimes it's not the expected arithmetics block, but the surrounding code which somehow fits better into the insn fetcher/decoder or there's another unexpected magic).

Lynnfield (i7-860):
stream hash [CF9A41C4] computed in 0.36177 s (2830 MB/s)
SSE stream hash [CF9A41C4] computed in 0.33660 s (3042 MB/s)
stream hash [90E9D42C0E9C842E] computed in 0.67992 s (1506 MB/s)
SSE stream hash [90E9D42C0E9C842E] computed in 0.49775 s (2057 MB/s)
block hash [679FF1AD] computed in 0.25439 s (4025 MB/s)
SSE block hash [679FF1AD] computed in 0.31893 s (3211 MB/s)
block hash [FED919317D73CAF4] computed in 0.63002 s (1625 MB/s)
SSE block hash [FED919317D73CAF4] computed in 0.41982 s (2439 MB/s)

Sandy bridge (Xeon E3-1245):
stream hash [CF9A41C4] computed in 0.33235 s (3081 MB/s)
SSE stream hash [CF9A41C4] computed in 0.20926 s (4894 MB/s)
stream hash [90E9D42C0E9C842E] computed in 0.51995 s (1969 MB/s)
SSE stream hash [90E9D42C0E9C842E] computed in 0.40189 s (2548 MB/s)
block hash [679FF1AD] computed in 0.18349 s (5581 MB/s)
SSE block hash [679FF1AD] computed in 0.19391 s (5281 MB/s)
block hash [FED919317D73CAF4] computed in 0.44806 s (2285 MB/s)
SSE block hash [FED919317D73CAF4] computed in 0.24441 s (4190 MB/s)

I absolutely agree, that beating the performance of 3-operand imul is probably impossible.

easyaspi314:
Sandy Bridge and Lynnfield (Nehalem family) have a very fast pmulld (only 5 cycles latency), so their SSE4.1 XXH32 is going to be pretty fast.

However, Haswell and later have 10 cycles of latency, which is our main problem. While the Nehalem, Sandy, and Ivy Bridge families take up a huge chunk of the consumer market, any newer processors suffer from it, and the detection for the best algo is super complicated.

czMaxx:
CannonLake:
stream hash [CF9A41C4] computed in 0.26344 s (3887 MB/s)
SSE stream hash [CF9A41C4] computed in 0.30219 s (3389 MB/s)
stream hash [90E9D42C0E9C842E] computed in 0.40090 s (2554 MB/s)
SSE stream hash [90E9D42C0E9C842E] computed in 0.28709 s (3567 MB/s)
block hash [679FF1AD] computed in 0.17261 s (5932 MB/s)
SSE block hash [679FF1AD] computed in 0.27785 s (3685 MB/s)
block hash [FED919317D73CAF4] computed in 0.34587 s (2961 MB/s)
SSE block hash [FED919317D73CAF4] computed in 0.24794 s (4130 MB/s)

CoffeeLake:
stream hash [CF9A41C4] computed in 0.17960 s (5702 MB/s)
SSE stream hash [CF9A41C4] computed in 0.20777 s (4929 MB/s)
stream hash [90E9D42C0E9C842E] computed in 0.27846 s (3677 MB/s)
SSE stream hash [90E9D42C0E9C842E] computed in 0.19850 s (5159 MB/s)
block hash [679FF1AD] computed in 0.12234 s (8370 MB/s)
SSE block hash [679FF1AD] computed in 0.20088 s (5098 MB/s)
block hash [FED919317D73CAF4] computed in 0.24541 s (4173 MB/s)
SSE block hash [FED919317D73CAF4] computed in 0.17324 s (5911 MB/s)

SkyLake:
stream hash [CF9A41C4] computed in 0.29083 s (3521 MB/s)
SSE stream hash [CF9A41C4] computed in 0.33418 s (3064 MB/s)
stream hash [90E9D42C0E9C842E] computed in 0.48059 s (2131 MB/s)
SSE stream hash [90E9D42C0E9C842E] computed in 0.37403 s (2738 MB/s)
block hash [679FF1AD] computed in 0.19223 s (5327 MB/s)
SSE block hash [679FF1AD] computed in 0.31053 s (3298 MB/s)
block hash [FED919317D73CAF4] computed in 0.38735 s (2644 MB/s)
SSE block hash [FED919317D73CAF4] computed in 0.28073 s (3648 MB/s)

You're right and I've made a wrong estimation when I thought that the SSE can't be so terribly slow on newer architectures to completely screw the benefit of processing all 4 pipelines in one round.

czMaxx:
Haswell:
stream hash [CF9A41C4] computed in 0.31486 s (3252 MB/s)
SSE stream hash [CF9A41C4] computed in 0.28814 s (3554 MB/s)
stream hash [90E9D42C0E9C842E] computed in 0.40215 s (2546 MB/s)
SSE stream hash [90E9D42C0E9C842E] computed in 0.28295 s (3619 MB/s)
block hash [679FF1AD] computed in 0.16532 s (6194 MB/s)
SSE block hash [679FF1AD] computed in 0.27712 s (3695 MB/s)
block hash [FED919317D73CAF4] computed in 0.32878 s (3115 MB/s)
SSE block hash [FED919317D73CAF4] computed in 0.26844 s (3815 MB/s)

KabyLake:
stream hash [CF9A41C4] computed in 0.28494 s (3594 MB/s)
SSE stream hash [CF9A41C4] computed in 0.30995 s (3304 MB/s)
stream hash [90E9D42C0E9C842E] computed in 0.43545 s (2352 MB/s)
SSE stream hash [90E9D42C0E9C842E] computed in 0.36625 s (2796 MB/s)
block hash [679FF1AD] computed in 0.17962 s (5701 MB/s)
SSE block hash [679FF1AD] computed in 0.31578 s (3243 MB/s)
block hash [FED919317D73CAF4] computed in 0.36237 s (2826 MB/s)
SSE block hash [FED919317D73CAF4] computed in 0.25510 s (4014 MB/s)

Ivy bridge has faster SSE as well. We can conclude that SSE versions can only speed up the XXH64 on 32bit. The demand for such version is therefore questionable. However, there's still a chance to at least use the fast/inline multiply instead of _aullmul and to use __movsb intrinsic to enforce an inline memcpy in both stream versions (XXH32 & XXH64).

Cyan4973:
The code has evolved a lot since that patch,
so I guess it would require an update to circumvent all conflicts.

PR #203, opened 2019-05-06
NDEBUG

dnbaker:
`#undef NDEBUG` at the end of xxh3.h so as to not interfere with other code intentionally not defining NDEBUG. (IE, for debug builds of software using xxh3.)


Cyan4973:
Yes, thanks @dnbaker , preserving `NDEBUG` is a very nice feature.

PR #204, opened 2019-05-06
Updated xxh128 formula

Cyan4973:
Input now influences 128-bit of internal accumulator,
which requires 2 passes.
Also, ingestion is now bijective.
It comes with increased difficulties to pass sparse distribution tests, 
but the second pass seems to be the right ingredient which solves the issue.

Finally, in contrast with the 64-bit variant, AVX2 is not 2x faster.
It's more in the +40% faster range.

This might be because accumulator bits are now continuously mixed, which creates a stronger dependency between rounds, and there might not be enough work to do in parallel, in spite of the 512-bit wide accumulator.
Maybe enlarging accumulators to 1024-bit might help this issue, if diagnosis is correct.


Cyan4973:
> Maybe enlarging accumulators to 1024-bit might help this issue, if diagnosis is correct.

I tried increasing accumulator size to 1024-bit,
and if the tests are correct,
while it increases the speed difference between AVX2 and SSE2 version,
it most importantly makes the SSE2 variant slower, by 10-20%,
maybe as a consequence to shuffling vectors in and out of RAM more often.

So it's not a good lead.

Cyan4973:
OK, so the test was actually not good.
I had removed some `force_inline` statements for debugging purpose, and forgot to enable them again.

When I re-enable `force_inline` correctly, I'm getting some strange outcome : 
The `clang + SSE2` path takes a huge plunge in performance (almost -50%), 
while everything else seems to benefit.

`clang Apple LLVM 10.0.1`:

| __XXH_VECTOR__ | 512 bit acc | 1024 bit acc |
| --- | --- | --- |
| __0__ (scalar) | 6.7 GB/s | 7.4 GB/s |
| __1__ (SSE2) | 15.0 GB/s | ___7.8___ GB/s |
| __2__ (AVX2) | 22.0 GB/s | 31.0 GB/s |

`gcc-8` : 

| __XXH_VECTOR__ | 512 bit acc | 1024 bit acc |
| --- | --- | --- |
| __0__ (scalar) | 7.0 GB/s | 7.1 GB/s |
| __1__ (SSE2) | 13.3 GB/s | 13.9 GB/s |
| __2__ (AVX2) | 19.3 GB/s | 27.8 GB/s |

So, if it was not for the specific `clang + SSE2` combination, it would appear to be positive.

But unfortunately, the cost for `clang +SSE2` is really huge,
especially as this is more or less the target version
(speed on `AVX2` is "nice", but not the main target, due to AVX down-clocking issues).

The 1024-bit accumulator can be tested in branch `xxh1024`.

easyaspi314:
I believe it has to do with the `#pragma clang loop unroll(enable)`.

`#pragma clang loop unroll(enable)` will always completely unroll a loop with a fixed number of iterations, whether it is 1, 10, or 10000. It might be causing cache misses.

Try replacing it with `unroll_count(N)` and tweak with the N.

easyaspi314:
Yeaaaaah.......

XXH3_hashLong_128b alone is 8366 bytes of code for SSE2.... that is sure to be a lot of cache pressure.

It also does a lot of operations directly in memory as I guess it runs out of SSE registers.

As for NEON, it seems to be slightly slower on my G3, although it isn't stable enough.

Cyan4973:
> I believe it has to do with the #pragma clang loop unroll(enable).

You are right @easyaspi314 .
Just commenting out the `#pragma`, and performance goes back to "normal" levels (~15 GB/s).
The 1024 bits accumulator is not faster (than the 512 bits one) with SSE2, just the same performance.

> Try replacing it with unroll_count(N) and tweak with the N.

Any value of `N` I could try just worsens performance,
starting with 2, and getting worse as `N` increases.

Interestingly, `unroll(enable)` _does_ improve performance for `AVX2`. Not by a huge lot, but still (from 28 GB/s to 31 GB/s).

So I guess it makes the `#pragma` dependent on the vectorization path ?
I may become a pain point for a potential multi-target extension.

Cyan4973:
OK, it has been some time,
now there is a need to select a variant for `xxh128` : 512 or 1024 bits accumulators ?

As measured previously, 1024 bits improve `AVX2` speed substantially,
but only `AVX2`. Impact on `scalar` and `SSE2` is more limited, almost neutral (generally slightly positive, sometimes slightly negative).

This is important because the most important variant (for `x64`) is the `SSE2` mode, presumed always available, hence always selected by default.

There is another downside worth mentioning : with more accumulators, it makes the final stage (merging) more expensive. It doesn't matter much for large files (like the standard 100KB test), but for smaller ones, it is perceptible : up to ~20% speed difference on "small" data (< 1 KB), then the effects fades away with larger sizes. It's about unnoticeable circa 16 KB.

_Note_: very small inputs are treated differently, increased accumulator size does not matter there.

Overall, I believe I lean towards the 1024 bits,
on the following ground : 
- It will also help speed of a potential future AVX512 variant, should it exist one day
- `AVX2` can be useful for `xxh128`, because the `SSE2` variant does not reach "RAM speed limit". Data already stored in RAM will benefit speed when hashed by the `AVX2` variant.

The main downside is that lower power devices, with limited RAM capabilities, will need more RAM to store accumulators. I think it's fair to say `xxh128` is not optimized for such scenario, and merely try to remain compatible / achievable on such platforms.

easyaspi314:
Sorry I've been inactive, finals have been going on.

On my Core 2 Duo in WSL Ubuntu 18.04 with Clang 8
```
./xxhsum 0.7.0 (64-bits x86_64 + SSE2 little endian), Clang 8.0.0 (tags/RELEASE_800/final), by Yann Collet
Sample of 100 KB...
XXH32               :     102400 ->    53319 it/s ( 5206.9 MB/s)
XXH32 unaligned     :     102400 ->    40591 it/s ( 3964.0 MB/s)
XXH64               :     102400 ->    65536 it/s ( 6400.0 MB/s)
XXH64 unaligned     :     102400 ->    39322 it/s ( 3840.0 MB/s)
XXH3_64bits         :     102400 ->    98304 it/s ( 9600.0 MB/s)
XXH3_64b unaligned  :     102400 ->    65536 it/s ( 6400.0 MB/s)
XXH128              :     102400 ->    39947 it/s ( 3901.1 MB/s)
XXH128 unaligned    :     102400 ->    32768 it/s ( 3200.0 MB/s)
```

**XXH1024 can't even beat XXH32.**

Cyan4973:
An interesting comparison would be between the 512-bit variant (branch `xxh128`) and the 1024-bit one (branch `xxh1024`).

We know that `xxh128` is slower, it is required by the 2-passes mechanism. The goal is only to make it "decent". On modern cpus (laptop and desktop), it seems to be the case : the `SSE2` variant has roughly the same speed as `XXH64`.

I'm afraid the characteristics of Core 2 Duo are a bit too specific by now, modern cpus are designed with different trade-offs, so it's not reasonable to optimize _for_ core2 platform.

Cyan4973:
OK, time to move on.
I believe the 1024-bit variant is valuable, thanks to its speed gains at `AVX2` variant and up, and its drawbacks seem comparatively acceptable.
So I'll merge it into `xxh128` and make it the new proposal for the 128-bit variant.

Cyan4973:
I'm scrapping this design.

It introduces too much speed regression, and too much specific code, on top of more expensive memory usage.

I'll go for a much simpler design, closer to original, sharing a lot more common code with the `_64bit` variant, and hopefully preserving most of the speed.

PR #206, opened 2019-05-08
update intrinsic include policy to be compatible with gcc < 4.4 

### LINKED ISSUE #205: XXH3 doesn't compile with (old) GCC EL5
[rpmbuild@rhel5-x86-64 xxHash-0.7.0]$ gcc -v
Using built-in specs.
Target: x86_64-redhat-linux
Configured with: ../configure --prefix=/usr --mandir=/usr/share/man --infodir=/usr/share/info --enable-shared --enable-threads=posix --enable-checking=release --with-system-zlib --enable-__cxa_atexit --disable-libunwind-exceptions --enable-libgcj-multifile --enable-languages=c,c++,objc,obj-c++,java,fortran,ada --enable-java-awt=gtk --disable-dssi --disable-plugin --with-java-home=/usr/lib/jvm/java-1.4.2-gcj-1.4.2.0/jre --with-cpu=generic --host=x86_64-redhat-linux
Thread model: posix
gcc version 4.1.2 20080704 (Red Hat 4.1.2-55)

[rpmbuild@rhel5-x86-64 xxHash-0.7.0]$ make
cc -O3    -c -o xxhash.o xxhash.c
In file included from xxhash.c:1017:
xxh3.h:62:27: error: x86intrin.h: No such file or directory
In file included from xxhash.c:1017:
xxh3.h: In function ‘XXH3_accumulate_512’:
xxh3.h:384: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘*’ token
xxh3.h:384: error: expected expression before ‘const’
xxh3.h:385: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘*’ token
xxh3.h:385: error: expected expression before ‘const’
xxh3.h:386: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘*’ token
xxh3.h:386: error: expected expression before ‘const’
xxh3.h:389: error: ‘__m128i’ undeclared (first use in this function)
xxh3.h:389: error: (Each undeclared identifier is reported only once
xxh3.h:389: error: for each function it appears in.)
xxh3.h:390: error: expected ‘;’ before ‘const’
xxh3.h:391: error: expected ‘;’ before ‘const’
xxh3.h:392: error: expected ‘;’ before ‘const’
xxh3.h:393: error: expected ‘;’ before ‘const’
xxh3.h:394: error: expected ‘;’ before ‘const’
xxh3.h:395: error: ‘xacc’ undeclared (first use in this function)
xxh3.h:395: error: ‘res’ undeclared (first use in this function)
xxh3.h:395: error: ‘add’ undeclared (first use in this function)
xxh3.h: In function ‘XXH3_scrambleAcc’:
xxh3.h:483: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘*’ token
xxh3.h:483: error: expected expression before ‘const’
xxh3.h:484: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘*’ token
xxh3.h:484: error: expected expression before ‘const’
xxh3.h:487: error: ‘__m128i’ undeclared (first use in this function)
xxh3.h:488: error: expected ‘;’ before ‘data’
xxh3.h:489: error: expected ‘;’ before ‘const’
xxh3.h:490: error: ‘data’ undeclared (first use in this function)
xxh3.h:490: error: ‘shifted’ undeclared (first use in this function)
xxh3.h:492: error: expected ‘;’ before ‘const’
xxh3.h:493: error: expected ‘;’ before ‘const’
xxh3.h:495: error: expected ‘;’ before ‘const’
xxh3.h:496: error: expected ‘;’ before ‘const’
xxh3.h:497: error: expected ‘;’ before ‘const’
xxh3.h:499: error: ‘xacc’ undeclared (first use in this function)
xxh3.h:499: error: ‘dk’ undeclared (first use in this function)
xxh3.h:499: error: ‘dk2’ undeclared (first use in this function)
make: *** [xxhash.o] Error 1

Cyan4973:
fix #205

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -61,13 +61,10 @@
 #endif
 
 #if defined(__GNUC__)
-#  if (__GNUC__ < 4) || ((__GNUC__ == 4) && (__GNUC_MINOR__ < 4))
-#    /* x86intrin.h was introduced in gcc 4.4 */
-#    undef __SSE2__
-#    undef __AVX2__
-#  endif
-#  if defined(__SSE2__)
-#    include <x86intrin.h>
+#  if defined(__AVX2__)
+#    include <immintrin.h>
+#  elif defined(__SSE2__)
+#    include <emmintrin.h>
 #  elif defined(__ARM_NEON__) || defined(__ARM_NEON)
 #    define inline __inline__ /* clang bug */
 #    include <arm_neon.h>
--- a/xxh3.h
+++ b/xxh3.h
@@ -61,6 +61,11 @@
 #endif
 
 #if defined(__GNUC__)
+#  if (__GNUC__ < 4) || ((__GNUC__ == 4) && (__GNUC_MINOR__ < 4))
+#    /* x86intrin.h was introduced in gcc 4.4 */
+#    undef __SSE2__
+#    undef __AVX2__
+#  endif
 #  if defined(__SSE2__)
 #    include <x86intrin.h>
 #  elif defined(__ARM_NEON__) || defined(__ARM_NEON)


easyaspi314:
Woah, hold your horses.

Why are we using the non-standard x86intrin.h when we can use the standardized immintrin.h?

Cyan4973:
No idea, I don't remember where this comes from.

While I don't mind changing the `#include` if that's all there is, 
I wonder what source makes you state that one is "standardized",
implying that the other is not ?

easyaspi314:
By "standardized" I mean it is required to be compatible with Intel's intrinsic specifications. I think AMD also requires it, but it would be stupid not to because it would break things.

emmintrin, pmmintrin, zmmintrin, etc are target specific, immintrin covers all intrinsics.

While I haven't tested, GCC 4.1 should also have that header.

jccleaver:
As mentioned in the issue, this seems to fix it. If there are other headers you want me to try, I can run those too.

Cyan4973:
I can update the patch to `#include <emmintrin.h>` instead on older compilers (`gcc` before `4.4.0`).

For more recent ones, I keep finding both `x86intrin.h` and `immintrin.h`, with no clear preference nor explanation on which one is "better", and why. They seem to achieve basically the same thing.

Cyan4973:
Updated the patch, to `#include` `emmintrin.h` on `SSE2`,
and `immintrin.h` on `AVX2`.

My understanding is that `gcc` 4.1 cannot be in the second case, so it will include `emmintrin.h`, which is available.

The advantage is that it can benefit for `SSE2` code path.

Cyan4973:
OK, let's go with this latest version and move on.

jccleaver:
Sorry; Yep this version compiled clean for me! Thanks again.

PR #207, opened 2019-05-17
Add configuration to reroll most loops.

easyaspi314:
`XXH_REROLL`: Rerolls the following functions:
 - `XXH32_finalize`
 - `XXH64_finalize`
 - `XXH3_64bitsWithSeed`
 - `XXH3_128bitsWithSeed`
 - `XXH3_accumulate` (disables pragma)

Automatically enabled with `__OPTIMIZE_SIZE__`.

`XXH_REROLL_XXH64`: Specifically controls rerolling `XXH64_finalize`. This is automatically defined on non-64-bit targets, as the unroll makes the binary huge (`XXH64_finalize` on i386 alone is the largest function in the program), and it actually slows things down.

If `XXH_REROLL` is defined, this is ignored and the loop is rerolled anyways.

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -666,7 +666,7 @@ static void XXH3_accumulate(U64* acc, const void* restrict data, const U32* rest
 {
     size_t n;
     /* Clang doesn't unroll this loop without the pragma. Unrolling can be up to 1.4x faster. */
-#if defined(__clang__) && !defined(__OPTIMIZE_SIZE__) && !defined(__ARM_ARCH)
+#if defined(__clang__) && !defined(__ARM_ARCH) && !XXH_REROLL
 #  pragma clang loop unroll(enable)
 #endif
     for (n = 0; n < nbStripes; n++ ) {
@@ -774,28 +774,33 @@ XXH3_64bits_withSeed(const void* data, size_t len, XXH64_hash_t seed)
     const char* const key = (const char*)kKey;
 
     if (len <= 16) return XXH3_len_0to16_64b(data, len, seed);
-
+    if (len > 128) return XXH3_hashLong_64b(data, len, seed);
     {   U64 acc = len * PRIME64_1;
-        if (len > 32) {
-            if (len > 64) {
-                if (len > 96) {
-                    if (len > 128) return XXH3_hashLong_64b(data, len, seed);
+        if (XXH_REROLL) {
+            size_t i = ((len & 127) - 1) / 32;
+            do {
+                acc += XXH3_mix16B(p + (i * 16), key + (i * 32), seed);
+                acc += XXH3_mix16B(p + len - ((i * 16) + 16), key + (i * 32) + 16, seed);
+            } while (i-- > 0);
+        } else {
+            if (len > 32) {
+                if (len > 64) {
+                    if (len > 96) {
+                        acc += XXH3_mix16B(p+48, key+96, seed);
+                        acc += XXH3_mix16B(p+len-64, key+112, seed);
+                    }
 
-                    acc += XXH3_mix16B(p+48, key+96, seed);
-                    acc += XXH3_mix16B(p+len-64, key+112, seed);
+                    acc += XXH3_mix16B(p+32, key+64, seed);
+                    acc += XXH3_mix16B(p+len-48, key+80, seed);
                 }
 
-                acc += XXH3_mix16B(p+32, key+64, seed);
-                acc += XXH3_mix16B(p+len-48, key+80, seed);
+                acc += XXH3_mix16B(p+16, key+32, seed);
+                acc += XXH3_mix16B(p+len-32, key+48, seed);
             }
 
-            acc += XXH3_mix16B(p+16, key+32, seed);
-            acc += XXH3_mix16B(p+len-32, key+48, seed);
+            acc += XXH3_mix16B(p+0, key+0, seed);
+            acc += XXH3_mix16B(p+len-16, key+16, seed);
         }
-
-        acc += XXH3_mix16B(p+0, key+0, seed);
-        acc += XXH3_mix16B(p+len-16, key+16, seed);
-
         return XXH3_avalanche(acc);
     }
 }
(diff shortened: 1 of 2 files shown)


Cyan4973:
That looks great @easyaspi314 !

Do you have some figures to share ? such as speed or binary size impact ?

easyaspi314:
size of `xxhash.o`, compiled with clang 6.0 -O3 -DXXH_VECTOR=0. 

| arch| -O3, no reroll | -O3, reroll XXH64 | -O3, reroll all | -Oz, no reroll | -Oz, reroll XXH64 | -Oz, reroll | 
|------|-------------:|----------:|---------:|---------:|---------:|------------:|
| x86_64 | 14 KiB | 13 KiB | 12 KiB | 13 KiB | 11 KiB | 10 KiB |
| i386 | 20 KiB | 16 KiB | 16 KiB | 16 KiB | 12 KiB | 10 KiB |
| aarch64 | 12 KiB | 11 KiB | 10 KiB | 12 KiB | 11 KiB | 9 KiB |
| armv4t | 29 KiB | 24 KiB | 20 KiB | 24 KiB | 19 KiB | 16 KiB |
| armv7a | 21 KiB | 17 KiB | 16 KiB | 18 KiB | 14 KiB | 12 KiB |
| thumbv4t | 27 KiB | 23 KiB | 18 KiB | 21 KiB | 16 KiB | 13 KiB |
| thumbv7a | 17 KiB | 15 KiB | 14 KiB | 15 KiB | 12 KiB | 10 KiB |
| mipsel | 76 KiB | 57 KiB | 50 KiB | 68 KiB | 46 KiB | 40 KiB |
| mipseb | 75 KiB | 56 KiB | 50 KiB | 69 KiB | 46 KiB | 41 KiB |
| ppc64le | 19 KiB | 18 KiB | 20 KiB | 17 KiB | 14 KiB | 12 KiB |
| ppc64 | 24 KiB | 21 KiB | 20 KiB | 22 KiB | 19 KiB | 15 KiB |
| powerpc | 26 KiB | 22 KiB | 21 KiB | 21 KiB | 16 KiB | 14 KiB |
| wasm32 | 15 KiB | 13 KiB | 11 KiB | 13 KiB | 10 KiB | 9 KiB |

mips64 doesn't seem to work, and [the wasm64 backend is borked right now](https://bugs.llvm.org/show_bug.cgi?id=41926). 

i386 and ARM both get about a 10% speedup on 16 byte inputs.

Cyan4973:
Thanks @easyaspi314, these are great results !

I believe the only thing missing is a little doc, so that users get informed that the build macro exists and what it does.
This is typically [added here](https://github.com/Cyan4973/xxHash#build-modifiers).

easyaspi314:
OK, will add.

Also, was curious as to why the mips code was so huge. Turns out, Clang is pulling a thumb-gcc: It is turning all the multiplies to shifts and adds. :thinking:

Cyan4973:
That's surprising.
Replacing multiplications by shift + add can work when the number of bits set is not too large,
but the multipliers used in XXH do have a lot of bits set (about half of them),
so I'm surprised the compiler would nonetheless decide that a large sequence of shift+add is still better than a single multiplier.

Maybe it's not subtle enough to really "decide", and just go for shift+add unconditionally.

easyaspi314:
Well apparently, Clang's logic here is f***ed up majorly:

https://github.com/llvm-mirror/llvm/blob/36cbe77a6eee47bfaebd17a713decac4433d6ae8/lib/Target/Mips/MipsSEISelLowering.cpp#L718

It will do shift and add unconditionally if the multiple is the same as the word size, and it doesn't even do the shift and add in the best way possible:

```c
#include <stdint.h>
uint32_t multiply(uint32_t x)
{
    return x * 0xAAAAAAAA; // 0b10101...
}
```

```asm
multiply:
        sll     $1, $4, 1
        sll     $2, $4, 3
        sll     $3, $4, 29
        sll     $5, $4, 27
        sll     $6, $4, 25
        sll     $7, $4, 23
        sll     $8, $4, 21
        sll     $9, $4, 19
        sll     $10, $4, 17
        sll     $11, $4, 15
        sll     $12, $4, 13
        addu    $1, $2, $1
        sll     $2, $4, 5
        addu    $1, $2, $1
        sll     $2, $4, 7
        addu    $1, $2, $1
        sll     $2, $4, 9
        addu    $1, $2, $1
        sll     $2, $4, 31
        sll     $4, $4, 11
        addu    $1, $4, $1
        addu    $1, $12, $1
        addu    $1, $11, $1
        addu    $1, $10, $1
        addu    $1, $9, $1
        addu    $1, $8, $1
        addu    $1, $7, $1
        addu    $1, $6, $1
        addu    $1, $5, $1
        addu    $1, $3, $1
        jr      $ra
        addu    $2, $2, $1
```

Not only is that way too many cycles for shift and add to be beneficial, the shifting and adding *itself* is f***ed up as well. This would be a far more efficient way to multiply:

```asm
multiply:
        sll     $2, $4, 2
        addu    $2, $2, $4
        sll     $4, $2, 4
        addu    $2, $2, $4
        sll     $4, $2, 8
        addu    $2, $2, $4
        sll     $4, $2, 16
        addu    $2, $2, $4
        jr      $ra
        sll     $2, $2, 1
```

Cyan4973:
I think the only thing missing for this path to merge was the doc request.
I can probably add it myself, I'll look into it.

Cyan4973:
I'm merging this PR into its own feature branch, so I can properly control / test / bench all aspects, and ensure I can take care of any future source collisions, as there are several changes in flight.

Just to be sure I understand correctly : 
- This patch does not change the behavior of current code, unless `XXH_REROLL` or `XXH_REROLL_XXH64` are defined.
- `XXH_REROLL` is automatically defined on `-Os` or `-Oz`, tracking macro `__OPTIMIZE_SIZE__`
- `XXH_REROLL_XXH64` is necessarily enabled when `XXH_REROLL` is enabled
- `XXH_REROLL_XXH64` is automatically defined on anything non-x64

The "does not change behavior" is not completely correct when getting into the little details, but I'll get to that in the code comments.

xxh3.h
```diff
@@ -902,32 +907,39 @@ XXH_PUBLIC_API XXH128_hash_t
 XXH3_128bits_withSeed(const void* data, size_t len, XXH64_hash_t seed)
 {
     if (len <= 16) return XXH3_len_0to16_128b(data, len, seed);
+    if (len > 128) return XXH3_hashLong_128b(data, len, seed);
```
Cyan4973: It seems this new order adds a branch for any `len > 16`, thus affecting length in the 16-32 range,
which happen to be my primary target at work.
On benchmark, it seems I can measure a small performance degradation in such range, but I need more time to test, as there is ample opportunity for measurement noise.

Cyan4973:
Merged in feature branch `reroll`

Issue #211, opened 2019-06-06
XXH3 performance issue on ARM & WASM with >128byte inputs


aras-p:
I was playing around with XXH3 from 0.7.0 release, and there's a curious performance drop with data sizes >128 bytes on Arm64 and on WebAssembly, but not on x64.

Here's x64 (Xcode 10.1 Clang, on a MacBookPro), things look great:
<img width="953" alt="xxh-x64" src="https://user-images.githubusercontent.com/348087/59063351-93fd8d80-88b0-11e9-8470-e7bc54a6c585.png">

However on Arm64 (iPhone SE, compiled with Xcode 10.1 Clang) -- notice the **big drop** after 128 bytes input size:
<img width="802" alt="xxh-arm64" src="https://user-images.githubusercontent.com/348087/59063444-c4ddc280-88b0-11e9-900b-5850206b36c4.png">

On WebAssembly (Firefox 67, running in the same MacBookPro), there's a drop after 128 bytes too, where XXH3 becomes much slower than previous xxHash:
<img width="818" alt="xxh-wasm" src="https://user-images.githubusercontent.com/348087/59063646-3584df00-88b1-11e9-940d-b81f00995a43.png">

Cyan4973:
Thanks @aras-p , that's very instructive.

128-bytes is the limit between the "small input" and "large input" modes.
The "large input" mode highly depends on performance of vectorial instructions.
This works better for `SSE2` and `AVX2`, as it was designed for these ones.
@easyaspi314 created a specific variant for `NEON`, which is very helpful.

However, it's difficult to tell if the `NEON` code path is correctly triggered in your test.
It's supposed to auto-sense local conditions, but the build-macro magic may be missing it.
An alternative could be to force the implementation to use a specific code path, using the `XXH_VECTOR` build macro. For `NEON`, the code is `3`, which gives for `gcc` : `-DXXH_VECTOR=3`.

Web Assembly is unlikely to be able to generate vectorial instructions, at least not yet.
So it probably relies on the universal "scalar" code.
This code wasn't properly optimized in `v0.7`, relying too much on the compiler to do optimal transformations.
In the current `dev` branch, it has been rewritten, in a way which is easier to optimize by most compilers, providing generally nice performance improvements (`clang` is even able to auto-vectorize it ...)

The rewritten "scalar" mode is expected to perform as well or better than `XXH32`.

aras-p:
[...]

PR #213, opened 2019-06-10
Detect SSE2 on MSVC 32 bit as well

aras-p:
Detecting SSE2 instructions on Visual Studio when it's compiling for x86 32-bit needs to be done by yet another macro, see https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=vs-2019

The change turns performance graph from the red line into the blue one.

Clipboard01

Somewhat related to things discussed at #211


Cyan4973:
Looks great ! Thanks @aras-p !

PR #214, opened 2019-06-10
Rename ALIGN to XXH_ALIGN

orivej:
ALIGN is a common name that may be defined before xxh3.h is included.


Cyan4973:
Agreed @orivej,
this is actually fixed in a later development branch not yet merged into `dev`,
but there's a case to make this feature available sooner into `dev`, without tying it to more disruptive features.
I'll take care of any potential merge conflict later on.

PR #222, opened 2019-07-03
Fix XXH3 streaming result when update block size happens to land exactly on buffer size/

aras-p:
I was testing the non-streaming XXH3 on 512 byte block, and a streaming API that happened to do a "3 bytes update, then 509 bytes update", and the result was different. Looks like `bufferedSize` was not properly updated if the incoming block size landed exactly on multiple of the internal buffer size. With this proposed fix it seems to be better.

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -1066,7 +1066,7 @@ XXH3_64bits_update(XXH3_state_t* state, const void* input, size_t len)
             } while (p<=limit);
         }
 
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks for notification @aras-p !

I should add a test case with this scenario, to witness it failing with current version, and ensure it remains fixed in the future.

Cyan4973:
bug reproduced and fix verified, thanks @aras-p !

PR #223, opened 2019-07-03
Initial attempt at 128-bit XXH3 streaming variant

aras-p:
- seems to produce same results as non-streaming functions,
- the 128-bit non-streaming ones don't support "custom secret", so
  neither does the streaming variant
- the 64-bit functions seem to do something more clever in order to
  avoid leaking the key/secret, which none of current 128-bit functions
  do
- the naming of the streaming functions is a bit weird now, since
  most of the ones with "64" in the name are what should be used in
  128 bit case too


Cyan4973:
Thanks for this great starting point @aras-p ! 

Don't worry about the naming and variants, we'll take care of them in next patches.

PR #230, opened 2019-07-18
Fixed rerolled version of 64-bit hash. 

gregzaitsev:
Was wrong for length == 4 on 32-bit platforms


Cyan4973:
Nice catch ! Thanks @gregzaitsev !

PR #231, opened 2019-07-22
Update XXH128

Cyan4973:
`XXH128()` is updated to follow the same model as `XXH3_64bits()`.
It can accept a custom secret of any size (> minimum), implements a "mid-size" strategy for 128-240 range, and a streaming implementation is available.

Not completed yet (hence failing tests) :
`ARM` and `VSX` targets must be updated for 128-bit output (`accumulate512` function).

Also to add :
utility function for 128-bit hashes, such as `isEqual` and `cmp`,
and 128-bit canonical representation.

_edit_ : utility functions completed
_edit 2_: 128-bit canonical representation is completed

### DIFF
--- a/tests/bench/hashes.h
+++ b/tests/bench/hashes.h
@@ -96,6 +96,12 @@ size_t xxh3_wrapper(const void* src, size_t srcSize, void* dst, size_t dstCapaci
 }
 
 
+size_t XXH128_wrapper(const void* src, size_t srcSize, void* dst, size_t dstCapacity, void* customPayload)
+{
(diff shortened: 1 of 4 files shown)


Cyan4973:
OK, seems that all remaining points are fixed.
Just restarting all tests, quality and speed, just for verification, but no surprise expected.

Last details are the update to `accumulator512()` for `NEON` and `VSX` code path.

I could disable them just to pass tests, but that would disable them for both 64 and 128-bit, and 64-bit variant is fine.

Maybe time to learn some intrinsics ...

PR #237, opened 2019-07-27
Add CMake Config.

Mezozoysky:
Hello!
Patch allows easy integration within downstream CMake projects and fixes some issues of BUNDLED_MODE.
Unfortunately it requires CMake 3.13 but i dont think that is a realy big deal.
Have a good day!

### DIFF
--- a/cmake_unofficial/CMakeLists.txt
+++ b/cmake_unofficial/CMakeLists.txt
@@ -5,6 +5,8 @@
 #
 # For details, see <http://creativecommons.org/publicdomain/zero/1.0/>.
 
+cmake_minimum_required (VERSION 2.8.12 FATAL_ERROR)
+
 set(XXHASH_DIR "${CMAKE_CURRENT_SOURCE_DIR}/..")
 
 file(STRINGS "${XXHASH_DIR}/xxhash.h" XXHASH_VERSION_MAJOR REGEX "^#define XXH_VERSION_MAJOR +([0-9]+) *$")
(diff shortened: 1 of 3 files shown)


Mezozoysky:
I see i wasnt right about CMake version :)
Should i decrease minimum CMake version?

Cyan4973:
> Should i decrease minimum CMake version?

Yes, unfortunately. Requiring a too recent version has an outsized impact on user-base, as not all systems can depend on up to date versions.

The current minimum version is `2.8.12`. Simplest no-brainer scenario is to keep supporting that.

If it feels preferable to set a higher bar, it's not "forbidden", but we must have a discussion to evaluate pros and cons.

Mezozoysky:
Ok! I'll work on it.

Mezozoysky:
Cases tested with CMake 2.8.12, 3.0.2, 3.13.4:
- build xxhash
- install xxhash
- using xxhash from another CMake project via config
- using xxhash from another CMake project via add_subdirectiry

Note: xxHash itself compiles well with CMake 2.8.12, but it fails to build c++ project which depends on xxHash due to the bug in Cmake 2.8.12 (https://github.com/mdadams/jasper/issues/135).

Mezozoysky:
> Note: xxHash itself compiles well with CMake 2.8.12, but it fails to build c++ project which depends on xxHash due to the bug in Cmake 2.8.12 (mdadams/jasper#135)

This can be handled easily within downstream project.

Also, here is a test downstream project (c++) which can select how to consume xxHash (add_subdirectory or find_package) with option (SHXX_USE_BUNDLED_XXHASH) -- https://github.com/Mezozoysky/SHXX

Cyan4973:
Excellent ! Thanks @Mezozoysky !

Mezozoysky:
@Cyan4973 
Glad to contribute!

PR #238, opened 2019-07-30
Cmake opts

Mezozoysky:
Hello!

## first commit

>For historical reasons in CMake 3.12 and below the `option()` command removes a normal (non-cached) variable
>of the same name when:
>
>- a cache entry of the specified name does not exist at all, or
>- a cache entry of the specified name exists but has not been given a type (e.g. via `-D<name>=ON` on the command line).

I've replaced `option()` commands with `set(... CACHE ...)` for build options (`BUILD_XXHSUM` and `BUILD_ENABLE_INLINE_API`) in order to provide possibility to hardcode build options (within downstream project) in BUNDLED_MODE with normal `set()` (for CMake < 3.13).

Also i've prefixed these options with `XXHASH_` for convenience.

## second commit

Fixes bundled mode determining logic and makes `XXHASH_BUNDLED_MODE` overridable cached var which can be set from downstream projects.

## third commit

It's common practice to create separate `build`/`build-<arhc>`/etc directory within project for CMake files/cache/artifatcs so i've added `build*/` entry into .gitignore.
 Also i've added `.projectile` (it's useful for emacs/projectile users) -- if it is "too much" i can remove it from commit.

Good luck!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,9 @@ xxhsum_inlinedXXH.exe
 *.html
 *.wasm
 *.js
+
+# CMake build directories
+build*/
+
+# project managers artifacts
+.projectile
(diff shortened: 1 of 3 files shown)


Mezozoysky:
## fourth commit
Forgot to update readme with new options names!
:)

Issue #145, opened 2018-08-01
[cmake] avoid reference to `xxhsum` build directory


vp1981:
Hello,
I'm making (unofficial) package for Archlinux (my distribution) and ```makepkg``` (the distro package build tool) warns me that ```xxhsum``` has reference to build directory.

I checked that with
```
$ strings xxhsum | grep BUILD_DIR
```
and found only one occurrence. I tried to build the package with ```--save-temps``` flag for gcc and found the ```BUILD_DIR``` in ```xxhsum.o.s``` file. I don't understand the format of that file but according to surroundings of the string it somehow might be related to ```XXXbenchHash``` function. Unfortunately, I couldn't figure out how that string comes to the ```xxhsum```, I don't see any usage of special macros there.

I would like to get rid of this string (see, for example, "reproducible builds") but now I'm lost and don't have any clue what to do further. Does anyone has idea how to get rid of this string in compiled binary or know how to debug the problem?

Cyan4973:
That's pretty strange.
There's nothing within `xxhsum` source code susceptible to generate a reference to `BUILD_DIR`.

I compiled it on a linux mint vm, and made the same check as you : 
```
$ strings xxhsum | grep BUILD_DIR
$
```

It did not returned anything.

It might be an artefact of the build system itself.

Cyan4973:
Ah, I haven't tested `cmake`, and it might be introducing this issue.
I only tested the simpler `make` build process  : 

```
$ cd XXH_SRC_DIR
$ make
$ strings xxhsum | grep BUILD_DIR
$
```

`cmake` is 3rd-party maintained (hence the "unofficial" tag), I don't understand all its internals and side-effects.
I only directly maintain `Makefile`.

vp1981:
Hello,
for the record, I found the reason and workaround for the issue. In short: use the following [```CMakeLists.txt```](https://gist.github.com/vp1981/a1850615db33e0575fb153d835498b18) file and put it in top directory.

[...]

PR #239, opened 2019-07-30
Add CMake build types for single-configuration generators and set def…

Mezozoysky:
Patch adds CMake build types for single-configuration generators and set default build type to Release.
Build types are: **Debug**, **Release**, **RelWithDebInfo**, **MinSizeRel**.

All the build types except **Debug** define `NDEBUG`, so probably this solves #145.

Have a nice day!

### DIFF
--- a/cmake_unofficial/CMakeLists.txt
+++ b/cmake_unofficial/CMakeLists.txt
@@ -34,6 +34,15 @@ else()
         LANGUAGES C)
 endif()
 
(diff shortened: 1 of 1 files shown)


Cyan4973:
There's apparently a conflict with your previous patch.

Mezozoysky:
I've resolve it soon!

Mezozoysky:
@Cyan4973 Fixed!

PR #244, opened 2019-08-19
[PPC64] Fix it, make it compatible with POWER8, and don't allow POWER7

easyaspi314:
The VSX codepath now works fine, and seems to compile on GCC 4.8.5.

The POWER9 code is now optional, and the only time it is used is for
swapping on big endian.

On big endian, we now use vperm and vpermxor to combine the swap with
the data_key xor.

Wasn't able to test the big endian hardware, but the little endian version works on a physical POWER8E.

### DIFF
--- a/xxhash.c
+++ b/xxhash.c
@@ -297,12 +297,15 @@ XXH_PUBLIC_API unsigned XXH_versionNumber (void) { return XXH_VERSION_NUMBER; }
 /* *******************************************************************
 *  32-bit hash functions
 *********************************************************************/
+#ifdef __cplusplus
+#  define const constexpr
+#endif
 static const U32 PRIME32_1 = 0x9E3779B1U;   /* 0b10011110001101110111100110110001 */
 static const U32 PRIME32_2 = 0x85EBCA77U;   /* 0b10000101111010111100101001110111 */
 static const U32 PRIME32_3 = 0xC2B2AE3DU;   /* 0b11000010101100101010111000111101 */
 static const U32 PRIME32_4 = 0x27D4EB2FU;   /* 0b00100111110101001110101100101111 */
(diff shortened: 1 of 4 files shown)


xxhash.c
```diff
@@ -297,12 +297,15 @@ XXH_PUBLIC_API unsigned XXH_versionNumber (void) { return XXH_VERSION_NUMBER; }
 /* *******************************************************************
 *  32-bit hash functions
 *********************************************************************/
+#ifdef __cplusplus
+#  define const constexpr
```
Cyan4973: This seems to break the g++ compilation test.

What was the expected benefit from this change ?

Cyan4973:
Thanks,
could it be possible to as a `VSX` compilation and runtime test on travis CI ?
It would also validated\ that all sample values are properly generated on target system.

There is already this test for PowerPC : 
https://github.com/Cyan4973/xxHash/blob/dev/.travis.yml#L54

but it doesn't trigger `VSX`, because it used to fail, either at compilation or `qemu` stage.

xxhash.c
easyaspi314: Gah, I was testing something and forgot to change it back.

easyaspi314:
I enabled some tests but it is taking a very long time to apt-get install.

easyaspi314:
Oh, damn it. The real problem is that qemu is too old. :frowning:

I'm gonna try using Xenial.

easyaspi314:
Gah, @Cyan4973 , can you figure out what is going on?

easyaspi314:
I'm going to reopen this with these fiddling squashed once travis is fixed.

Cyan4973:
wow, `bionic` is available on Travis !
That's pretty good news.

Cyan4973:
I can merge that, or you can force-push a squashed version first, as you wish.

easyaspi314:
I suck at git lol

PR #247, opened 2019-08-28
Prevent Clang from emitting unaligned ldm/ldrd on ARMv6, better arm macros

easyaspi314:
Clang was using ldmia and ldrd on unaligned pointers on ARMv6. These instructions don't support unaligned access.

When `-munaligned-access` is used, this will emit normal ldr sequences which do support unaligned access.
I also check the numerical value of __ARM_ARCH.

### DIFF
--- a/xxhash.c
+++ b/xxhash.c
@@ -50,10 +50,10 @@
  * Prefer these methods in priority order (0 > 1 > 2)
  */
 #ifndef XXH_FORCE_MEMORY_ACCESS   /* can be defined externally, on command line for example */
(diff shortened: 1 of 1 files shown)


easyaspi314:
```c
xxhash.c:56:27: error: "__ARM_ARCH" is not defined [-Werror=undef]
   (defined(__GNUC__) && ( __ARM_ARCH >= 7))
```
reeeee

Cyan4973:
Thanks @easyaspi314 !

PR #254, opened 2019-09-16
Better 128-bit multiply, multiple bugfixes.

easyaspi314:
Sorry about the disorganized commit. :(

Yet again, I had to fix ARMv6. Clang went from ldm to ldrd which also bus errors.

Therefore, I decided to fix the root problem and remove the `XXH_FORCE_DIRECT_MEMORY_ACCESS` hack, using only memcpy.

This will kill alignment memes for good, and besides, it didn't seem to make much of a difference.

Additionally, I added my better 128-bit long multiply and applied DRY to `XXH3_mul128_fold64`. This also removes the cryptic inline assembly hack, as UMAAL is properly generated now.

Each method was documented, too (we need more comments).

Also, I added a warning for users who are compiling Thumb-1 code for a target supporting ARM instructions.

While all versions of ARM and Thumb-2 meet XXH3's base requirements, Thumb-1 does not.

First of all, UMULL is inaccessible in the 16-bit subset. This means that every `XXH_mult32to64` means a call to `__aeabi_lmul`.

Since every operation in XXH3 needs to happen in the Lo registers plus having to setup r0-r3 many times for `__aeabi_lmul`, the output resembles a game of Rush Hour:

```
 $ clang -O3 -S --target=arm-none-eabi -march=armv4t -mthumb xxhash.c
 $ grep -c mov xxhash.s
 5472
 $ clang -O3 -S --target=arm-none-eabi -march=armv4t xxhash.c
 $ grep -c mov xxhash.s
 2071
```

It is much more practical to compile xxHash with the wider instruction sets, as these restrictions do not apply.

This doesn't warn if ARMv6-M is targeted; Thumb-1 is unavoidable.

Lastly, I removed the pragma clang loop hack which didn't work anymore since the number of iterations can't be constant evaluated. Now, we don't have 20 warnings when compiling for x86.

### DIFF
--- a/xxhash.c
+++ b/xxhash.c
@@ -50,10 +50,10 @@
  * Prefer these methods in priority order (0 > 1 > 2)
  */
 #ifndef XXH_FORCE_MEMORY_ACCESS   /* can be defined externally, on command line for example */
-#  if defined(__GNUC__) && defined(__ARM_FEATURE_UNALIGNED) && defined(__ARM_ARCH) && (__ARM_ARCH == 6)
+#  if !defined(__clang__) && defined(__GNUC__) && defined(__ARM_FEATURE_UNALIGNED) && defined(__ARM_ARCH) && (__ARM_ARCH == 6)
 #    define XXH_FORCE_MEMORY_ACCESS 2
-#  elif (defined(__INTEL_COMPILER) && !defined(_WIN32)) || \
-  (defined(__GNUC__) && (defined(__ARM_ARCH) && __ARM_ARCH >= 7))
+#  elif !defined(__clang__) && ((defined(__INTEL_COMPILER) && !defined(_WIN32)) || \
+  (defined(__GNUC__) && (defined(__ARM_ARCH) && __ARM_ARCH >= 7)))
 #    define XXH_FORCE_MEMORY_ACCESS 1
 #  endif
 #endif
--- a/xxhash.c
+++ b/xxhash.c
@@ -36,27 +36,6 @@
 /* *************************************
 *  Tuning parameters
 ***************************************/
-/*!XXH_FORCE_MEMORY_ACCESS :
- * By default, access to unaligned memory is controlled by `memcpy()`, which is safe and portable.
- * Unfortunately, on some target/compiler combinations, the generated assembly is sub-optimal.
- * The below switch allow to select different access method for improved performance.
- * Method 0 (default) : use `memcpy()`. Safe and portable.
- * Method 1 : `__packed` statement. It depends on compiler extension (ie, not portable).
- *            This method is safe if your compiler supports it, and *generally* as fast or faster than `memcpy`.
- * Method 2 : direct access. This method doesn't depend on compiler but violate C standard.
- *            It can generate buggy code on targets which do not support unaligned memory accesses.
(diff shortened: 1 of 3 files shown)


Cyan4973:
> Each method was documented, too (we need more comments).

This is excellent. I really like the detailed comments explaining the logic behind each choice. This is a keeper.

> I decided to fix the root problem and remove the XXH_FORCE_DIRECT_MEMORY_ACCESS hack, using only memcpy.

I believe this is one step too far. Let me explain.

Not only does this remove the "direct access" method `2`, which is problematic for the ARMv6 case you just fixed, it also removes the more cautious and effective "packed" method `1`, which has proven very effective in a number of scenarios.

These memory access methods were tested on various compilers, and `clang` do not need them : `memcpy()` is always the better choice. So when tested under `clang`, it's easy to dismiss the other methods as useless.

That's because they were not created for `clang`. `gcc` is more finicky in this regard, and performance differences between versions and methods can be absolutely huge. See :
http://fastcompression.blogspot.com/2015/08/accessing-unaligned-memory.html

`gcc` is not the only compiler with such issue, I could measure `icc` back then, and it wasn't great either when using `memcpy()`. I could not test `visual` back then, but I wouldn't bet too fast that `memcpy()` will be good enough for this compiler, nor for all versions of this compiler.

So, in short, these different access methods may prove useful for other compilers than `clang`, including some we don't know about, and that could select one method or another thanks to this build constant (which default to `0` == `memcpy()` anyway, except for a set of pre-defined configuration).

What is possible and welcomed is to update the set of pre-defined configurations that use non-default access methods. In this regard, it seems useful to ensure that methods `1` and `2` are disabled for `clang`. The current set of macros did not do that, because it's based on `__GNUC__` detection, which unfortunately is also triggered by `clang` ! It's an overlook that can be solved by additionally testing the _absence_ of `__clang__`.

xxhash.c
```diff
@@ -36,6 +36,27 @@
...
+ * Method 1 : `__packed` statement. It depends on compiler extension (ie, not portable).
+ *            This method is safe if your compiler supports it, and *generally* as fast or faster than `memcpy`.
+ * Method 2 : direct access. This method doesn't depend on compiler but violate C standard.
+ *            It can generate buggy code on targets which do not support unaligned memory accesses.
+ *            But in some circumstances, it's the only known way to get the most performance (ie GCC + ARMv6)
+ * See http://stackoverflow.com/a/32095106/646947 for details.
+ * Prefer these methods in priority order (0 > 1 > 2)
+ */
+#ifndef XXH_FORCE_MEMORY_ACCESS   /* can be defined externally, on command line for example */
+#  if defined(__GNUC__) && defined(__ARM_FEATURE_UNALIGNED) && defined(__ARM_ARCH) && (__ARM_ARCH == 6)
```
Cyan4973: I believe you want to add something like `!defined(__clang__)` to ensure that `XXH_FORCE_MEMORY_ACCESS` will not be automatically set to `1` or `2`.

easyaspi314:
Oh wow, I need coffee. I fixed it in the vanilla copy I was using to revert. 

Thanks for checking that lol

Cyan4973:
Thanks @easyaspi314 !

PR #266, opened 2019-10-02
Try to improve some variable names.

easyaspi314:
It's a start, but an improvement. I still have more things I would like
to change but it is good for now.

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -201,7 +201,7 @@ XXH_FORCE_INLINE U64x2 XXH_vec_revb(U64x2 val)
  *     return ret;
  * }
  *
- * Because both of the main loops load the key, swap, and xor it with data,
+ * Because both of the main loops load the key, swap, and xor it with input,
  * we can combine the key swap into this instruction.
  */
 #  ifdef vec_permxor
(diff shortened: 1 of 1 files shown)


easyaspi314:
Coding without coffee again
:man_facepalming:

easyaspi314:
@Cyan4973 do you agree with these names? I am still trying to decide on some of the shorter names, although personally, I don't think we should use SSA.

xxh3.h
```diff
@@ -474,53 +474,53 @@ typedef enum { XXH3_acc_64bits, XXH3_acc_128bits } XXH3_accWidth_e;
...
+        const         __m256i* const xsecret = (const __m256i *) secret;   /* not really aligned, just for ptr arithmetic, and because _mm256_loadu_si256() requires this type */
 
         size_t i;
         for (i=0; i < STRIPE_LEN/sizeof(__m256i); i++) {
-            __m256i const d   = _mm256_loadu_si256 (xdata+i);
-            __m256i const k   = _mm256_loadu_si256 (xkey+i);
-            __m256i const dk  = _mm256_xor_si256 (d,k);                                  /* uint32 dk[8]  = {d0+k0, d1+k1, d2+k2, d3+k3, ...} */
-            __m256i const mul = _mm256_mul_epu32 (dk, _mm256_shuffle_epi32 (dk, 0x31));  /* uint64 mul[4] = {dk0*dk1, dk2*dk3, ...} */
+            __m256i const data_vec = _mm256_loadu_si256 (xinput+i);
+            __m256i const key_vec = _mm256_loadu_si256 (xsecret+i);
```
Cyan4973: minor : for consistency, it might preferable to use `secret` in this name too ?

Cyan4973:
I share the objective to converge towards a more uniform naming scheme, using `secret` and `input`.

xxh3.h
easyaspi314: Yeah, but I'm still thinking of something that isn't going to be 40 chars long, as once you combine them it gets pretty messy.

easyaspi314:
Also, dunno what is up with Travis — it already passed hours ago

PR #267, opened 2019-10-03
[SCALAR] Improve scalar XXH3_accumulate_512 loop

easyaspi314:
The previous XXH3_accumulate_512 loop didn't fare well since XXH128
started swapping the addition.

Neither GCC nor Clang could follow the barely-readable loop, resulting
in garbage code output.

This made XXH3 much slower. Take 32-bit scalar ARM.

Ignoring loads and potential interleaving optimizations, in the main
loop, XXH32 takes 16 cycles for 8 bytes on a typical ARMv6+ CPU, or 2 cpb.

```asm
        mla     r0, r2, r5, r0  @ 4 cycles
	ror     r0, r0, #19     @ 1 cycle
	mul     r0, r0, r6      @ 3 cycles
	mla     r1, r3, r5, r1  @ 4 cycles
	ror     r1, r1, #19     @ 1 cycle
	mul     r1, r1, r6      @ 3 cycles
```

XXH3_64b takes 9, or 1.1 cpb:
```asm
        adds    r0, r0, r2      @ 2 cycles
	adc     r1, r1, r3      @ 1 cycle
	eor     r4, r4, r2      @ 1 cycle
	eor     r5, r5, r3      @ 1 cycle
	umlal   r0, r1, r4, r5  @ 4 cycles
```

Benchmarking on a Pixel 2 XL (with a binary for ARMv4T), previously,
XXH32 got 1.8 GB/s, while XXH3_64b got 1.7.

Now, XXH3_64b gets 2.3 GB/s! This calculates out well (as additional
loads and stores have some overhead).

Unlike before, it is better to disable autovectorization completely, as
the compiler can't vectorize it as well. (Especially with Clang and
NEON, where it extracts to multiply instead of the obvious vmlal.u32!).

On that same device in aarch64 mode XXH3's scalar version when compiled
with `clang-8 -O3 -DXXH_VECTOR=0 -fno-vectorize -fno-slp-vectorize`,
XXH3 went from 2.3 GB/s to 4.3 GB/s. For comparison, the NEON version
gets 6.0 GB/s.

However, almost all platforms with decent autovectorization have a
handwritten intrinsics version which is much faster.

For optimal performance, use -fno-tree-vectorize -fno-tree-slp-vectorize
(or simply disable SIMD instructions entirely).

From testing, ARM32 also prefers forced inlining, so I enabled it.

I also fixed some typos.

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -650,22 +650,16 @@ XXH3_accumulate_512(      void* XXH_RESTRICT acc,
     const BYTE* const xsecret  = (const BYTE*) secret;   /* no alignment restriction */
     size_t i;
     XXH_ASSERT(((size_t)acc & (XXH_ACC_ALIGN-1)) == 0);
-    for (i=0; i < ACC_NB; i+=2) {
-        U64 const in1 = XXH_readLE64(xinput + 8*i);
-        U64 const in2 = XXH_readLE64(xinput + 8*(i+1));
-        U64 const key1  = XXH_readLE64(xsecret + 8*i);
-        U64 const key2  = XXH_readLE64(xsecret + 8*(i+1));
-        U64 const data_key1 = key1 ^ in1;
-        U64 const data_key2 = key2 ^ in2;
-        xacc[i]   += XXH_mult32to64(data_key1 & 0xFFFFFFFF, data_key1 >> 32);
-        xacc[i+1] += XXH_mult32to64(data_key2 & 0xFFFFFFFF, data_key2 >> 32);
-        if (accWidth == XXH3_acc_128bits) {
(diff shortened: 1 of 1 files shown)


easyaspi314:
> On that same device in aarch64 mode XXH3's scalar version when compiled
> with clang-8 -O3 -DXXH_VECTOR=0 -fno-vectorize -fno-slp-vectorize,
> XXH3 went from 2.3 GB/s to 4.3 GB/s. For comparison, the NEON version
gets 6.0 GB/s.

Nice grammar, Devin

Cyan4973:
Yes, that version of the scalar code path is much better

PR #268, opened 2019-10-04
Unify AVX/SSE2 and NEON code

easyaspi314:
SSE2 and AVX2 now use macros which are defined to either SSE2 intrinsics
or AVX2 to unify the implementation.

They keep the same name pattern as the original intrinsics, avoiding
confusion:

```c
    __m128i vec = _mm_loadu_si128(ptr);
    __m128i sum = _mm_add_epi64(vec, acc);
```
becomes
```c
    XXH_vec_t vec = XXH_loadu_si(ptr);
    XXH_vec_t sum = XXH_add_epi64(vec, acc);
```

NEON now has a macro abstracting the vzip hack and the generic
vshrn/vmovn version, and I also try to document it better without
getting too lost in explanation.

Now, there are 3 less implementations to worry about.

I also tried to sync up some inconsistencies between SSE2 and NEON's
layout. I can see if reordering is better for SSE2 later (I am doing
everything on my phone right now).

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -155,9 +155,114 @@
 #    define XXH_mult32to64(x, y) ((U64)((x) & 0xFFFFFFFF) * (U64)((y) & 0xFFFFFFFF))
 #endif
 
+/*
+ * AVX2 and SSE2 use the exact same algorithm, just with different function names.
+ *
+ * We use these macros/typedefs to reduce duplication.
+ */
+#if XXH_VECTOR == XXH_AVX2
+   typedef __m256i XXH_vec_t;
+#  define XXH_loadu_si      _mm256_loadu_si256
+#  define XXH_set1_epi32    _mm256_set1_epi32
+#  define XXH_xor_si        _mm256_xor_si256
+#  define XXH_add_epi64     _mm256_add_epi64
+#  define XXH_shuffle_epi32 _mm256_shuffle_epi32
+#  define XXH_mul_epu32     _mm256_mul_epu32
+#  define XXH_slli_epi64    _mm256_slli_epi64
+#  define XXH_srli_epi64    _mm256_srli_epi64
+#elif XXH_VECTOR == XXH_SSE2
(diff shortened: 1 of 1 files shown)


Cyan4973:
I'm a bit more puzzled about this one.

If I do understand correctly, the high level strategy of this PR 
is to move from this design : 

```C
int function()
{
#if ARCH1
intrinsic_arch1()
...
#elif ARCH2
intrinsic_arch2()
...
#elif ARCH3
intrinsic_arch3()
...
#endif  
}
```

onto this new design : 

```C
#if ARCH1
#define multiarch_intrinsic() intrinsic_arch1()
...
#elif ARCH2
#define multiarch_intrinsic() intrinsic_arch2()
...
#endif  

int function()
{
#if ARCH1 || ARCH2
multiarch_intrinsic()
...
#elif ARCH3
intrinsic_arch3()
...
#endif  
}

```

I'm not sure if it really helps complexity.
It seems to mostly transfer it, from the function, to the macro definition.

Besides, it's a partial strategy, which is mostly suitable for `SSE2` / `AVX2` transition, because it forces the code to be structured in a way which fits all target architectures, with each step defined as a `multiarch_intrinsic()`, requiring to find an exact equivalent for all target intrinsics. This is not the optimal choice for other architectures, which still have to be defined with their own code path.

Cyan4973:
For `XXH_SPLIT_IN_PLACE()`, which is included in this PR, I would be more favorable,
because it puts a name and a good explanation on an operation which is tricky to follow, helping code readability.

Cyan4973:
I would like to keep and merge the `NEON` part of this patch.

easyaspi314:
Ok, will do

Cyan4973:
Hi @easyaspi314 , 
what do we do about this PR ? Should we close it ?
I think there was a part of it, around `NEON`, which felt interesting

easyaspi314:
Closed in favor of #300

PR #269, opened 2019-10-04
Fix endianness detection on GCC, avoid XXH_cpuIsLittleEndian.

easyaspi314:


Cyan4973:
FYI, be aware that I had nasty surprises in the past with endian detection by macros.

It kind of work, generally, and is good enough for most "mainstream" systems.
But for less common ones, things quickly start to go awry.

Long story short : a compiler-controlled (as opposed to preprocessor-controlled) endian detection solved all these problems, and proved resilient with any system tested ever since (not a single issue filed on this topic).

Re-introducing pre-processor controlled endian detection may open the door to these past problems. I see that you kept the function `XXH_isLittleEndian()` for cases that could not be correctly detected with preprocessor. Hopefully, that will be enough. Time will tell.

PR #271, opened 2019-10-07
Improve typedefs, fix 16-bit int/seed type bug

### LINKED ISSUE #258: BYTE is predefined by some compiler
One of the targets with which we wish to use xxHash requires compilation with a very old version of MSVC++ (Visual Studio 6), which defines a type named BYTE. This conflicts with xxHash. I've gotten around it for now by renaming xxHash's BYTE to XXBYTE. Perhaps this can be addressed officially in the sources?

Also, _byteswap_ulong is not supported by that version of MSVC++ either.

easyaspi314:
Fixes #258.

```c
BYTE -> xxh_u8
U32  -> xxh_u32
U64  -> xxh_u64
```

Additionally, I hopefully fixed an issue for targets where int is 16
bits. XXH32 used unsigned int for its seed, and in C90 mode, unsigned
int as its U32. This would cause truncation issues. I check limits.h in
C90 mode to make sure UINT_MAX == 0xFFFFFFFFUL, and if it isn't, use
unsigned long.

We should see if we can set up an AVR CI test. Just to run the
verification program, though, as the benchmark will take a very long
time.

Lastly, the seed types are XXH32_hash_t and XXH64_hash_t for XXH32/64.
This matches xxhash.c and prevents the aforementioned 16-bit int bug.

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -148,12 +148,12 @@
 #  endif
 #endif
 
-/* U64 XXH_mult32to64(U32 a, U64 b) { return (U64)a * (U64)b; } */
+/* xxh_u64 XXH_mult32to64(xxh_u32 a, xxh_u64 b) { return (xxh_u64)a * (xxh_u64)b; } */
 #if defined(_MSC_VER) && defined(_M_IX86)
 #    include <intrin.h>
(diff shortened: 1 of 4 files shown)


Cyan4973:
Thanks @easyaspi314 !

PR #278, opened 2019-11-04
transferred implementation inside xxhash.h

Cyan4973:
instead of `xxhash.c` .

This seems preferable for some build systems,
which don't like the `#include "xxhash.c"` statement
when inlining xxhash, as reported by @pdillinger .

Note that `xxhash.c` still exists,
it just includes the implementation and instantiates it.


easyaspi314:
Not sure how I feel about this one.

Cyan4973:
Merging.
This is a big change, pushing all the code into `xxhash.h`, for easier inlining experience.

PR #283, opened 2019-12-14
Use Travis CI Arm64 and PPC64le arch

adamretter:
This offloads the Arm64 and PPC64le builds and tests to real hardware that is now available on Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,16 +6,19 @@ matrix:
 
     - name: General linux tests (Xenial)
       dist: xenial
-      before_install:
-        - sudo apt-get update  -qq
-        - sudo apt-get install -qq clang
-        - sudo apt-get install -qq g++-multilib
-        - sudo apt-get install -qq gcc-multilib
-        - sudo apt-get install -qq cppcheck
+      arch: amd64
+      addons:
+        apt:
+          packages:
+            - clang
+            - g++-multilib
+            - gcc-multilib
+            - cppcheck
       script:
         - make -B test-all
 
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks @adamretter ! That's excellent suggestion !

I wasn't even aware that `ppc64le` was a thing at TravisCI !

adamretter:
@Cyan4973 well I have two reasons to contribute:
1. we have a slightly modified xxHash that we use in RocksDB, but I am having trouble building it under ppc64le on Travis CI. So I wanted to see if it is the same upstream with you.
2. I think the pcc64le, arm64, and s390x support by Travis is pretty cool, and it would be nice for your project too.

adamretter:
@Cyan4973 well it seems to work just fine here :-)

If you support IBM Z (s390x), you might also be interested to add that to Travis - https://blog.travis-ci.com/2019-11-12-multi-cpu-architecture-ibm-power-ibm-z

Cyan4973:
> If you support IBM Z (s390x), you might also be interested to add that to Travis 

Good point ! Worth a try !

easyaspi314:
Hmm, s390x has SIMD too…

Seems like its intrinsics are mostly compatible with VSX, too. Just need to find some docs though, I don't know anything about this architecture aside from the Wikipedia article :joy:

Cyan4973:
s390x seems to work fine too : https://travis-ci.org/Cyan4973/xxHash/jobs/625109243

One more target added to CI :)

PR #289, opened 2020-01-02
fixed a bunch of minor cppcheck warnings

Cyan4973:
nothing terribly important, just some clean up.
Also : a few `cppcheck` warnings are plain false positives with no obvious replacement, in which case, no code modification is performed.


Cyan4973:
once again, an issue with `ppc64le` tests,
but travisCI states that it's now fixed, so let's merge that.

PR #297, opened 2020-02-07
gcc10 altivec

ellert:
When gcc was updated to version 10 in Fedora the build of the xxhash
package started failing. The compilation succeeds, but the checks
trigger an error. This change, suggested by the maintainers of the gcc
package in Fedora, addresses the issue. F0r details see:

https://bugzilla.redhat.com/show_bug.cgi?id=1798908


easyaspi314:
Why do `vmuleuw` and `vmulouw` have to keep switching… I want to *avoid* inline assembly if possible because it breaks patterns...

Cyan4973:
Simple fix. Thanks @ellert !

Indeed, it's a pity for `gcc` speed on target architecture, but I guess this could be addressed in a different PR.

PR #300, opened 2020-02-11
Unify NEON code, fully document the VZIP.32 hack.

easyaspi314:
AArch64 and ARMv7 now use the same codepath, with the VZIP.32 hack being
abstracted away in a macro.

I also fully (over?) documented it to explain the hack (as well as make a note of the use of `vmull_u32` instead of `vmul_u32` in `XXH3_scrambleAcc` to work around a bug in Clang 6 and 7).

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -156,9 +156,109 @@
 #    define XXH_mult32to64(x, y) ((xxh_u64)((x) & 0xFFFFFFFF) * (xxh_u64)((y) & 0xFFFFFFFF))
 #endif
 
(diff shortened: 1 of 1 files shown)


Cyan4973:
> I also fully (over?) documented it

I'm sure you will have a lot of interested readers for this section, it's very welcomed

Quick question :
I presume that this code path is verified with this test :
https://github.com/Cyan4973/xxHash/blob/dev/.travis.yml#L47 ?

easyaspi314:
Yes. I also verified it on my phone in both 32-bit and 64-bit.

PR #301, opened 2020-02-12
S390x

Cyan4973:
@easyaspi314 : I tried to merge the `s390x` into `dev`, fixing all the conflicts along the way.
Please have a look into the final outcome (`xxh3.h` mostly),
to ensure it preserves the expected behavior.


xxh3.h
```diff
@@ -290,53 +305,54 @@ XXH_FORCE_INLINE U64x2 XXH_vec_revb(U64x2 val)
...
+# if defined(__clang__) && __has_builtin(__builtin_altivec_vmuleuw)
+/* Clang has a better way to control this, we can just use the builtin which doesn't swap. */
 #  define XXH_vec_mulo __builtin_altivec_vmulouw
 #  define XXH_vec_mule __builtin_altivec_vmuleuw
-#else
+# elif defined(__s390x__) && defined(__GNUC__) && (__GNUC__ >= 8)
+/* s390x is always big endian, there is no endian issue. */
+#  define XXH_vec_mulo vec_mulo
+#  define XXH_vec_mule vec_mule
+# else
```
Cyan4973: Considering [this test result](https://travis-ci.org/Cyan4973/xxHash/jobs/649629973#L210), I presume a mistake here.

cc @easyaspi314 .

Cyan4973:
OK, so here is, I presume, the issue :
In the original s390x code, intrinsics `vec_mulo` and `vec_mule` are always used, even though a comment states that "GCC lacks this before version 8.0". So I added a `GCC` version test.

But maybe what was actually meant is that "GCC lacks this before version 8.0 **for powerPC**" ?

Copy/pasting the original code : 

```
/*
 * vec_mulo and vec_mule are very problematic intrinsics:
 *
 *  1. Compilers like to swap them at will.
 *  2. They are endian dependent
 *  2. GCC lacks this before version 8.0
 */
#if defined(__s390x__)
/* s390x is always big endian, this is not an issue. */
#  define XXH_vec_mulo vec_mulo
#  define XXH_vec_mule vec_mule
#elif __has_builtin(__builtin_altivec_vmuleuw)
(...)
#else
(... use inline assembly vmulouw & vmuleuw...)
```

compared with new code  : 

```
/*
 * vec_mulo and vec_mule are very problematic intrinsics:
 *
 * The intrinsic weren't added until GCC 8, despite existing for a while.
 * Besides, they are endian dependent, and their meaning swap depending on version.
 * */
# if defined(__clang__) && __has_builtin(__builtin_altivec_vmuleuw)
(...)
# elif defined(__s390x__) && defined(__GNUC__) && (__GNUC__ >= 8)
/* s390x is always big endian, there is no endian issue. */
#  define XXH_vec_mulo vec_mulo
#  define XXH_vec_mule vec_mule
# else
/* gcc needs inline assembly */
(... use inline assembly vmulouw & vmuleuw...)
```

easyaspi314:
Let me take a look at it and do more testing.

easyaspi314:
```c
#include <altivec.h>

vector unsigned long long vec_mule_(vector unsigned a, vector unsigned b)
{
    return __builtin_altivec_vmuleuw(a, b);
}
```

`clang-9 --target=powerpc64le-linux-gnu`:
```asm
vec_mule_:
    vmuleuw 2, 2, 3
    blr
```
`powerpc64le-linux-gnu-gcc-8`:
```asm
vec_mule_:
    vmulouw 2,2,3
    blr
```
## Excellent.

Also, QEMU doesn't support zvector... 😦

xxh3.h
easyaspi314: `vec_mulo` and `vec_mule` seem to be available with gcc-5 and later with `-march=z13 -mzvector`.

Cyan4973: Changed the policy, so that, on `s390x`, `vec_mulo` and `vec_mule` intrinsics are always used, as in the original patch. There is no `gcc` version check.

This change fixed the tests.

We may want to add more tests if we want to catch issues not detected by current test suite.

PR #302, opened 2020-02-12
xxhash can be inlined even when previously included

Cyan4973:
Inlining hash functions is generally beneficial for performance. The benefit becomes critically important whenever input size can be defined as a compile-time constant.

To inline xxhash functions, one just needs to include it this way :

```
   #define XXH_INLINE_ALL
   #include "xxhash.h"
```

One potential issue is that `xxhash.h` may have already been included previously,
typically as part of another included `*.h` .
In which case, the second `#include` statement will have no effect : the content of `xxhash.h`, having already been parsed, would be skipped.

This patch fixes this situation :
now, when `XXH_INLINE_ALL` is defined, all identifiers are renamed,
in order to avoid naming collisions and confusion with already included identifiers.

The renaming process uses `XXH_NAMESPACE`.
For this to work correctly, `XXH_NAMESPACE` must be available (i.e. not already used).

A test has been added in `tests/` to ensure this scenario works correctly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -282,9 +282,13 @@ preview-man: man
 test: DEBUGFLAGS += -DDEBUGLEVEL=1
 test: all namespaceTest check test-xxhsum-c c90test test-tools
 
+.PHONY: test-inline
+test-inline:
+	$(MAKE) -C tests test
+
 .PHONY: test-all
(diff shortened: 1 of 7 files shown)


easyaspi314:
Nice find.

Issue #293, opened 2020-01-22
Does not accept unicode file path


KonstantinMogilev:
Does not accept unicode file path, for example:

```
xxhsum.exe "ÐµÐÌ.pdf"
Error: Could not open '?╡?I.pdf': Invalid argument.
```

Cyan4973:
Unfortunately, the code handling file names does nothing special, and relies entirely on standard C `fopen()`.

I have no idea how `fopen()` behaves with unicode characters.
This may be system dependent.

I presume the issue happens on Windows ?

KonstantinMogilev:
Yes, the problem is in Windows, checked on 2 systems:
Windows Server 2008 R2
Windows 7
Maybe fopen () has options for working correctly with unicode, or use another function that can correctly handle such names?

Cyan4973:
How was the binary produced ?
Did you compiled it using Visual Studio ?

KonstantinMogilev:
I used your latest compiled version from here: https://github.com/Cyan4973/xxHash/releases/download/v0.7.2/xxhsum_win64_0_7_2_avx2.zip

easyaspi314:
Oof, this will be fun.

`fopen` on Windows uses ANSI encoding, while files are stored in UTF-16.

I might take a look into this when I get a chance. It might be as simple as `_wfopen` with a `mbtowc`, but maybe not.

t-mat:
Hi, I've made [small code](https://github.com/t-mat/xxHash/commit/56914231a8c12b04cd7ee241cce95bd43ef889b3) which introduces UTF-8 compatibility for Windows+MSVC build.
@easyaspi314 since I don't test this code much, I don't send PR so far.  But hope it helps.

t-mat:
Here's some notes.

(1) We can safely use UTF-8 to represent Unicode filename for modern systems (except Windows).

> I have no idea how fopen() behaves with unicode characters. This may be system dependent.

Nowadays as for modern Linux, *NIX and OS X systems, basically we can say all possible filenames are sequence of `char`s.  So in these systems, `fopen()` is fine for Unicode filenames.
Some OS X file system/API has [Unicode normalization issues](https://en.wikipedia.org/wiki/Filename#Unicode_interoperability) and it may cause issue of interoperability of `xxhsum -c`.  But I think we can address this problem later.

(2) I believe "[UTF-8 Everywhere](http://utf8everywhere.org/)" is good approach.
[...]

PR #304, opened 2020-02-13
Fix Unicode support on Windows, minor Windows tweaks

easyaspi314:
- Unicode filenames should now work, with a method that works with and without Unicode mode on Windows.
   - Added a test in the Makefile.
 - Use unbuffered stderr output on Windows, fixes output not updating immediately on MinGW.
 - Fix some missing `$(EXT)`s in the Makefile, causing Clang to emit `xxhsum` instead of `xxhsum.exe` on Windows, as well as xxhsum's rule ignoring `$(FLAGS)`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -169,15 +169,9 @@ check: xxhsum   ## basic tests for xxhsum CLI, set RUN_ENV for emulated environm
 	# request incorrect variant
 	$(RUN_ENV) ./xxhsum$(EXT) -H9 xxhash.c ; test $$? -eq 1
 
-# Make sure that Unicode works.
-# https://github.com/Cyan4973/xxHash/issues/293
-# Japanese: echo "This filename is Unicode." > "Unicode.txt"
 .PHONY: test-unicode
-test-unicode: xxhsum check
-	# Test Unicode filenames.
-	echo "このファイル名はユニコードです。" > "ユニコード.txt"
-	$(RUN_ENV) ./xxhsum$(EXT) "ユニコード.txt"
-	@$(RM) "ユニコード.txt"
+test-unicode:
+	$(MAKE) -C tests test_unicode
 
 .PHONY: test-mem
 VALGRIND = valgrind --leak-check=yes --error-exitcode=1
@@ -295,7 +289,7 @@ test: all namespaceTest check test-xxhsum-c c90test test-tools
 
 .PHONY: test-inline
 test-inline:
-	$(MAKE) -C tests test
+	$(MAKE) -C tests test_multiInclude
 
 .PHONY: test-all
 test-all: CFLAGS += -Werror
--- a/Makefile
+++ b/Makefile
@@ -76,9 +76,10 @@ default: lib xxhsum_and_links
 all: lib xxhsum xxhsum_inlinedXXH
 
 xxhsum: xxhash.o xxhsum.o  ## generate command line interface (CLI)
+	$(CC) $(FLAGS) $^ $(LDFLAGS) -o $@$(EXT)
 
-xxhsum32: CFLAGS += -m32  ## generate CLI in 32-bits mode
-xxhsum32: xxhash.c xxhsum.c  ## do not generate object (avoid mixing different ABI)
+xxhsum32$(EXT): CFLAGS += -m32  ## generate CLI in 32-bits mode
+xxhsum32$(EXT): xxhash.c xxhsum.c  ## do not generate object (avoid mixing different ABI)
 	$(CC) $(FLAGS) $^ $(LDFLAGS) -o $@$(EXT)
 
 xxhash.o: xxhash.h xxh3.h
@@ -89,7 +90,7 @@ xxhsum.o: xxhash.h
 xxhsum_and_links: xxhsum xxh32sum xxh64sum xxh128sum
 
 xxh32sum xxh64sum xxh128sum: xxhsum
-	ln -sf $^ $@
+	ln -sf $<$(EXT) $@$(EXT)
 
 xxhsum_inlinedXXH: CPPFLAGS += -DXXH_INLINE_ALL
 xxhsum_inlinedXXH: xxhsum.c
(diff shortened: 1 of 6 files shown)


easyaspi314:
This should fix #293.

Makefile
```diff
@@ -76,6 +76,7 @@ default: lib xxhsum_and_links
 all: lib xxhsum xxhsum_inlinedXXH
 
 xxhsum: xxhash.o xxhsum.o  ## generate command line interface (CLI)
+	$(CC) $(FLAGS) $^ $(LDFLAGS) -o $@$(EXT)
```
Cyan4973: minor : isn't this rule implied ?

Makefile
```diff
@@ -154,20 +155,29 @@ clean:  ## remove all build artifacts
...
+	$(RUN_ENV) ./xxhsum$(EXT) -H2 xxhash.c
 	# request incorrect variant
-	$(RUN_ENV) ./xxhsum -H9 xxhash.c ; test $$? -eq 1
-
+	$(RUN_ENV) ./xxhsum$(EXT) -H9 xxhash.c ; test $$? -eq 1
+
+# Make sure that Unicode works.
+# https://github.com/Cyan4973/xxHash/issues/293
+# Japanese: echo "This filename is Unicode." > "Unicode.txt"
+.PHOHY: test-unicode
```
Cyan4973: nit : `.PHONY`

Makefile
easyaspi314: Not the FLAGS/CFLAGS part. Make defines it as the following:
```makefile
    $(CC) $(LDFLAGS) n.o $(LOADLIBES) $(LDLIBS)
```

Makefile
easyaspi314: oof

Makefile
```diff
@@ -281,7 +291,7 @@ preview-man: man
 
 .PHONY: test
 test: DEBUGFLAGS += -DDEBUGLEVEL=1
-test: all namespaceTest check test-xxhsum-c c90test test-tools
+test: all namespaceTest check test-xxhsum-c c90test test-tools test-unicode
```
Cyan4973: How will the test itself behave on systems which do not support unicode (if they exist) ?

If there are risks, and this test should rather be reserved to a few "good enough" environments, such as Windows, and I assume Linux and OS-X, then it should rather belong to `test-all` category.
`make test` is reserved for tests which are supposed to be "reasonably" universal

easyaspi314: Fair point. I'll move it.

Makefile
easyaspi314: So, for example, if you did this:
```
make CFLAGS="-m32 -O3"
```
GCC would fail at the link step due to expecting 64-bit object files

easyaspi314:
Hmmmm....

easyaspi314:
```c
Warning: corrupt .drectve at end of def file
xxhsum.o:xxhsum.c:(.text[fprintf]+0x38): undefined reference to `__stdio_common_vfprintf'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [Makefile:79: xxhsum] Error 1
```
Um.....

easyaspi314:
Got classic MinGW to work by changing `_wfopen_s` to `_wfopen`

s u c h   u n i c o d e

`ｅｖｅｒｙｔｈｉｎｇ      ｉｓ      ｆｉｎｅ`

(It is actually an encoding issue on the terminal's side, it defaults to OEM encoding instead of UTF-8. The hash value is correct).

easyaspi314:
Classic MinGW doesn't even like Unicode text at all:

```
$ /c/Windows/System32/chcp.com 65001
Active code page: 65001

$ cat Makefile
....
# Make sure that Unicode works.
# https://github.com/Cyan4973/xxHash/issues/293
# Japanese: echo "This filename is Unicode." > "Unicode.txt"
.PHONY: test-unicode
test-unicode: xxhsum check
        # Test Unicode filenames.
        cat: write error: Permission denied
```

However, there is no way to tell the old MinGW from the new MinGW-w32 AFAIK, and the exe works in the correct console mode.

Will investigate more tomorrow.

easyaspi314:
Maybe we should move the Unicode to a separate file. Or generate it on the fly, like so:
```c
#include <stdio.h>

/* "Unicode.txt" in Japanese */
#define FILENAME "\u30e6\u30fc\u306b\u30b3\u30fc\u30c9.txt"
#define WSTR(x) L##x
int main(void)
{
    FILE *f, *script;
#ifdef _WIN32
    if (!(f = _wfopen(WSTR(FILENAME), L"wb"))) return 1;
    if (!(script = fopen("unicode_test.bat", "w"))) {
        fclose(f);
        return 1;
    }
    fprintf(script, "@ECHO OFF\N");
    fprintf(script, "\"..\\xxhsum.exe\" \"%s\"\n", FILENAME);
    fprintf(script, "EXIT /B %%ERRORLEVEL%%\n");
#else
    if (!(f = fopen(FILENAME, "wb"))) return 1;
    if (!(script = fopen("unicode_test.sh", "w"))) {
        fclose(f);
        return 1;
    }
    fprintf(script, "#!/bin/sh\n");
    fprintf(script, "exec \"../xxhsum\" \"%s\"\n", FILENAME);
#endif
    fprintf(f, "test\n");

    fclose(f);
    fclose(script);
}
```

Just theory-crafting here, it is untested.

Cyan4973:
We could probably get the same effect without a need for a separate file, just with some `#ifdef` macro magic.

easyaspi314:
Actually this is too complex. I'm going to just check `$LANG` for `UTF-8` and skip the test if it isn't found.

That way you can see the output.

easyaspi314:
Oh, I get it why Clang is failing. It has to do with the new Makefile rules.

The `MOREFLAGS=--target=x86_64-w64-mingw32` flag was being applied to the .exe but not to the object files, and Clang was trying to link MSVC object files as MinGW object files.

easyaspi314:
@Cyan4973 thoughts?

Cyan4973:
Great work @easyaspi314 !

PR #305, opened 2020-02-14
Modifications for small inputs (xxh3)

Cyan4973:
The modifications pass the new Perlin noise test (not yet published).
It also helps speed for the 4-8 range (latency test from 112MH/s to 125MH/s).

I think the speed could be improved even more, but to do this, it would be necessary to drop the "64-bit bijective" property (0-collisions when all inputs are 64-bit). I haven't ruled that yet, but if bijection is not important it could be the next move.

I will have to port the modifications to `xxh128` too.

The final avalanche has been slightly altered, improving bit distribution bias across the range.

This however changes all values produced by `xxh3`. Some verification values in `xxhsum` still need updating.

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -58,6 +58,16 @@
 #  define XXH_RESTRICT   /* disable */
 #endif
 
+#if (defined(__GNUC__) && (__GNUC__ >= 3))  \
+  || (defined(__INTEL_COMPILER) && (__INTEL_COMPILER >= 800)) \
+  || defined(__clang__)
+#    define XXH_likely(x) __builtin_expect(x, 1)
+#    define XXH_unlikely(x) __builtin_expect(x, 0)
+#else
+#    define XXH_likely(x) (x)
+#    define XXH_unlikely(x) (x)
+#endif
+
 #if defined(__GNUC__)
 #  if defined(__AVX2__)
 #    include <immintrin.h>
@@ -563,7 +573,7 @@ XXH3_len_4to8_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_h
 {
     XXH_ASSERT(input != NULL);
     XXH_ASSERT(secret != NULL);
-    XXH_ASSERT(4 <= len && len <= 8);
+    XXH_ASSERT(4 <= len && len < 8);
     {   xxh_u32 const input_1 = XXH_readLE32(input);
         xxh_u32 const input_2 = XXH_readLE32(input + len - 4);
         xxh_u64 const input_64 = input_2 | ((xxh_u64)input_1 << 32);
@@ -578,7 +588,7 @@ XXH3_len_9to16_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_
 {
     XXH_ASSERT(input != NULL);
     XXH_ASSERT(secret != NULL);
-    XXH_ASSERT(9 <= len && len <= 16);
+    XXH_ASSERT(8 <= len && len <= 16);
     {   xxh_u64 const input_lo = XXH_readLE64(input)           ^  XXH_readLE64(secret);
         xxh_u64 const input_hi = XXH_readLE64(input + len - 8) ^ (XXH_readLE64(secret + 8) - seed);
         xxh_u64 const acc = len + (input_lo + input_hi) + XXH3_mul128_fold64(input_lo, input_hi);
@@ -590,8 +600,8 @@ XXH_FORCE_INLINE XXH64_hash_t
 XXH3_len_0to16_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_hash_t seed)
 {
     XXH_ASSERT(len <= 16);
-    {   if (len > 8) return XXH3_len_9to16_64b(input, len, secret, seed);
-        if (len >= 4) return XXH3_len_4to8_64b(input, len, secret, seed);
+    {   if (XXH_likely(len >= 8)) return XXH3_len_9to16_64b(input, len, secret, seed);
+        if (XXH_likely(len >= 4)) return XXH3_len_4to8_64b(input, len, secret, seed);
         if (len) return XXH3_len_1to3_64b(input, len, secret, seed);
         return XXH3_avalanche((PRIME64_1 + seed) ^ XXH_readLE64(secret));
     }
--- a/xxh3.h
+++ b/xxh3.h
@@ -532,7 +532,7 @@ XXH3_mul128_fold64(xxh_u64 lhs, xxh_u64 rhs)
 static XXH64_hash_t XXH3_avalanche(xxh_u64 h64)
 {
     h64 ^= h64 >> 37;
-    h64 *= PRIME64_3;
+    h64 *= 0x165667919E3779F9ULL;
     h64 ^= h64 >> 32;
     return h64;
 }
@@ -551,7 +551,7 @@ XXH3_len_1to3_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_h
     {   xxh_u8 const c1 = input[0];
         xxh_u8 const c2 = input[len >> 1];
         xxh_u8 const c3 = input[len - 1];
-        xxh_u32  const combined = ((xxh_u32)c1) | (((xxh_u32)c2) << 8) | (((xxh_u32)c3) << 16) | (((xxh_u32)len) << 24);
+        xxh_u32  const combined = ((xxh_u32)c1<<16) | (((xxh_u32)c2) << 24) | (((xxh_u32)c3) << 0) | (((xxh_u32)len) << 8);
         xxh_u64  const keyed = (xxh_u64)combined ^ (XXH_readLE32(secret) + seed);
         xxh_u64  const mixed = keyed * PRIME64_1;
         return XXH3_avalanche(mixed);
@@ -564,12 +564,12 @@ XXH3_len_4to8_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_h
     XXH_ASSERT(input != NULL);
     XXH_ASSERT(secret != NULL);
     XXH_ASSERT(4 <= len && len <= 8);
-    {   xxh_u32 const input_lo = XXH_readLE32(input);
-        xxh_u32 const input_hi = XXH_readLE32(input + len - 4);
-        xxh_u64 const input_64 = input_lo | ((xxh_u64)input_hi << 32);
-        xxh_u64 const keyed = input_64 ^ (XXH_readLE64(secret) + seed);
-        xxh_u64 const mix64 = len + ((keyed ^ (keyed >> 51)) * PRIME32_1);
-        return XXH3_avalanche((mix64 ^ (mix64 >> 47)) * PRIME64_2);
+    {   xxh_u32 const input_1 = XXH_readLE32(input);
+        xxh_u32 const input_2 = XXH_readLE32(input + len - 4);
+        xxh_u64 const input_64 = input_2 | ((xxh_u64)input_1 << 32);
+        xxh_u64 const keyed = input_64 ^ (seed + XXH_readLE64(secret));
+        xxh_u64 const mix64 = len + ((keyed ^ (keyed >> 39)) * PRIME64_2);
+        return XXH3_avalanche(mix64 ^ (mix64 >> 47));
     }
 }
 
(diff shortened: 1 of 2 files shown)


easyaspi314:
XXH128 does have a lot of latency on short keys, it would be nice to see some improvements.

For keys <= 16 bytes:

32-bit: XXH32 > XXH3_64b >>> XXH64 > XXH128
x86_64: XXH32 ≈ XXH3_64b > XXH128 > XXH64
AArch64: XXH3_64b > XXH32 > XXH128 > XXH64

XXH3 is faster than XXH32 on AArch64 mainly because it can xor fold without any temporaries:
```asm
        eor     x0, x0, x0, lsr #47
```
compared to x86_64:
```asm
        mov     rcx, rax
        shr     rcx, 47
        xor     rax, rcx
```

xxh3.h
```diff
@@ -579,7 +579,7 @@ XXH3_len_9to16_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_
     XXH_ASSERT(input != NULL);
     XXH_ASSERT(secret != NULL);
     XXH_ASSERT(9 <= len && len <= 16);
-    {   xxh_u64 const input_lo = XXH_readLE64(input)           ^ (XXH_readLE64(secret)     + seed);
+    {   xxh_u64 const input_lo = XXH_readLE64(input)           ^  XXH_readLE64(secret);
```
easyaspi314: I presume this is to prevent the seed from being roughly cancelled out when we add them together below?

xxh3.h
```diff
@@ -564,12 +564,12 @@ XXH3_len_4to8_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_h
...
-        xxh_u64 const input_64 = input_lo | ((xxh_u64)input_hi << 32);
-        xxh_u64 const keyed = input_64 ^ (XXH_readLE64(secret) + seed);
-        xxh_u64 const mix64 = len + ((keyed ^ (keyed >> 51)) * PRIME32_1);
-        return XXH3_avalanche((mix64 ^ (mix64 >> 47)) * PRIME64_2);
+    {   xxh_u32 const input_1 = XXH_readLE32(input);
+        xxh_u32 const input_2 = XXH_readLE32(input + len - 4);
+        xxh_u64 const input_64 = input_2 | ((xxh_u64)input_1 << 32);
+        xxh_u64 const keyed = input_64 ^ (seed + XXH_readLE64(secret));
+        xxh_u64 const mix64 = len + ((keyed ^ (keyed >> 39)) * PRIME64_2);
+        return XXH3_avalanche(mix64 ^ (mix64 >> 47));
```
easyaspi314: Removing the multiply improves the strength? Interesting.

xxh3.h
Cyan4973: Yes

xxh3.h
Cyan4973: Nope, it's not stronger. I actually had to select all others parameters very carefully to compensate this operation.
But removing one multiplication from the serial path does improve speed, obviously. That was the main point.

Cyan4973:
Dropping the requirement of bijectivity for 64-bit inputs, the speed climbs to > 150 MH/s for the 4-7 range, and > 160 MH/s for the length 8 itself.
So it's pretty big boost.

Apart from the loss of bijectivity, all other quality metrics seem preserved (good dispersion, collisions in accordance with birthday paradox).

xxh3.h
easyaspi314: Ah, I see. It did have the most multiplies of the three:

| Length | 64-bit muls | 32-bit muls |
|:---|---:|----:|
| 1-3 | 2 | 6 |
| 4-8 | 3 | 8 |
| 9-16 | 2 | 7 |

easyaspi314:
What does the non-bijective version look like?

easyaspi314:
On my Pixel:

64-bit
```
./xxhsum 0.7.3 (64-bits aarch64 little endian), Clang 9.0.1 , by Yann Collet
Sample of 5 bytes...
XXH32                    :          5 -> 78979139 it/s (  376.6 MB/s)
XXH32 unaligned          :          5 -> 80623227 it/s (  384.4 MB/s)
XXH64                    :          5 -> 52919520 it/s (  252.3 MB/s)
XXH64 unaligned          :          5 -> 50402130 it/s (  240.3 MB/s)
XXH3_64b                 :          5 -> 99918214 it/s (  476.4 MB/s)
XXH3_64b unaligned       :          5 -> 100049695 it/s (  477.1 MB/s)
XXH3_64b seeded          :          5 -> 94150987 it/s (  448.9 MB/s)
XXH3_64b seeded unaligne :          5 -> 94381426 it/s (  450.0 MB/s)
XXH128                   :          5 -> 66374743 it/s (  316.5 MB/s)
XXH128 unaligned         :          5 -> 66631507 it/s (  317.7 MB/s)
XXH128 seeded            :          5 -> 61666322 it/s (  294.0 MB/s)
XXH128 seeded unaligned  :          5 -> 61957284 it/s (  295.4 MB/s)
XXH3_new                 :          5 -> 104639925 it/s (  499.0 MB/s)
XXH3_new unaligned       :          5 -> 105611273 it/s (  503.6 MB/s)
XXH3_new seeded          :          5 -> 100946493 it/s (  481.4 MB/s)
XXH3_new seeded unaligne :          5 -> 100562905 it/s (  479.5 MB/s)
```
```
clang -O3 -m32 -march=armv6t2
$ ./xxhsum -b -B5
Sample of 5 bytes...
XXH32                    :          5 -> 57626760 it/s (  274.8 MB/s)
XXH32 unaligned          :          5 -> 59118209 it/s (  281.9 MB/s)
XXH64                    :          5 -> 27138541 it/s (  129.4 MB/s)
XXH64 unaligned          :          5 -> 25513446 it/s (  121.7 MB/s)
XXH3_64b                 :          5 -> 45701062 it/s (  217.9 MB/s)
XXH3_64b unaligned       :          5 -> 45433547 it/s (  216.6 MB/s)
XXH3_64b seeded          :          5 -> 43698427 it/s (  208.4 MB/s)
XXH3_64b seeded unaligne :          5 -> 43364902 it/s (  206.8 MB/s)
XXH128                   :          5 -> 27789511 it/s (  132.5 MB/s)
XXH128 unaligned         :          5 -> 27637196 it/s (  131.8 MB/s)
XXH128 seeded            :          5 -> 25321305 it/s (  120.7 MB/s)
XXH128 seeded unaligned  :          5 -> 25297310 it/s (  120.6 MB/s)
XXH3_new                 :          5 -> 50940968 it/s (  242.9 MB/s)
XXH3_new unaligned       :          5 -> 50938749 it/s (  242.9 MB/s)
XXH3_new seeded          :          5 -> 55158194 it/s (  263.0 MB/s)
XXH3_new seeded unaligne :          5 -> 54682827 it/s (  260.7 MB/s)
```

It's interesting how the seeded variant is faster. I'm guessing it takes advantage of a barrel shift or a multiply+accumulate.

easyaspi314:
Oops, just realized I forgot `-munaligned-access -fomit-frame-pointer` which I usually use.

Here is `clang -O3 -m32 -march=armv6t2 -munaligned-access -fomit-frame-pointer`

```
./xxhsum 0.7.3 (32-bits arm little endian), Clang 9.0.1 , by Yann Collet
Sample of 5 bytes...
XXH32                    :          5 -> 73493698 it/s (  350.4 MB/s)
XXH32 unaligned          :          5 -> 71677171 it/s (  341.8 MB/s)
XXH64                    :          5 -> 29337822 it/s (  139.9 MB/s)
XXH64 unaligned          :          5 -> 29028419 it/s (  138.4 MB/s)
XXH3_64b                 :          5 -> 52521477 it/s (  250.4 MB/s)
XXH3_64b unaligned       :          5 -> 52145183 it/s (  248.6 MB/s)
XXH3_64b seeded          :          5 -> 54742558 it/s (  261.0 MB/s)
XXH3_64b seeded unaligne :          5 -> 54912692 it/s (  261.8 MB/s)
XXH128                   :          5 -> 31115296 it/s (  148.4 MB/s)
XXH128 unaligned         :          5 -> 31410234 it/s (  149.8 MB/s)
XXH128 seeded            :          5 -> 29905792 it/s (  142.6 MB/s)
XXH128 seeded unaligned  :          5 -> 29889476 it/s (  142.5 MB/s)
XXH3_new                 :          5 -> 73039944 it/s (  348.3 MB/s)
XXH3_new unaligned       :          5 -> 72427877 it/s (  345.4 MB/s)
XXH3_new seeded          :          5 -> 66844700 it/s (  318.7 MB/s)
XXH3_new seeded unaligne :          5 -> 68765537 it/s (  327.9 MB/s)
```

That benchmark better matches the AArch64 version, as they shouldn't be too far off thanks ARM's fast 64-bit multiplies.

Cyan4973:
> What does the non-bijective version look like?

There are multiple versions possible, but I would start by one resembling the 8-16 segment, which has already proven its worth in term of bit dispersion.
I will make it a separate branch, so that it can be compared with the current one.

Thanks for your measurements on ARM, it's very useful. It shows that this variant is already a sensible improvement for ARM. It will be useful again when comparing with the non-bijective variant.

easyaspi314:
No problem.

As for bijectivity, as long as it doesn't have any major collisions, I don't think it is necessary.

Besides, in practice, inputs will be all different lengths.

easyaspi314:
As for performance of the 9-16 variant:
 - 64-bit is, unsurprisingly, no contest.
 - 32-bit is faster than XXH32 on lengths 11 and 13-16.

Of course, ARM is naturally going to be better at the 9-16 routine than x86, as the long multiply is only 4 (8 without `umaal`) instructions compared to the 20-30 required on x86.

easyaspi314:
If it helps, here is the output of `./benchHash --maxl=0 --maxs=16`. 
`clang -O3`
[aarch64.txt](https://github.com/Cyan4973/xxHash/files/4222802/aarch64.txt)
`clang -O3 -m32 -march=armv6t2 -munaligned-access -fomit-frame-pointer`
[armv6t2.txt](https://github.com/Cyan4973/xxHash/files/4222817/armv6t2.txt)

easyaspi314:
LMK if you have something that you want me to bench.

easyaspi314:
[I made spreadsheets/charts of the latency for aarch64, arm, x86, and x86_64](https://docs.google.com/spreadsheets/d/1Gb6SFF2gsndmelb_Lq2mBaaTJ1Z-pb6_qhaAn3aW1ak/edit?usp=sharing)

I realized that Clang is vectorizing the XXH128 short inputs on i686, significantly reducing the speed. I wonder if I can fix this.

xxh3.h
```diff
@@ -578,7 +588,7 @@ XXH3_len_9to16_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_
 {
```
easyaspi314: Shouldn't you rename it to `XXH3_len_8to16_64b`?

easyaspi314:
```
Build started
git clone -q --depth=2 --branch=smallFold https://github.com/Cyan4973/xxHash.git C:\projects\xxhash
git checkout -qf 6456c04490667acf035884e91aac32208d09dbbe
fatal: reference is not a tree: 6456c04490667acf035884e91aac32208d09dbbe
Command exited with code 128
```
Not sure what that is about.

easyaspi314:
Added the x86 benchmarks to the spreadsheet for your `smallFold` branch I noticed, will update ARM benches later.

xxh3.h
Cyan4973: yeah, either this, or revert `len==8` into `4to8`.
nothing decided yet, this is just test.

easyaspi314: Ok

Cyan4973:
Thanks for benchmarking @easyaspi314 ,
from these measurements, it seems that `smallFold` is beneficial for `x64`, but detrimental for `x86`.

Note that also measured speed on a local `aarch64` system,
and `smallFold` seems detrimental on this system too (compared to `smallInputs`).

easyaspi314:
That is to be expected.

aarch64 has a significant overhead on 64-bit and 64->128-bit multiplies.

Unlike most x86_64 chips which have basically the same speed, most aarch64 chips have a 5 cycle `madd/mul` (x-form) and a 6 cycle `umulh`, and some chips even stall the pipeline a bit. This means that `XXH_mult64to128` takes roughly 11 cycles on average, compared to x86_64's 2-4 cycles.

That is also the reason why XXH64 is roughly the same speed as XXH32: The speedup from increased bandwidth is cancelled out by the slower `mul`.

easyaspi314:
ARM benchmarks added for smallFold and smallFold32.

Cyan4973:
Superseded by #308

PR #306, opened 2020-02-20
mul128fold64 for len_4to8 

Cyan4973:
This is an alternative formula for the `4to8` segment.
It uses the `mul128fold64` instruction, similar to the `9to16` segment.

This variant seems faster than `smallInputs` on `x64`,
but it also seems slower for `x86` and `arm`.
So it's not a perfect replacement.


Cyan4973:
This variant improves speed for `x64`,
but makes it worse for 32-bit `x86` and for `ARM` targets.

I believe the overall trade off is not worth it.
So I'll close this branch.

PR #307, opened 2020-02-20
mul32to64 for len_4to8

Cyan4973:
This is an alternative formula for the `4to8` segment.
It uses the `mul32to64` instruction, which is also used in the inner loop for long inputs.

This variant seems all around faster, especially for `x86` and `arm`.

Quality still needs to be tested. I'm relatively confident for lengths `4to7`,
but len8 itself must be thoroughly checked.

_edit_ : indeed, this fails the newer "Perlin Noise" test, although relatively narrowly (a handful of collisions, while none was expected). Should be fixable.

_edit 2_ : latest update fixes the Perlin Noise issue. There are still some (small) weaknesses with regards to Differential Distribution, which specifically targets `len8`.

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -576,10 +576,9 @@ XXH3_len_4to8_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_h
     XXH_ASSERT(4 <= len && len < 8);
     {   xxh_u32 const input_1 = XXH_readLE32(input);
         xxh_u32 const input_2 = XXH_readLE32(input + len - 4);
-        xxh_u64 const input_64 = input_2 | ((xxh_u64)input_1 << 32);
-        xxh_u64 const keyed = input_64 ^ (seed + XXH_readLE64(secret));
-        xxh_u64 const mix64 = len + ((keyed ^ (keyed >> 39)) * PRIME64_2);
-        return XXH3_avalanche(mix64 ^ (mix64 >> 47));
+        xxh_u32 const key1 = input_1 ^ ((seed >> 32) + XXH_readLE32(secret));
+        xxh_u32 const key2 = input_2 ^ (seed + XXH_readLE32(secret+4));
+        return XXH3_avalanche(XXH_mult32to64(key1, key2) + len);
     }
 }
 
--- a/xxh3.h
+++ b/xxh3.h
@@ -577,7 +577,7 @@ XXH3_len_4to8_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_h
     seed = XXH_rotl64(seed, 13);
     {   xxh_u32 const input1 = XXH_readLE32(input);
         xxh_u32 const input2 = XXH_readLE32(input + len - 4);
-        xxh_u32 const key1 = input1 ^ ((xxh_u32)(seed >> 32) + XXH_readLE32(secret));
+        xxh_u32 const key1 = XXH_swap32(input1) ^ ((xxh_u32)(seed >> 32) + XXH_readLE32(secret));
         xxh_u32 const key2 = input2 ^ (XXH_readLE32(secret+4) - (xxh_u32)seed);
         xxh_u64 const mix = XXH_mult32to64(key1, key2)
                           + input1
--- a/xxh3.h
+++ b/xxh3.h
@@ -574,11 +574,11 @@ XXH3_len_4to8_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_h
     XXH_ASSERT(input != NULL);
     XXH_ASSERT(secret != NULL);
     XXH_ASSERT(4 <= len && len < 8);
-    {   xxh_u32 const input_1 = XXH_readLE32(input);
-        xxh_u32 const input_2 = XXH_readLE32(input + len - 4);
-        xxh_u32 const key1 = input_1 ^ ((seed >> 32) + XXH_readLE32(secret));
-        xxh_u32 const key2 = input_2 ^ (seed + XXH_readLE32(secret+4));
-        return XXH3_avalanche(XXH_mult32to64(key1, key2) + len);
+    {   xxh_u32 const input1 = XXH_readLE32(input);
+        xxh_u32 const input2 = XXH_readLE32(input + len - 4);
+        xxh_u32 const key1 = input1 ^ ((xxh_u32)(seed >> 32) + XXH_readLE32(secret));
+        xxh_u32 const key2 = input2 ^ ((xxh_u32)seed + XXH_readLE32(secret+4));
+        return XXH3_avalanche(XXH_mult32to64(key1, key2) + input1 + input2 + len);
     }
 }
 
--- a/xxh3.h
+++ b/xxh3.h
@@ -574,11 +574,16 @@ XXH3_len_4to8_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_h
     XXH_ASSERT(input != NULL);
     XXH_ASSERT(secret != NULL);
     XXH_ASSERT(4 <= len && len <= 8);
+    seed = XXH_rotl64(seed, 13);
     {   xxh_u32 const input1 = XXH_readLE32(input);
         xxh_u32 const input2 = XXH_readLE32(input + len - 4);
         xxh_u32 const key1 = input1 ^ ((xxh_u32)(seed >> 32) + XXH_readLE32(secret));
-        xxh_u32 const key2 = input2 ^ ((xxh_u32)seed + XXH_readLE32(secret+4));
-        return XXH3_avalanche(XXH_mult32to64(key1, key2) + input1 + input2 + len);
+        xxh_u32 const key2 = input2 ^ (XXH_readLE32(secret+4) - (xxh_u32)seed);
+        xxh_u64 const mix = XXH_mult32to64(key1, key2)
+                          + input1
+                          + input2
+                          + len;
+        return XXH3_avalanche(mix);
     }
 }
 
--- a/xxh3.h
+++ b/xxh3.h
@@ -58,6 +58,16 @@
 #  define XXH_RESTRICT   /* disable */
 #endif
 
+#if (defined(__GNUC__) && (__GNUC__ >= 3))  \
+  || (defined(__INTEL_COMPILER) && (__INTEL_COMPILER >= 800)) \
+  || defined(__clang__)
+#    define XXH_likely(x) __builtin_expect(x, 1)
+#    define XXH_unlikely(x) __builtin_expect(x, 0)
+#else
+#    define XXH_likely(x) (x)
+#    define XXH_unlikely(x) (x)
+#endif
+
 #if defined(__GNUC__)
 #  if defined(__AVX2__)
 #    include <immintrin.h>
@@ -563,7 +573,7 @@ XXH3_len_4to8_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_h
 {
     XXH_ASSERT(input != NULL);
     XXH_ASSERT(secret != NULL);
-    XXH_ASSERT(4 <= len && len <= 8);
+    XXH_ASSERT(4 <= len && len < 8);
     {   xxh_u32 const input_1 = XXH_readLE32(input);
         xxh_u32 const input_2 = XXH_readLE32(input + len - 4);
         xxh_u64 const input_64 = input_2 | ((xxh_u64)input_1 << 32);
@@ -578,7 +588,7 @@ XXH3_len_9to16_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_
 {
     XXH_ASSERT(input != NULL);
     XXH_ASSERT(secret != NULL);
-    XXH_ASSERT(9 <= len && len <= 16);
+    XXH_ASSERT(8 <= len && len <= 16);
     {   xxh_u64 const input_lo = XXH_readLE64(input)           ^  XXH_readLE64(secret);
         xxh_u64 const input_hi = XXH_readLE64(input + len - 8) ^ (XXH_readLE64(secret + 8) - seed);
         xxh_u64 const acc = len + (input_lo + input_hi) + XXH3_mul128_fold64(input_lo, input_hi);
@@ -590,8 +600,8 @@ XXH_FORCE_INLINE XXH64_hash_t
 XXH3_len_0to16_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_hash_t seed)
 {
     XXH_ASSERT(len <= 16);
-    {   if (len > 8) return XXH3_len_9to16_64b(input, len, secret, seed);
-        if (len >= 4) return XXH3_len_4to8_64b(input, len, secret, seed);
+    {   if (XXH_likely(len >= 8)) return XXH3_len_9to16_64b(input, len, secret, seed);
+        if (XXH_likely(len >= 4)) return XXH3_len_4to8_64b(input, len, secret, seed);
         if (len) return XXH3_len_1to3_64b(input, len, secret, seed);
         return XXH3_avalanche((PRIME64_1 + seed) ^ XXH_readLE64(secret));
     }
--- a/xxh3.h
+++ b/xxh3.h
@@ -569,7 +569,7 @@ XXH3_len_1to3_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_h
 }
 
 XXH_FORCE_INLINE XXH64_hash_t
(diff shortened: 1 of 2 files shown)


easyaspi314:
> edit : indeed, this fails the newer "Perlin Noise" test, although relatively narrowly (a handful of collisions, while none was expected). Should be fixable.

How does it compare to `dev`?

easyaspi314:
Maybe we need to mix better, as 32-bit to 64-bit multiplication has a lot of bias, which isn't mitigated by the xor fold of `mul128_fold64`

Google documents this in HighwayHash:
```c
    // Multiplication mixes/scrambles bytes 0-7 of the 64-bit result to
    // varying degrees. In descending order of goodness, bytes
    // 3 4 2 5 1 6 0 7 have quality 228 224 164 160 100 96 36 32.
    // As expected, the upper and lower bytes are much worse.
```

Cyan4973:
> How does it compare to dev?

`dev` doesn't fail any quality test,
but it's much slower.

easyaspi314:
> `dev` doesn't fail any quality test, but it's much slower on the `len_4to8` segment.

Gotcha.

This is definitely a lot faster.

Cyan4973:
The differential distribution test could fail with `mul32to64`.

The diffDist test is specifically targeting 64-bit inputs for 64bit hashes.

It's not a huge fail : the original test itself, looking for 64-bit collisions, was fine (0 collisions). The extended truncation to 32-bit was fine. But the advanced test, which extracts a variable N nb of bits from both the top and lower part, could fail for some specific values of N, and even then, not by much (from 2x to 4x, which is a 1 to 2 bits dispersion loss).

Anyway, if this issue is deemed worth solving, a quick fix is to make `len==8` part of the `8to16` segment, instead of part of the `4to8` segment. In which case, it uses the `mul128fold64` instruction instead, making it slower than `4-7`, though identical to `9-16`, and passing all tests.

This is the suggested approach in the last commit.

easyaspi314:
> Anyway, if this issue is deemed worth solving, a quick fix is to make `len==8` part of the `8to16` segment, instead of part of the `4to8` segment.

So basically from what I am reading, just cheat SMHasher?

Or is it just the lack of overlap that makes the 8b worse than the others?

Cyan4973:
> is it just the lack of overlap that makes the 8b worse than the others?

Indeed, I presume that the `mul32to64` instruction produces a larger space contraction than `mul128fold64`. Calculating "how much" is not trivial, but I remember measuring something like a ~5-6 bits contraction quite some time ago, versus ~2 bits.

For inputs up to 7 bytes long, since output space is significantly larger than input space, I believe the contraction introduced by `mul32to64` is not large enough to be problematic. But for the full 64-bits inputs, the expectation is that it will show, one way or another. Adding `input1 + input2` in the loop help expand the space a bit, but it's unclear if that's enough.

To be fair, the results I've measured with `mul32to64` are not bad, and I would even qualify them as quite good, better than my expectation anyway. It's still not "good enough", but it's not that far off.

Anyway, if perfection is the goal, `mul128fold64` is closer to it.

> So basically from what I am reading, just cheat SMHasher?

It's questionable that SMHasher's `DiffDist` test specializes in same input-output size transformation only. Maybe extending it to more input lengths would be a good idea, avoiding the feeling of optimizing for just a single input length.

easyaspi314:
> It's questionable that SMHasher's DiffDist test specializes in same input-output size transformation only. Maybe extending it to more input lengths would be a good idea, avoiding the feeling of optimizing for just a single input length.

Even though it is open source, it has always felt like a black box.

Plus some things don't make sense: `halfsiphash` apparently struggles with the `zeroes` test, but `siphash` does not, despite being *literally the same algorithm but downscaled*

easyaspi314:
Plus, SMHasher literally ignores

easyaspi314:
Here are some basic assembly versions I wrote for armv6t2, aarch64, i686, and x86_64.

It excludes kSecret/seed being folded out for simplicity.

Seems to adapt pretty well.

```asm
@ armv6t2 (unaligned access)
@ r0: inputP, r1: len, r2: secretP, [sp]{0-7}: seed
@ could be interleaved better
XXH3_len4to8_64b:
        push    {r4, lr}
        @ load seed
        ldr     r4, [sp, #8]
        ldr     r12, [sp, #12]
        @ {r12,lr} = XXH_rotl64(seed, 13)
        lsl     r12, r4, #13
        orr     r12, r12, lr, lsr #(32 - 13)
        lsl     lr, lr, #13
        orr     lr, lr, r4, lsr #(32 - 13)
        @ secret1 = XXH_readLE32(secret)
        ldr     r3, [r2]
        @ input1 = XXH_readLE32(input)
        ldr     r4, [r0]
        @ input += len
        add     r0, r0, r1
        @ secret1 += seed >> 32
        add     lr, lr, r3
        @ secret2 = XXH_readLE32(secret + 4)
        ldr     r2, [r2, #4]
        @ input1 = XXH_readLE32(input + len - 4)
        ldr     r0, [r0, #-4]
        @ secret2 -= (u32)seed
        sub     r2, r2, r12
        @ secret1 ^= input1
        eor     lr, lr, r4
        @ secret2 ^= input2
        eor     r2, r2, r0
        @ ret {r4,r12} = (u64)secret1 * (u64)secret2 + (u64)input1 + (u64)input2
        umaal   r0, r4, r2, lr
        @ start to load 0x165667919E3779F9
        movw    r2, #0x79f9
        movt    r2, #0x9e37
        @ ret += len
        adds    r0, r0, r1
        adc     r1, r4, #0
        @ avalanche
        @ ret ^= ret >> 37
        eor     r0, r0, r1, lsr #(37 - 32)
        @ ret *= 0x165667919E3779F9
        umull   r12, r4, r2, r0
        movw    r3, #0x6791
        mla     r4, r2, r1, r4
        movt    r3, #0x1656
        mla     r1, r0, r3, r4
        @ ret ^= ret >> 32
        eor     r0, r12, r1
        @ return ret
        pop     {r4, pc}

// aarch64
// x0: inputP, x1: len, x2: secretP, x3: seed
XXH3_len_4to8_64b:
        // seed = XXH_rotl64(seed, 13)
        ror     x3, x3, #(64 - 13)
        // secret1 = XXH_readLE32(secret), secret2 = XXH_readLE32(secret+4)
        ldp     w4, w5, [x2]
        // seedHi = seed >> 32
        lsr     x6, x3, #32
        // secret2 -= (u32)seed
        sub     w5, w5, w3
        // input1 = XXH_readLE32(input)
        ldr     w7, [x0]
        // secret1 += seed >> 32
        add     w4, w4, w6
        // input + len
        add     x0, x0, x1
        // secret1 ^= input1
        eor     w4, w4, w7
        // input2 = XXH_readLE32(input + len - 4)
        ldr     w0, [x0, #-4]
        // secret2 ^= input2
        eor     w5, w5, w0
        // len += (u64)input1
        add     x1, x1, x7
        // start loading 0x165667919E3779F9
        mov     x2, #0x79f9
        // ret x0 = (u64)secret1 + (u64)secret2 + (u64)input2
        umaddl  x0, w4, w5, x0
        movk    x2, #0x9e37, lsl #16
        movk    x2, #0x6791, lsl #32
        // ret += len + input1
        add     x0, x0, x1
        movk    x2, #0x1656, lsl #48
        // avalanche
        // ret ^= ret >> 37
        eor     x0, x0, x0, lsr #37
        // ret *= 0x165667919E3779F9
        mul     x0, x0, x2
        // ret ^= ret >> 32
        eor     x0, x0, x0, lsr #32
        // return ret
        ret

# i686 (cdecl)
# ebp+8: inputP, ebp+12: len, ebp+16: secretP, ebp+20/24: seed
XXH3_len_4to8_64b:
        push    ebp
        mov     ebp, esp
        push    esi
        push    ebx
        push    edi
        # load seed
        mov     eax, dword ptr [ebp + 24]
        mov     ecx, dword ptr [ebp + 20]
        # seed {edx,eax} = XXH_rotl64(seed, 13)
        mov     edx, ecx
        shld    edx, eax, 13
        shld    eax, ecx, 13
        # load input ptr
        mov     edi, dword ptr [ebp + 8]
        # load len
        mov     esi, dword ptr [ebp + 12]
        # input1 = XXH_readLE32(input)
        mov     ecx, dword ptr [edi]
        # input2 = XXH_readLE32(input + len - 4)
        mov     ebx, dword ptr [edi + esi - 4]
        # load secret ptr
        mov     esi, dword ptr [ebp + 16]
        # secret1 = (seed >> 32) + XXH_readLE32(secret)
        add     eax, dword ptr [esi]
        # secret2 = XXH_readLE32(secret+4)
        mov     esi, dword ptr [esi + 4]
        # secret2 -= (u32)seed
        sub     esi, edx
        # secret1 ^= input1
        xor     eax, ecx
        # secret2 ^= input2
        xor     esi, ebx
        # ret = (u64)secret1 * (u64)secret2
        mul     esi
        # input2 += len
        add     ebx, dword ptr [ebp + 12]
        adc     edi, 0
        # ret += input1
        add     eax, ecx
        adc     edx, 0
        # ret += input2 + len
        add     eax, ebx
        adc     edx, edi
        # avalanche
        # ret ^= ret >> 37
        mov     edi, edx
        shr     edi, 37 - 32
        xor     eax, edi
        # ret *= 0x165667919E3779F9
        imul    ebx, eax, 0x16566791 # lo_hi
        mov     ecx, 0x9E3779F9
        imul    edi, edx, ecx        # hi_lo
        add     edi, ebx             # upper
        mul     ecx
        add     edx, edi
        # ret ^= ret >> 32
        xor     eax, edx
        # return ret
        pop     edi
        pop     ebx
        pop     esi
        mov     esp, ebp
        pop     ebp
        ret

# x86_64 (sysv)
# rdi: inputP, rsi: len, rdx: secretP, r8: seed
XXH3_len_4to8_64b:
        # seed = XXH_rotl64(seed, 13)
        rol     r8, 13
        # seedHi = (u32)(seed >> 32)
        mov     rax, r8
        shr     rax, 32
        # secretHi = XXH_readLE32(secret + 4)
        mov     ecx, dword ptr [rdx + 4]
        # secret1 = (seed >> 32) + XXH_readLE32(secret)
        add     eax, dword ptr [rdx]
        # input1 = XXH_readLE32(input)
        mov     r10d, dword ptr [rdi]
        # secret2 = secretHi - (u32)seed
        sub     ecx, r8d
        # input2 = XXH_readLE32(input + len - 4)
        mov     edx, dword ptr [rdi + rsi - 4]
        # secret1 ^= input1
        xor     ecx, r10d
        # secret2 ^= input1
        xor     eax, edx
        # ret = (u64)secret1 * (u64)secret2
        imul    rax, rcx, rax
        # input2 += len
        add     rdx, rsi
        # ret += input1
        add     rax, r10
        # ret += input2 + len
        add     rax, rdx
        # avalanche
        movabs  rdx, 0x165667919E3779F9
        # ret ^= ret >> 37
        mov     rcx, rax
        shr     rcx, 37
        xor     rax, rcx
        # ret *= 0x165667919E3779F9
        imul    rax, rax, rdx
        # ret ^= ret >> 32
        mov     rcx, rax
        shr     rcx, 32
        xor     rax, rcx
        # return ret
        ret
```

Cyan4973:
Interesting.
One difference I could spot is that `len` seems to be joined directly inside `input1`, instead of later, after the multiplication. Are there other differences ?

Do you suggest this version has an advantage ?

easyaspi314:
It is just taking advantage of the commutative property, avoiding a dependency chain.

```c
     a += b;
     a += c;
     a += d;
```
becomes
```c
    a += b;
    c += d;
    a += c;
```

I didn't have any rhyme or reason to choosing `len`.

easyaspi314:
```c
XXH_FORCE_INLINE XXH64_hash_t
XXH3_len_4to8_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_hash_t seed)
{
    XXH_ASSERT(input != NULL);
    XXH_ASSERT(secret != NULL);
    XXH_ASSERT(4 <= len && len <= 8);
    seed = XXH_rotl64(seed, 13);
    {   xxh_u32 input1 = XXH_readLE32(input);
        xxh_u32 input2 = XXH_readLE32(input + len - 4);
        xxh_u32 key1 = input1 ^ ((xxh_u32)(seed >> 32) + XXH_readLE32(secret));
        xxh_u32 key2 = input2 ^ (XXH_readLE32(secret+4) - (xxh_u32)seed);
        xxh_u64 mix = XXH_mult32to64(key1, key2)
                          + input1
                          + input2;
                       /* + len; (split for ARMv6 hack) */
       /* clang && ARM && has UMAAL */
#if defined(__clang__) /* XXX; does GCC ARM also need this? */ \
     && __ARM_ARCH >= 6 /* UMAAL needs ARMv6+ */ \
     && ((defined(__arm__) && !defined(__thumb__)) || defined(__thumb2__)) /* ARM/Thumb-2 */ \
     && __ARM_ARCH_PROFILE != 'M' /* ARMv7-M's Thumb-2 subset lacks UMAAL */
       /*
        * UGLY HACK: Clang will always convert this to something like the following:
        *
        *   u64 tmp = (u64)input1 + (u64)len;
        *   u64 ret = (u64)key1 * (u64)key2;
        *   ret += tmp;
        *   ret += input2;
        *
        * Usually, this is a good generic optimization which reduces dependency
        * chaining, but on ARMv6+, it prevents the blatantly obvious UMAAL
        * optimization (see XXH_mult64to128)
        *
        * Most optimal output:
        *
        *       @ tmp (input1,input2) = (u64)key1 * (u64)key2 + input1 + input2
        *       umaal   input1, input2, key1, key2
        *       @ ret = tmp + (u64)len
        *       adds    r0, tmp.lo, len
        *       adc     r1, tmp.hi, #0
        *
        * Clang 9.1.0:
        *
        *       @ tmp = (u64)input1 + (u64)len
        *       mov     tmp.hi, #0
        *       adds    tmp.lo, input1, len
        *       adc     tmp.hi, tmp.hi, #0
        *       @ tmp += (u64)key1 * (u64)key2
        *       umlal   tmp.lo, tmp.hi, key1, key2
        *       @ ret = tmp + input2
        *       adds    r0, tmp.lo, input2
        *       adc     r1, tmp.hi, #0
        *
        * This barrier splits the third add, causing Clang to optimize to UMAAL
        * instead.
        */
        __asm__("" : "+r" (mix));
#endif
        mix += len;
        return XXH3_avalanche(mix);
    }
}
```
\*sigh\* 

* Slightly faster on `-O3` (ARM)
* 76 bytes smaller on `-O3` (ARM)
* 132 bytes smaller on `-O3` (Thumb-2)
* 12 bytes smaller on `-Oz` (ARM)
* 8 bytes smaller on `-Oz` (Thumb-2)

Nah, not worth it.

Cyan4973:
That's a pretty nice trick.

132 bytes smaller just for `UMAAL` ? That seems like an outsized impact considering the minor code change.

easyaspi314:
Well Thumb-2 is weird.

The classic 16-bit Thumb ISA is a lot like x86:
 - it has 8 primary registers (the other 8 have limited access)
 - most instructions modify the source operand (a.k.a. `+=`, `-=`, etc)
 - everything updates the condition flags
 - branch-only, no conditional execution
 - no barrel shifter

The major difference is that you can't work directly in memory like x86 (meaning it spends a lot of time frantically swapping between memory and the Hi and Lo registers), and it has a *lot* fewer instructions (e.g. `umull`).

It was basically designed to be used on stuff like the GBA which had a 16-bit memory bus, and isn't really good for general use (and is a pain in the ass to read and write). It can barely run XXH32.

Thumb-2 is an extension which makes it a lot more viable for general purpose programming by giving you the 16-bit instructions, but also allowing access to almost all 32-bit instructions, with only a few restrictions compared to the full ARM ISA.

However, the density depends on how much of the 16-bit instructions you use.

For example:

```asm
        movs    r2, #128 @ 16-bit, Lo register
        movs    r8, #128 @ 32-bit, Hi register
        adds    r4, r4, r5 @ 16-bit, dest==src
        adds    r6, r4, r5 @ 32-bit, different dest
        subs    r4, r4, r5 @ 16-bit, updates flags
        sub     r4, r4, r5 @ 32-bit, doesn't update flags
        push    {r4-r7, lr} @ 16-bit, all Lo or lr
        push    {r4-r7, r11, or} @ 32-bit
```

What is likely happening is that Clang changed its register allocation, resulting in using more Lo registers, meaning smaller code.

Cyan4973:
FYI : there is now `#308` tested as a possible variant of this PR.
But note that #307 is still in the contender list.
It's actually quite good, offers better speed on the [4-7] segment, and is still decent for `len==8`, especially on `x64`, which is an important target.
So that's also a matter of how much we value non-`x64` targets, such as `x86` and `ARM`, where `len==8` suffers more.

Cyan4973:
Superseded by #308

PR #308, opened 2020-02-20
Last variant for the 4to8 segment (mul32to64)

Cyan4973:
This is an iteration over #307 .

The motive is that #307 give up mixing bits at `len==8`, on the ground that it cannot completely recover from space contraction, and therefore delegate that length to the now extended `8to16` segment. The issue is that the performance for `len==8` is now "bad" (or "less good") on 32-bit and ARM platform.
However, the 4to7 mixer, based on `mul32to64`, was "not so far" on quality even on 64-bit input, passing all tests bar one, and by a relatively small margin. So this variant tries to fix the remaining issue, located in the extended `DiffDist` test dealing with partial extraction of bitfields.

It improves the small bias in the lower bits by using a more complex insertion policy, in order to feed `input` directly in the upper region, before the avalanche.

It works, improving the collision rate of partial bitfields. It's not perfect, but is a measurable improvement, and make the collision ratio fall below the 2x threshold.

The cost though is a more complex algorithm, which is paid in speed and latency. It's therefore slower than #307 , but keeps most of its properties, preserving generally good performance on 32-bit and ARM (in contrast with #306 (closed)).

The real competitor here is `smallInputs` (#305), which already improved speed noticeably compared to `v0.7.2` , and has the advantage to offer bijectivity for 64-bit inputs. So this patch has tough foe to beat.

Latency test, fixed size inputs [4-8] , performance reported in Millions of Hashes per Second (MH/s)

| arch | cpu | compiler | v0.7.2 | [smallInputs](https://github.com/Cyan4973/xxHash/pull/305) | this PR | #307 [4-7] | #307 [8] |
| --- | --- | --- | --- | --- | --- | --- | --- |
| `x64` | i7-8559U | Apple clang 11.0 |  135 | 159  | 165  | 182  | 170  |
| `x64` | i7-9700K | gcc v9.2.1 | 114 | 125 | 135  | 144  | 138  |
| `x86` | i7-9700K | gcc v9.2.1 | 87 | 98  | 104  | 115  | 70  |
| `aarch64` | snapdragon 855 | clang v9.0.1 | 92 | 105  |  108 | 120 | 99 |

So yes, it's _slightly_ better than `smallInputs`. Is that enough though ?....

Also : note how well #307 behaves on `x64`. It looks actually pretty good, up to the point where it's tested on `x86` 32-bit and `ARM`, then `len==8` takes a serious hit.

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -574,16 +574,17 @@ XXH3_len_4to8_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_h
     XXH_ASSERT(input != NULL);
     XXH_ASSERT(secret != NULL);
     XXH_ASSERT(4 <= len && len < 8);
-    seed = XXH_rotl64(seed, 13);
+    //seed = XXH_rotl64(seed, 13);
     {   xxh_u32 const input1 = XXH_readLE32(input);
         xxh_u32 const input2 = XXH_readLE32(input + len - 4);
         xxh_u32 const key1 = input1 ^ ((xxh_u32)(seed >> 32) + XXH_readLE32(secret));
         xxh_u32 const key2 = input2 ^ (XXH_readLE32(secret+4) - (xxh_u32)seed);
         xxh_u64 const mix = XXH_mult32to64(key1, key2)
-                          + input1
-                          + input2
+                          + (XXH_swap64(input1) >> 12)
+                          + (XXH_swap64(input2) >> 13)
                           + len;
-        return XXH3_avalanche(mix);
+        //return XXH3_avalanche(mix);
+        return XXH3_avalanche(mix ^ (mix >> 47));
     }
 }
 
@@ -604,7 +605,7 @@ XXH_FORCE_INLINE XXH64_hash_t
 XXH3_len_0to16_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_hash_t seed)
 {
     XXH_ASSERT(len <= 16);
-    {   if (XXH_likely(len >= 8)) return XXH3_len_9to16_64b(input, len, secret, seed);
+    {   if (XXH_likely(len >  8)) return XXH3_len_9to16_64b(input, len, secret, seed);
         if (XXH_likely(len >= 4)) return XXH3_len_4to8_64b(input, len, secret, seed);
         if (len) return XXH3_len_1to3_64b(input, len, secret, seed);
         return XXH3_avalanche((PRIME64_1 + seed) ^ XXH_readLE64(secret));
--- a/xxh3.h
+++ b/xxh3.h
@@ -576,10 +576,9 @@ XXH3_len_4to8_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_h
     XXH_ASSERT(4 <= len && len < 8);
     {   xxh_u32 const input_1 = XXH_readLE32(input);
         xxh_u32 const input_2 = XXH_readLE32(input + len - 4);
-        xxh_u64 const input_64 = input_2 | ((xxh_u64)input_1 << 32);
-        xxh_u64 const keyed = input_64 ^ (seed + XXH_readLE64(secret));
-        xxh_u64 const mix64 = len + ((keyed ^ (keyed >> 39)) * PRIME64_2);
-        return XXH3_avalanche(mix64 ^ (mix64 >> 47));
+        xxh_u32 const key1 = input_1 ^ ((seed >> 32) + XXH_readLE32(secret));
+        xxh_u32 const key2 = input_2 ^ (seed + XXH_readLE32(secret+4));
+        return XXH3_avalanche(XXH_mult32to64(key1, key2) + len);
     }
 }
 
--- a/xxh3.h
+++ b/xxh3.h
@@ -574,12 +574,13 @@ XXH3_len_4to8_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_h
     XXH_ASSERT(input != NULL);
     XXH_ASSERT(secret != NULL);
     XXH_ASSERT(4 <= len && len < 8);
+    seed ^= XXH_swap64(seed);
     {   xxh_u32 const input1 = XXH_readLE32(input);
         xxh_u32 const input2 = XXH_readLE32(input + len - 4);
-        xxh_u32 const key1 = input1 ^ ((xxh_u32)(seed >> 32) + XXH_readLE32(secret));
+        xxh_u32 const key1 = XXH_swap32(input1) ^ ((xxh_u32)(seed >> 32) + XXH_readLE32(secret));
         xxh_u32 const key2 = input2 ^ (XXH_readLE32(secret+4) - (xxh_u32)seed);
         xxh_u64 const mix = XXH_mult32to64(key1, key2)
-                          + XXH_swap64(input1)
+                          + ((xxh_u64)input1 << 32)
                           + ((xxh_u64)(XXH_rotl32(input2,23)) << 32)
                           + len;
         return XXH3_avalanche(mix ^ (mix >> 59));
--- a/xxh3.h
+++ b/xxh3.h
@@ -574,17 +574,15 @@ XXH3_len_4to8_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_h
     XXH_ASSERT(input != NULL);
     XXH_ASSERT(secret != NULL);
     XXH_ASSERT(4 <= len && len < 8);
-    //seed = XXH_rotl64(seed, 13);
     {   xxh_u32 const input1 = XXH_readLE32(input);
         xxh_u32 const input2 = XXH_readLE32(input + len - 4);
         xxh_u32 const key1 = input1 ^ ((xxh_u32)(seed >> 32) + XXH_readLE32(secret));
         xxh_u32 const key2 = input2 ^ (XXH_readLE32(secret+4) - (xxh_u32)seed);
         xxh_u64 const mix = XXH_mult32to64(key1, key2)
-                          + (XXH_swap64(input1) >> 12)
-                          + (XXH_swap64(input2) >> 13)
+                          + (XXH_swap64(input1) >> 0)
+                          + (XXH_swap64(input2) >> 1)
                           + len;
-        //return XXH3_avalanche(mix);
-        return XXH3_avalanche(mix ^ (mix >> 47));
+        return XXH3_avalanche(mix ^ (mix >> 59));
     }
 }
 
--- a/xxh3.h
+++ b/xxh3.h
@@ -574,11 +574,11 @@ XXH3_len_4to8_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_h
     XXH_ASSERT(input != NULL);
     XXH_ASSERT(secret != NULL);
     XXH_ASSERT(4 <= len && len < 8);
-    {   xxh_u32 const input_1 = XXH_readLE32(input);
-        xxh_u32 const input_2 = XXH_readLE32(input + len - 4);
-        xxh_u32 const key1 = input_1 ^ ((seed >> 32) + XXH_readLE32(secret));
-        xxh_u32 const key2 = input_2 ^ (seed + XXH_readLE32(secret+4));
-        return XXH3_avalanche(XXH_mult32to64(key1, key2) + len);
+    {   xxh_u32 const input1 = XXH_readLE32(input);
+        xxh_u32 const input2 = XXH_readLE32(input + len - 4);
+        xxh_u32 const key1 = input1 ^ ((xxh_u32)(seed >> 32) + XXH_readLE32(secret));
+        xxh_u32 const key2 = input2 ^ ((xxh_u32)seed + XXH_readLE32(secret+4));
+        return XXH3_avalanche(XXH_mult32to64(key1, key2) + input1 + input2 + len);
     }
 }
 
--- a/xxh3.h
+++ b/xxh3.h
@@ -579,8 +579,8 @@ XXH3_len_4to8_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_h
         xxh_u32 const key1 = input1 ^ ((xxh_u32)(seed >> 32) + XXH_readLE32(secret));
         xxh_u32 const key2 = input2 ^ (XXH_readLE32(secret+4) - (xxh_u32)seed);
         xxh_u64 const mix = XXH_mult32to64(key1, key2)
-                          + (XXH_swap64(input1) >> 0)
-                          + (XXH_swap64(input2) >> 1)
+                          + XXH_swap64(input1)
+                          + ((xxh_u64)(XXH_rotl32(input2,23)) << 32)
                           + len;
         return XXH3_avalanche(mix ^ (mix >> 59));
     }
--- a/xxh3.h
+++ b/xxh3.h
@@ -574,11 +574,16 @@ XXH3_len_4to8_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_h
     XXH_ASSERT(input != NULL);
     XXH_ASSERT(secret != NULL);
     XXH_ASSERT(4 <= len && len <= 8);
+    seed = XXH_rotl64(seed, 13);
     {   xxh_u32 const input1 = XXH_readLE32(input);
         xxh_u32 const input2 = XXH_readLE32(input + len - 4);
         xxh_u32 const key1 = input1 ^ ((xxh_u32)(seed >> 32) + XXH_readLE32(secret));
-        xxh_u32 const key2 = input2 ^ ((xxh_u32)seed + XXH_readLE32(secret+4));
-        return XXH3_avalanche(XXH_mult32to64(key1, key2) + input1 + input2 + len);
+        xxh_u32 const key2 = input2 ^ (XXH_readLE32(secret+4) - (xxh_u32)seed);
+        xxh_u64 const mix = XXH_mult32to64(key1, key2)
+                          + input1
+                          + input2
+                          + len;
+        return XXH3_avalanche(mix);
     }
 }
 
--- a/xxh3.h
+++ b/xxh3.h
@@ -58,6 +58,16 @@
 #  define XXH_RESTRICT   /* disable */
 #endif
 
+#if (defined(__GNUC__) && (__GNUC__ >= 3))  \
+  || (defined(__INTEL_COMPILER) && (__INTEL_COMPILER >= 800)) \
+  || defined(__clang__)
+#    define XXH_likely(x) __builtin_expect(x, 1)
+#    define XXH_unlikely(x) __builtin_expect(x, 0)
+#else
+#    define XXH_likely(x) (x)
+#    define XXH_unlikely(x) (x)
+#endif
+
 #if defined(__GNUC__)
 #  if defined(__AVX2__)
 #    include <immintrin.h>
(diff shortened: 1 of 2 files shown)


Cyan4973:
Completing with a quality evaluation of the 3 contenders.
Only reporting the tests where there is a small difference, since all other ones are successful.

| Test | SmallInput | SmallFold32 | mul32Len8 | comment |
| Avalanche (<= 64-bit) | 0.83% (56-bit) | 0.73% (40-bit) | | worse bias |
| DiffDist | 1 | 1 | | nb of tests above threshold (>x2) |
| DiffDist | x2.94 (bitflip 5, high 37 bits) | x2.47 (bitflip 25, low 37 bits) | | worse collision ratio |
| PRNG | 1835 | 401 | | Lower is better |

easyaspi314:
> SmallFold32 would be the next logical choice, also better quality and faster, especially on x64, but unfortunately a bit slower for len==8 on x86 and ARM, so it's not a "complete" win.

You are comparing apples to oranges.

The len==8 is slower on 32-bit and aarch64 because it switches to the 9to16 algorithm.

The 9to16 algorithm is slower on 32-bit and aarch64 because 

1. 32-bit and aarch64 are not particularly great at 64->128-bit multiplies.
2. Additionally (heh), 32-bit has to deal with 5 adc/sbc pairs which linearize because they depend on the carry flag.

That slowdown you see would still occur at `len==9`, just now one byte earlier. 

On my phone, `9to16` runs at ~82% the speed in 64-bit (similar to your 855) and ~66% the speed in 32-bit, compared to `smallFold32`'s `4to7` routine.

Besides, even with the slower performance, it still competes with XXH32 on both my PC and my phone, clearing it completely in the average case.

I wouldn't personally let that slowdown be the deciding factor.

easyaspi314:
Also, performance is roughly the same as smallFold32 on ARMv6T2, and a tiny bit faster than smallInputs on aarch64.

Cyan4973:
> performance is roughly the same as smallFold32 on ARMv6T2, and a tiny bit faster than smallInputs on aarch64.

Performance of ? this branch (`mul32len8test`) ? at `len==8` ?

Cyan4973:
I'm updating this branch with a minor change that makes it pass the prng test.
It looks a much better contender now,
although I keep in mind that you seem to prefer the `SmallFold32` branch.

The learning could be useful though, to make the `SmallFold32` branch also pass the test.

easyaspi314:
>Performance of ? this branch (mul32len8test) ? at len==8 ?

In the avg. case compared to smallFold32.

> It looks a much better contender now, although I keep in mind that you seem to prefer the SmallFold32 branch

Yeah, I have to admit I kinda like SmallFold32 mainly for its speed and simplicity. 🤷‍♂️

This one is also good, although as you stated, the speedups are rather marginal compared to `smallInputs` (aside from ARM32 which has no issues).

Also, I'm having trouble understanding the table, especially the comments which unless I am misreading it, seem to contradict the data.

Cyan4973:
> I'm having trouble understanding the table, especially the comments 

Well, the "comment" column is actually more an explanation of what the test is about

Cyan4973:
I've got a modification of `smallFold32` which makes it better at the prng test.
It increases the complexity by a single additional instruction, which is enough to bring the prng score down to 8, which is still technically labelled a "failure", but is close enough if you ask me, and way better than current 401.

It negatively impacts speed of the 4-7 segment, by very little (`x86`: 116=>115, `aarch64`: 122=>120, `x64`: 153 => 144 (more perceptible)).
Only the 4-7 segment is impacted, because the prng test is only concerned by inputs of size 4, which is a curious choice, so it never tests size 8.

Considering that all other quality metrics are unaltered, I suspect it's a better choice, though it makes #307 now a bit closer in speed to #308.

easyaspi314:
Clang x64 seems to be making an excess `mov` before `bswap` on the new smallFold32. Haven't checked GCC yet.

Cyan4973:
> Clang x64 seems to be making an excess mov before bswap (as bswap modifies in place) on the new smallFold32. 

Interesting,
on the benchmark side, `clang` seems less impacted than `gcc` though...

Cyan4973:
So, basically, the choice is now between #307 and #308,
I believe it's clear at that point that they are both superior to #305 (`smallInputs`).

An associated question is about the test : the only "problematic" ones are DiffDist and prng. But it's questionable why DiffDist only works on 64-bit input, while prng only works on 32-bit input.

Maybe I should consider extending those tests, so that they also probe different input length. That would be a good validation.

easyaspi314:
On `ms_abi` (not `sysv_abi`), GCC x64 is *pushing to the stack* for some reason.

no bswap:
```asm
XXH3_len_4to7_64b:
        mov     r11d, DWORD PTR -4[rcx+rdx]
        mov     r10d, DWORD PTR [rcx]
        rol     r9, 13
        mov     ecx, DWORD PTR 4[r8]
        mov     rax, r9
        shr     rax, 32
        add     eax, DWORD PTR [r8]
        sub     ecx, r9d
        xor     eax, r10d
        add     r10, rdx
        xor     ecx, r11d
        add     r10, r11
        imul    rax, rcx
        add     rax, r10
        mov     rdx, rax
        shr     rdx, 37
        xor     rax, rdx
        movabs  rdx, 1609587791953885689
        imul    rax, rdx
        mov     rdx, rax
        shr     rdx, 32
        xor     rax, rdx
        ret
```

with bswap:
```asm
XXH3_len_4to7_64b:
        rol     r9, 13
        push    rbx      # <<<<< wtf
        mov     eax, DWORD PTR [r8]
        mov     r11d, DWORD PTR -4[rcx+rdx]
        mov     r10d, DWORD PTR [rcx]
        mov     rcx, r9
        shr     rcx, 32
        add     eax, ecx
        mov     ecx, DWORD PTR 4[r8]
        mov     ebx, r10d
        add     r10, rdx
        bswap   ebx
        xor     eax, ebx
        add     r10, r11
        pop     rbx      # <<<<< wtf
        sub     ecx, r9d
        xor     ecx, r11d
        imul    rax, rcx
        add     rax, r10
        mov     rdx, rax
        shr     rdx, 37
        xor     rax, rdx
        movabs  rdx, 1609587791953885689
        imul    rax, rdx
        mov     rdx, rax
        shr     rdx, 32
        xor     rax, rdx
        ret
```

Cyan4973:
Strange, `rbx` is not mentioned anywhere in the code snippet, which implies it's not modified ?
Could it be the implicit target of another instruction, like the newer `bswap` ?

easyaspi314:
In this case, it is using `rbx` as the 32-bit version, `ebx`.

easyaspi314:
I see no major difference on my PC; but that may be because Ryzen has a really fast `bswap`.

Cyan4973:
I made an extended version of `DiffDist` and `prng`, extending them to test both 32 and 64-bit inputs.

`DiffDist` doesn't bring anything new.

But `prng` proved a harder foe for `SmallFold32`, breaking it again, especially when trying to use the `seed` to generate the suite of random numbers.

`mul32len8`, on the other hand, survived the test quite well.

So it seems this hints towards this later variant.

Another alternative is to tweak `SmallFold32` so that it can survive the test too. But `mul32len8` being already a tweaked variant of `SmallFold32`, at which point will they become close to each other ?

Cyan4973:
I tried to improve `SmallFold32` using only light modifications to make it pass the `prng` test, but failed. Without "heavier" modifications, it remains pretty bad.
I also tried to simplify `mul32len8` to make it slightly faster, but it would invariably break something.
 
So, unless you have something up your sleeves @easyaspi314, I think I will validate this one and move on.

easyaspi314:
Ok. I'm taking a closer look

For one, the bswap fold truncates the seed into 32 bits. Are you aware of that?

For any `x`, given `swapped = x ^ swap64(x);`, `(swapped & 0xffffffff) == swap32(swapped >> 32)`
    
```c
#include <stdio.h>
#include <stdlib.h>
#include <stdint.h>

uint64_t swap64_fold_v1(uint64_t x)
{
    x ^= __builtin_bswap64(x);
    return x;
}
uint64_t swap64_fold_v2(uint64_t x)
{
    uint32_t low = __builtin_bswap32(x >> 32) ^ (uint32_t)x;
    // The high bits are the low bits reversed!!!
    uint32_t high = __builtin_bswap32(low);
    return low | ((uint64_t)high << 32);
}
```

Cyan4973:
> the bswap fold truncates the seed into 32 bits

Which one ? 
Do you mean this line in `mul32len8` : `seed ^= XXH_swap64(seed);` ?

easyaspi314:
Yes. Look (using half scale with nibble swaps)
```
0x12ac329f ^ nibbleswap32(0x12ac329f)
0x12ac329f ^ 0xf923ca21
0xeb8ff8be
nibbleswap16(0xeb8ff8be >> 32) == (0xeb8ff8be & 0xffff)
nibbleswap16(0xeb8f) == 0xf8be
0xf8be == 0xf8be
```

Cyan4973:
Oh, I get it ! Great catch !
Yes, you are totally right. This symmetry reduces `seed`'s space, which is indeed not good.

It's impressive that it nonetheless passed all tests. It means there is no test that tries to pressure the higher 32-bits of the `seed`. That's something that must be added in the test suite. 

And of course, the `seed` folding must be fixed too.

easyaspi314:
Example seed-based collision:
```c
#include <stdio.h>
#include <stdint.h>

#define XXH_INLINE_ALL
#include "xxhash.h"

int main(void)
{
    const char *str = "Tests";
    uint64_t seed1 = 0x123456789ABCDEF0;
    uint64_t seed2 = 0xF034BC789A56DE12;

    uint64_t hash1 = XXH3_64bits_withSeed(str, 5, seed1);
    uint64_t hash2 = XXH3_64bits_withSeed(str, 5, seed2);
    printf("%016llx %016llx\n", hash1, hash2);
}
```

easyaspi314:
> It means there is no test that tries to pressure the higher 32-bits of the seed. 

SMHasher/Hashes.h:274:
```c
inline void xxh3_test( const void * key, int len, uint32_t seed, void * out ) {
  // objsize 12d0-15b8: 744
  *(uint64_t*)out = (uint64_t) XXH3_64bits_withSeed(key, (size_t) len, seed);
}
```
> `uint32_t seed`

Well there's our problem.

Cyan4973:
Yep, that's indeed the problem.

I guess I should consider developing my own tools, but hey, all this cost time...

easyaspi314:
```c
uint64_t XXH3_64bits_withSeed(const void *input, size_t len, uint64_t seed)
{
    if (seed > 0x7FFFFFFF) {
        /* Optimization: With high seed values, return zero.
         * This makes certain inputs a lot faster while still
         * passing SMHasher */
        return 0;
    }
    ...
}
```

Cyan4973:
Yes, you are right, that would work,
just, having to manually create a variant for each hash function to test just for the purpose of testing the higher bits of `seed` (which only matters in few selected tests) feels somewhat hackish and unsatisfying.
But hey, that's a nice hack, could make a good stop-over.

easyaspi314:
Honestly, a waaaay better method would be to just do something like this.

```c
#include <stdio.h>
#include <stdint.h>

// XXH3_avalanche is a fast bijective scramble which
// can generate reasonably random seeds without any
// collisions.
static inline uint64_t XXH3_avalanche(uint64_t h64)
{
    h64 ^= h64 >> 37;
    h64 *= 0x165667919E3779F9U;
    h64 ^= h64 >> 32;
    return h64;
}

int main(void)
{
    for (uint64_t i = 0; i < 512; i++) {
        printf("%016llx\n", XXH3_avalanche(i));
    }
}
```

easyaspi314:
Dumb 32-bit iterative version
```
--- Testing Murmur3F "MurmurHash3 for x64, 128-bit" GOOD

[[[ Keyset 'Seed' Tests ]]]

Keyset 'Seed' - 5000000 keys
Testing collisions (128-bit) - Expected    0.0, actual      0 (0.00x)
Testing collisions (high 64-bit) - Expected          0.0, actual      0 (0.00x)
Testing collisions (high 32-bit) - Expected       5820.8, actual   2939 (0.50x)
Testing collisions (high 26-40 bits) - Worst is 32 bits: 2939/5820 (0.50x)
Testing collisions (high 12-bit) - Expected    5000000.0, actual 4995904 (1.00x) (-4096)
Testing collisions (high  8-bit) - Expected    5000000.0, actual 4999744 (1.00x) (-256)
Testing collisions (low  64-bit) - Expected          0.0, actual      0 (0.00x)
Testing collisions (low  32-bit) - Expected       5820.8, actual   2989 (0.51x)
Testing collisions (low  26-40 bits) - Worst is 40 bits: 15/22 (0.66x)
Testing collisions (low  12-bit) - Expected    5000000.0, actual 4995904 (1.00x) (-4096)
Testing collisions (low   8-bit) - Expected    5000000.0, actual 4999744 (1.00x) (-256)
Testing distribution - Worst bias is the 19-bit window at bit 24 - 0.049%



Input vcode 0x00000001, Output vcode 0x00000001, Result vcode 0x00000001
Verification value is 0x00000001 - Testing took 30.155447 seconds
```

Full 64-bit pseudorandom seed
(Also, LOL @ divide by zero)
```
--- Testing Murmur3F "MurmurHash3 for x64, 128-bit" GOOD

[[[ Keyset 'Seed' Tests ]]]

Keyset 'Seed' - 5000000 keys
Testing collisions (128-bit) - Expected    0.0, actual   3598 (48973448041951073555854655488.00x) !!!!!
Testing collisions (high 64-bit) - Expected          0.0, actual   3598 (2654855938.06x) (3598) !!!!!
Testing collisions (high 32-bit) - Expected       5820.8, actual   6529 (1.12x) (709)
Testing collisions (high 26-40 bits) - Worst is 40 bits: 3607/22 (158.64x) !!!!!
Testing collisions (high 12-bit) - Expected    5000000.0, actual 4995904 (1.00x) (-4096)
Testing collisions (high  8-bit) - Expected    5000000.0, actual 4999744 (1.00x) (-256)
Testing collisions (low  64-bit) - Expected          0.0, actual   3598 (2654855938.06x) (3598) !!!!!
Testing collisions (low  32-bit) - Expected       5820.8, actual   6530 (1.12x) (710)
Testing collisions (low  26-40 bits) - Worst is 40 bits: 3611/22 (158.81x) !!!!!
Testing collisions (low  12-bit) - Expected    5000000.0, actual 4995904 (1.00x) (-4096)
Testing collisions (low   8-bit) - Expected    5000000.0, actual 4999744 (1.00x) (-256)
Testing distribution - Worst bias is the 19-bit window at bit 64 - 0.058%

*********FAIL*********


Input vcode 0x00000001, Output vcode 0x00000001, Result vcode 0x00000001
Verification value is 0x00000001 - Testing took 30.495486 seconds
```

Cyan4973:
Fixed the space reduction of `seed`, with a transform producing an equivalent outcome, but which is bijective.
I tried a simpler one too, `^= seed << 32`, but it fails the `prng` test.

easyaspi314:
LGTM. Curious, what is the new prng score? 

This also makes it so you don't need a fast byteswap for the 64bits version. 

Although I am sad that we don't get this output on byteshift+no bswap:
```c
typedef struct {
    uint32_t little_endian;
    uint32_t big_endian;
} bswap_pair;

bswap_pair read32_and_swap(const uint8_t *ptr)
{
    /* ptr = { 1, 2, 3, 4 } */
    /* 1030 */
    uint32_t even_le = ptr[0] | ((uint32_t)ptr[2] << 16);
    /* 4020 */
    uint32_t odd_be = ptr[3] | ((uint32_t)ptr[1] << 16);
    /* 0301 */
    uint32_t even_be = rotl32(even_le, 24);
    /* 0204 */
    uint32_t odd_le = rotl32(odd_be, 24);
    /* 4321 */
    uint32_t be = even_be | odd_be;
    /* 1234 */
    uint32_t le = even_le | odd_le;
    return (bswap_pair) { le, be };
}
```
On ARMv4T (no unaligned, no bswap),  ignoring the returned structs are in pointers:
```asm
@ ptr: r0
@ returns in {r0,r1}
read32_and_swap:
        @ r1 = ptr[3]
        ldrb    r1, [r0, #3]
        @ r2 = ptr[0]
        ldrb    r2, [r0]
        @ r3 = ptr[1]
        ldrb    r3, [r0, #1]
        @ r0 = ptr[2]
        ldrb    r0, [r0, #2]
        @ r1 |= r3 << 16
        orr     r1, r1, r3, lsl #16
        @ r2 |= r0 << 16
        orr     r2, r2, r0, lsl #16
        @ r0 = r2 | rotl(r1, 24)
        orr     r0, r2, r1, ror #8
        @ r1 = r1 | rotl(r2, 24)
        orr     r1, r1, r2, ror #8
        @ return {r0, r1}
        bx       lr
```
Clang unfortunately emits garbage with `__builtin_rotate` as it turns it into a funnel shift. Only with an asm guard does it work properly.

Cyan4973:
> Curious, what is the new prng score ?

There are 2 scores now, one is for 32-bits inputs, and one is for 64-bits inputs.
32-bits inputs is 2.66, and 64-bits inputs is 0.02  (which is good, since lower is better).
The original code considers anything < 3.88 to be a "success". no explanation as to why this threshold.

I wouldn't pay attention to the exact notation though, since I believe a lot of details in this test are rubbish, but the order of magnitude is interesting, and scoring in the thousands ands more tells something about using the hash function as a random number generator.
It tells that if someone was to generate a list pf random number from the hash function _by using sequential inputs (1,2,3,etc.)_, using either the input or the seed to insert the sequence, it would receive a reasonably good list of random numbers, analyzing not just the sequence of produced hash values, but the difference between 2 generated numbers (the derivative if you wish). It is designed to punish linear correlations. This property is actually detrimental for hash tables, since a linear correlation using an odd number would produce some flatter, hence more efficient, distribution pattern. But prng people tend to find this property important to consider a source "really" random. So here it is.
Also, I always thought that random number generators were supposed to use the previous output value as the entry for next one, and this is not what this test does.

easyaspi314:
> Also, I always thought that random number generators were supposed to use the previous output value as the entry for next one, and this is not what this test does.

Yes, that is how even the most basic garbage ones are implemented, such as in the ISO C standard:

```c
static unsigned long next = 1;
int myrand(void)  /* RAND_MAX assumed to be 32767. */
{
    next = next * 1103515245 + 12345;
    return((unsigned)(next/65536) % 32768);
}


void mysrand(unsigned seed)
{
    next = seed;
}
```

This test is 100% nonsense.

peteroupc:
What is this "PRNG test" you're referring to?  I see nothing like that in the original SMHasher or its forks.  Is this the same as the "Seed test"?

Cyan4973:
@peteroupc : it's not part of the original SMHasher test list. It was adder later to one fork, under the name "MomentChi2". And I call it prng test, because I believe the origin of this test was for checking prng properties, but what the source code of "MomentChi2" is doing is only partially following the original paper, adding or replacing several questionable details, and even inserting a backdoor to ensure that hash algorithms with symmetric input / seed necessarily pass it. And as you can guess, the author of the "test" is also the author of a hash algorithm  with symmetric input / seed...

easyaspi314:
> And I call it prng test, because I believe the origin of this test was for checking prng properties

***Well how the hell were we supposed to know that?*** :joy:

Here I was going on a wild goose chase checking every fork and branch of SMHasher to try and find this mysterious prng test, and it was just a headcanon nickname.

easyaspi314:
Also, is this the 'backdoor' you are talking about, where `(sa == sb) == 0`?

```c
double chi2=(sa-sb)*(sa-sb)/(saa+sbb);
```
https://github.com/rurban/smhasher/issues/65

Cyan4973:
Yes, that's the right thread.

The backdoor is in the process itself : it compares the output of a series of linear values sent to the input with sending the exact same series to the `seed`, and then compares both. If they feature the same statistics, then the test is successful.

Obviously, a hash algorithm with perfectly symmetric seed and input will pass this test with flying colors ! The hash algorithm of the test's author gets a perfect "0" score. But so does `donothing`, a fake hash algorithm which always return the same value (it's symmetric too ! since it ignores both seed & input ...).

That being said, a perfect random number generator will _also_ pass this test, although not with a "0" score. And that's why I keep an eye on it.

A real test would have compared vs a perfect gaussian distribution, which can be mathematically determined. When doing the test, I'm effectively checking that both the serie from the input and from the `seed` match the mathematical "perfect" distribution closely enough. And in my fork, I can vary the input size too.

Cyan4973:
OK, merging this variant, and moving to `XXH128`.
We can always open a new PR if something shows up.

PR #311, opened 2020-02-25
Simplify len [4,8] for xxh128

Cyan4973:
This new version is much simpler than `v0.7.2` for len [4,8].
It replaces most of the multiplications by a 64x64->128 one, which feels adequate for a 128-bit hash.
As a consequence, it runs faster on `x64` systems. Even `aarch64` prefers it.

Note however that speed on 32-bit `x86` is negatively impacted.
In the context of `xxh128`, I'm not sure though if this is a relevant platform to optimize for.

Latency test, fixed size inputs [4-8] , performance reported in Millions of Hashes per Second (MH/s) :

| arch | cpu | compiler | v0.7.2 | this PR | 
| `x64` | i7-8559U | Apple clang 11.0 |  130 |  161 | 
| `x64` | i7-9700K | gcc v9.2.1 | 114 | 137 | 
| `x86` | i7-9700K | gcc v9.2.1 | 83 | 73 | 
| `aarch64` | snapdragon 855 | clang v9.0.1 | 89 | 94 |  
| `armv7ve` | snapdragon 855 | clang v9.0.1 | 25 | 71 |
| `armv6t2` | snapdragon 855 | clang v9.0.1 | 62 | 72 |

This new variant features 2 minor `DiffDist` warnings, at `x2.15` and `x2.33` (limit is `x2.0`) at bits `8` and `9`.

And last issue is the usual one : the `prng` test doesn't pass, although the score is not that bad (100 with 32-bit inputs, 40 with 64-bit inputs).

Fixing this last one requires an additional rotation in the critical path, slowing down the hash noticeably.
Haven't arbitrated yet if that's worthwhile ...

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -1437,7 +1437,7 @@ XXH3_len_4to8_128b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_
         xxh_u64 const input_64 = input_hi + ((xxh_u64)input_lo << 32);
         xxh_u64 const keyed = input_64 ^ (XXH_readLE64(secret) + seed);
 
-        XXH128_hash_t m128 = XXH_mult64to128(keyed, PRIME64_1 + len);
+        XXH128_hash_t m128 = XXH_mult64to128(keyed, PRIME64_1 + (len << 2));
 
         // alternative more "classical" insertion of len
         // The one above is slightly faster (135->138)
--- a/xxh3.h
+++ b/xxh3.h
@@ -1439,14 +1439,8 @@ XXH3_len_4to8_128b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_
 
         XXH128_hash_t m128 = XXH_mult64to128(keyed, PRIME64_1 + (len << 2));
 
-        // alternative more "classical" insertion of len
-        // The one above is slightly faster (135->138)
-        //XXH128_hash_t m128 = XXH_mult64to128(keyed, PRIME64_1);
-        //m128.low64  += len;
-        //m128.high64 -= len;
-
-        m128.low64  ^= (m128.high64 >> 5);
-        m128.high64 += (m128.low64 << 3);
+        m128.high64 += (m128.low64 << 1);
+        m128.low64  ^= (m128.high64 >> 3);
 
         m128.low64   = XXH3_avalanche(m128.low64);
         m128.high64  = XXH3_avalanche(m128.high64);
--- a/xxh3.h
+++ b/xxh3.h
@@ -1408,8 +1408,8 @@ XXH3_len_1to3_128b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_
     {   xxh_u8 const c1 = input[0];
         xxh_u8 const c2 = input[len >> 1];
         xxh_u8 const c3 = input[len - 1];
-        xxh_u32  const combinedl = ((xxh_u32)c1) + (((xxh_u32)c2) << 8) + (((xxh_u32)c3) << 16) + (((xxh_u32)len) << 24);
-        xxh_u32  const combinedh = XXH_swap32(combinedl);
+        xxh_u32  const combinedl = ((xxh_u32)c1<<16) | (((xxh_u32)c2) << 24) | (((xxh_u32)c3) << 0) | (((xxh_u32)len) << 8);
+        xxh_u32  const combinedh = XXH_rotl32(XXH_swap32(combinedl), 13);
         xxh_u64  const keyed_lo = (xxh_u64)combinedl ^ (XXH_readLE32(secret)   + seed);
         xxh_u64  const keyed_hi = (xxh_u64)combinedh ^ (XXH_readLE32(secret+4) - seed);
         xxh_u64  const mixedl = keyed_lo * PRIME64_1;
--- a/xxh3.h
+++ b/xxh3.h
@@ -570,11 +570,17 @@ XXH3_mul128_fold64(xxh_u64 lhs, xxh_u64 rhs)
 }
 
 
+XXH_FORCE_INLINE xxh_u64 XXH_xorshift64(xxh_u64 v64, int shift)
+{
+    XXH_ASSERT(0 <= shift && shift < 64);
+    return v64 ^ (v64 >> shift);
+}
+
 static XXH64_hash_t XXH3_avalanche(xxh_u64 h64)
 {
-    h64 ^= h64 >> 37;
+    h64 = XXH_xorshift64(h64, 37);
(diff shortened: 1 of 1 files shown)


easyaspi314:
> (100 with 32-bit inputs, 40 with 32-bits inputs).

Typo?

> `prng` test

It is officially called `MomentChi2`...

The code is a lot faster and easier to implement. 

Additionally, even ARM32 enjoys a speedup here as well, going from 189 MB/s (for `./xxhsum -b12 -B5`) to 250 MB/s, and not generating inferior code on ARMv7-a.

easyaspi314:
By the way, if you use Termux, the way you can make 32-bit ARM binaries work is:

1. `pkg install ndk-multilib`
2. `mv $PREFIX/aarch64-linux-android $PREFIX/aarch64-linux-android.bak` (fixes link issues on Android 10)

When compiling, do `clang -m32 -march=armv7ve -static` (static is needed because otherwise it collides with the magic LD_PRELOAD which makes shell script shebangs work)

example:
```
$ make xxhsum CC="clang -m32 -march=armv7ve -static" (don't build libxxhash)
$ ./xxhsum -b
./xxhsum 0.7.3 (32-bits arm little endian), Clang 9.0.1 , by Yann Collet
```
As for `benchHash`:

```
$ make CC="clang -m32 -march=armv7ve -static" CXX='$(CC)'
```

Cyan4973:
Thanks for the hint @easyaspi314 !
Indeed, it's pretty useful to generate 32-bit ARM binaries and measure their performance !

I was surprised to notice that `armv6t2` seems to produce better performance than `armv7ve` (mostly on `v0.7.2`).

easyaspi314:
That is mostly due to LLVM autovectorizing the short hashes and emitting the nonexistent `vmulq_u64` intrinsic.

LLVM doesn't know how to emulate that intrinsic, so it generates (hilarious) garbage code:

Code (actually written in LLVM IR)
```c
uint64x2_t mul(uint64x2_t a, uint64x2_t b)
{
    return "vmulq_u32"(a, b);
}
```

`clang -O3 -march=armv7ve`
```asm
mul:
        push    {r4, r5, r6, r7, r11, lr}
        vmov    d17, r2, r3
        add     r12, sp, #24
        vmov    d16, r0, r1
        vld1.64 {d18, d19}, [r12]
        vmov.32 r0, d16[0]
        vmov.32 r1, d18[0]
        vmov.32 r12, d16[1]
        vmov.32 r3, d17[0]
        vmov.32 r2, d19[0]
        vmov.32 lr, d17[1]
        vmov.32 r6, d19[1]
        umull   r7, r5, r1, r0
        mla     r1, r1, r12, r5
        umull   r5, r4, r2, r3
        mla     r2, r2, lr, r4
        vmov.32 r4, d18[1]
        mla     r2, r6, r3, r2
        vmov.32 d17[0], r5
        vmov.32 d16[0], r7
        vmov.32 d17[1], r2
        mla     r0, r4, r0, r1
        vmov.32 d16[1], r0
        vmov    r2, r3, d17
        vmov    r0, r1, d16
        pop     {r4, r5, r6, r7, r11, pc}
```

Basically, it is moving everything into a vector (since `softfp` convention doesn't pass NEON registers, `a` is in `r0-r3` and `b` is on the stack), immediately extracts, does two scalar `uint64_t` multiplies (`umull, mla, mla`), puts them back in, and then extracts *again* to put the result in `r0-r3`.

Meanwhile, `armv6t2` has no SIMD, so it generates scalar code which is the better option.

I am thinking we should change `-O3` to  `-O3 -fno-tree-vectorize -fno-tree-slp-vectorize`.

easyaspi314:
Meanwhile, on ARMv6T2, Clang emits clean scalar code:
```asm
mul:
        push    {r4, lr}
        ldr     lr, [sp, #8]
        umull   r12, r4, lr, r0
        mla     r1, lr, r1, r4
        ldr     r4, [sp, #12]
        mla     r1, r4, r0, r1
        ldr     r0, [sp, #16]
        umull   lr, r4, r0, r2
        mla     r0, r0, r3, r4
        ldr     r3, [sp, #20]
        mla     r3, r3, r2, r0
        mov     r0, r12
        mov     r2, lr
        pop     {r4, pc}
```

easyaspi314:
Interesting, this change seems to cause Clang to autovectorize 1to3 on NEON.

Definitely starting to think that turning off autovec improves things overall.

Cyan4973:
Can we prevent autovec on a per-function basis, aka in source code rather than build script ?

easyaspi314:
AFAIK, not on Clang.
GCC has `#pragma GCC optimize("-fno-tree-slp-vectorize")`, but Clang only supports turning optimizations on and off. 

It also doesn't respect `__attribute__((__target__))`.

There seems to be too many points (which vary on the arch) to just put asm barriers.

Cyan4973:
> It also doesn't respect `__attribute__((__target__))`.

Note sure how applicable that is,
but in `zstd`, we are using
`__attribute__((__target__(target)))`
in order to selectively enable or disable `bmi2` instruction set :
https://github.com/facebook/zstd/blob/master/lib/common/compiler.h#L87

And it seems to work with `clang`.

Cyan4973:
Latest variant uploaded into `xxh128` branch seems to pass the remaining difficult tests that were not yet completely satisfying (`DiffDist` and `MC2`, or at least my fixed version of this test).

Currently running the extended quality tests to ensure nothing else show up.

Cyan4973:
OK, so the very very last detail is
the `MC2` test, when trying to generate a list of random numbers by providing a list of linearly increasing numbers from the `seed` (as opposed to ingestion from `input`, which works fine).

Fixing this could be achieved by avalanching the `seed` before ingestion.
But now, it introduces a more sensible speed difference between the "seed-less" and the "with-seed" variants, because the "seed-less" variant does not pay any cost for `seed` transformations. So would it be "cheating" ?

Moreover, the scenario measured seems quite a bit special, to say it mildly. Using a linear suite of numbers to generate random numbers and ingesting them from the `seed` rather than the `input` ? Where on Earth does such a thing ever happen ?

And finally, it's not that the current results are "really bad". They are merely "not perfect".

Except if there is some comments on this, I'm inclined to let it live this way.
In which case, this PR becomes a good candidate to merge.

easyaspi314:
> Note sure how applicable that is,
but in zstd, we are using __attribute__((__target__(target)))
in order to selectively enable or disable bmi2 instruction set :
> https://github.com/facebook/zstd/blob/master/lib/common/compiler.h#L87
> 
> And it seems to work with clang

Clang is a lot more picky than GCC in what it allows you to enable or disable.

It's pretty dumb.

> So would it be "cheating" ?

Is cheating on a test that itself is cheating even cheating? :smirk:

PR #312, opened 2020-02-27
Disable autovec, force inline hashLong

easyaspi314:
Autovectorization does more harm than good, especially on Clang+ARMv7 NEON where it assumes uint64x2_t multiplies are cheap, but in reality it scalarizes the entire operation.

XXH3_hashLong has also been FORCE_INLINEd - seems to benefit performance, and with the case of autovectorization, our handwritten intrinsics are better than anything GCC or Clang can emit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ LIBVER_MINOR := $(shell echo $(LIBVER_MINOR_SCRIPT))
 LIBVER_PATCH := $(shell echo $(LIBVER_PATCH_SCRIPT))
 LIBVER := $(LIBVER_MAJOR).$(LIBVER_MINOR).$(LIBVER_PATCH)
 
-# Shut off autovectorization, it does more harm than goodm
+# Shut off autovectorization, it does more harm than good.
 #
 # We write SIMD when we want SIMD and scalar code when we don't.
 # On 32-bit targets, Clang tends to vectorize 64-bit mutiplies.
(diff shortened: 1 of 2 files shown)


Cyan4973:
The only issue I see with this patch is that it is build-system based,
meaning it's applicable to `xxhsum` or `libxxhash`,
but does not apply when the `xxhash` source code is directly integrated "as is" into a 3rd party project,
and is the primary integration experience expected.

Cyan4973:
For information :
I tried comparing speed of both version `dev` and this PR on small data
with an Android smartphone with `termux`, selecting target `-m32 -march=armv7ve` ,
and I could not measure any difference.

Is that expected ?

easyaspi314:
`doubleSecret` may have had the side effect of breaking up the patterns which caused autovectorization.

`len==240` seems to be pretty noticable, going from 1.1 to 1.5 GB/s.

Cyan4973:
Impressive and unexpected side effect ...

Is this PR still interesting valuable then ?

easyaspi314:
Not sure anymore. I think an asm guard at mult64to128 will solve the main issues (also GCC x86?).

I will check before closing.

Cyan4973:
Well, I guess this is the last opened topic before a release.
Do you believe it's worth investigating ?

easyaspi314:
Ok, I found something decent.

Will open a new PR.

1. `#pragma clang loop vectorize(disable)` *actually works* on the 129to240_64b loop on NEON32, due to it being categorized as a normal vectorization instead of an SLP vectorization (e.g. XXH32)
2. I can fool GCC x86's SSE2 cost model by forcing `seed64` into a register in the top of `XXH3_mix16B`. This produces scalar code, but it still is much slower than Clang (3.3GB/s vs 4.6 GB/s) and would need further investigation. I would save it for a later patch, though.

After this, I have the second typo patch (already opened) and a quickfix for Unicode on Windows and I think we are fit for a release.

PR #314, opened 2020-02-27
xxh128 len[9-16] : small simplification

Cyan4973:
Insert `secret` from `input_hi` only, to avoid `xor` cancellation,
as suggested by @easyaspi314 .

Also, simplified a bit the next instructions,
resulting in a small improvement to distribution qualities.

Surprisingly, while I tested variants simplifying even more,
replacing the following larger multiplications by 2 smaller `mult32to64` ones,
or even a single one with some additional `xorshift`,
it would not result in any measurable speed benefit.

So I settled on this variant.
Long tests ongoing. No surprise expected, this variant already pass the most difficult ones.

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -1473,13 +1473,27 @@ XXH3_len_9to16_128b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64
     {   xxh_u64 const input_lo = XXH_readLE64(input) ^ XXH_readLE64(secret);
         xxh_u64 const input_hi = XXH_readLE64(input + len - 8) ^ (XXH_readLE64(secret+8) - seed);
         XXH128_hash_t m128 = XXH_mult64to128(input_lo ^ input_hi, PRIME64_1);
+        m128.low64  += ((len-1) << 54);
+        m128.high64 += input_hi + XXH_mult32to64(input_hi, PRIME32_2 - 1);
+
+        m128.low64  ^= m128.high64 >> 3;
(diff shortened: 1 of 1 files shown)


Cyan4973:
And surprise there is, as the 96-bit key test fails (it's 12 bytes, that's right in this range).

The low 64-bit show more collisions than it should. I suspect it might be missing some contribution.

easyaspi314:
Damn it, just as soon as I write a nice comment that shows the 128-bit arithmetic, you have to change it to something that isn't true 128-bit arithmetic. 😛

Looks interesting though.

Cyan4973:
This algorithm is supposed to be bijective (for 128-bit input).

Cyan4973:
Fixed remaining quality issues with regards to `len==12`.
It proved surprisingly more difficult than expected : like a too small blanket, each attempt at fixing one flaw was uncovering creating a different one.

Anyway, this version has passed all tests, including most thorough ones, it should be good to go.

easyaspi314:
> like a too small blanket, pulling on one side uncover another side, and each attempt at fixing one flaw was creating some different one.

I know *that* feeling.

Normally I would be a little skeptical about the even multiply, but since it isn't the sole source of mixing, it should be fine.

Cyan4973:
It's only even in disguise. Note the `+` just before. When you add `1` to an even number, it becomes ....

easyaspi314:
Took me a minute. Cheeky.

PR #315, opened 2020-02-28
Comments + 32-bit path for 9to16_128b, improve arch detection

easyaspi314:
Just some minor fixups.

Arch detection:
- MSVC x86 no longer shows up as `unknown`
- x86_64 and aarch64 always show `SSE2` and `NEON` respectively
- ARM gives more detail:
   - unaligned access
   - arch version (only number though)
   - thumb/thumb-2
   - example:
      - `ARMv6 + unaligned`
      - `ARMv4 + Thumb-1`
      - would previously show up as `arm` despite being entirely different

`9to16_128b`:
- Comments updated, switched to line-by-line because the whole thing isn't 128-bit arithmetic
- The `input_hi + XXH_mult32to64((xxh_u32)input_hi, PRIME32_2 - 1)` line is documented in detail
  - shows the usage of distributive property etc
- Add a second 32-bit path for said multiply
  - Saves a data dependency and an `adc` on 32-bit
  - Adds a mask on 64-bit
  - More reflective of the actual arithmetic

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -1615,24 +1615,63 @@ XXH3_len_9to16_128b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64
     XXH_ASSERT(input != NULL);
     XXH_ASSERT(secret != NULL);
     XXH_ASSERT(9 <= len && len <= 16);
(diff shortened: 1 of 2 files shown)


Cyan4973:
Great work @easyaspi314 !

PR #320, opened 2020-03-03
updated len [4-8] for xxh3

Cyan4973:
one more time ...

This is (hopefully) the last change for this release.

The recent brute-force collision tester found a weakness right in this range.

The new version is inspired by rrmxmx, and passes all tests with flying colors.
Btw, it also restores bijectivity for 64-bit inputs.

It's also a bit slower (138 - > 125).

I tried to restore its speed, and almost reached it, by replacing the first 2 `rotl` by a single one.
However, invariably, some quality test fail somewhere.
So I got tired of trying and settled on this version.

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -674,15 +674,15 @@ XXH3_len_4to8_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_h
     seed ^= (xxh_u64)XXH_swap32((xxh_u32)seed) << 32;
     {   xxh_u32 const input1 = XXH_readLE32(input);
         xxh_u32 const input2 = XXH_readLE32(input + len - 4);
-        xxh_u32 const bitflip1 = (XXH_readLE32(secret+8) ^ XXH_readLE32(secret+12)) + (xxh_u32)(seed >> 32);
-        xxh_u32 const bitflip2 = (XXH_readLE32(secret+16) ^ XXH_readLE32(secret+20)) - (xxh_u32)seed;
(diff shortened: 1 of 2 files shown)


easyaspi314:
Hmm. Wouldn't this be slower than `v0.7.2`?

Cyan4973:
Nope. `v0.7.2` used 3 multiplications (vs 2 here), and delivered a latency bench of 114 (vs 125 here).

easyaspi314:
```c
XXH_FORCE_INLINE XXH64_hash_t
XXH3_len_4to8_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_hash_t seed)
{
    XXH_ASSERT(input != NULL);
    XXH_ASSERT(secret != NULL);
    XXH_ASSERT(4 <= len && len <= 8);
    {   xxh_u32 const input_lo = XXH_readLE32(input);
        xxh_u32 const input_hi = XXH_readLE32(input + len - 4);
        xxh_u64 const input_64 = input_lo | ((xxh_u64)input_hi << 32);
        xxh_u64 const keyed = input_64 ^ (XXH_readLE64(secret) + seed);
        xxh_u64 const mix64 = len + ((keyed ^ (keyed >> 51)) * PRIME32_1); // 1
        return XXH3_avalanche((mix64 ^ (mix64 >> 47)) * PRIME64_2); // 2
    }
}
```
v0.7.2 has two multiplications, three if you count the avalanche. However, one of them is a 32x64 multiply which is faster on 32-bit and AArch64.

However, I presume the reused constants will help a little bit.

PR #319, opened 2020-03-02
[WIP] The Great Typo Hunt

easyaspi314:
- Fix many spelling/grammar issues, primarily in comments
 - Remove most spaces before punctuation
 - Update XXH3 comment
 - Wrap most comments to 80 columns
 - Unify most comments to use the same style
 - Use hexadecimal in the xxhash spec
 - Update help messages to better match POSIX/GNU conventions
 - Use HTML escapes in README.md to avoid UTF-8
 - Mark outdated benchmark/scores

PR #321, opened 2020-03-04
The Great Typo Hunt, Chapter 2

easyaspi314:
Finish what #319 started.

Not perfect, but should be good enough.

Also documented a few extra things like `doubleSecret`, `XXH_mult32to64`,  and the `clock()` issue in `BMK_benchHash`


Cyan4973:
Thanks @easyaspi314 ! This is helpful !

PR #322, opened 2020-03-04
Autovectorization in-source quickfixes

easyaspi314:
Fix some blatant autovec issues without the need for command line flags. Follow-up to #312.

 - Turn on XXH_FORCE_INLINE for XXH3_hashLong on Clang. The scalar path has
   no reason to be autovectorized, it would use the other codepaths.
 - Fix Clang NEONv7 autovectorization in 129to240_64b
 - Fix GCC x86 autovectorization of the long multiply


Cyan4973:
Looks fine !
Great explanations too !

PR #324, opened 2020-03-04
Add pkg-config (.pc) file

bket:


Cyan4973:
Thanks @bket !
Great addition, just on time !

PR #327, opened 2020-03-05
[Website] Fix navbar on mobile

easyaspi314:
Fix the awkward navbar on mobile:
xxhash mobile site issue

### DIFF
--- a/js/bootstrap-patch.css
+++ b/js/bootstrap-patch.css
@@ -10,4 +10,19 @@
 .tab-content > .active,
 .pill-content > .active {
     height: auto;              /* let the content decide it  */
-} /* bootstrap hack end */
+}
+
+/* Fix navbar padding on mobile, which bootstrap messes up for some reason */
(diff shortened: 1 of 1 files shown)


easyaspi314:
Also, @Cyan4973 it looks like xxhash.com still redirects to **http**://cyan4973.github.io/xxHash/ instead of **https**://cyan4973.github.io/xxHash/, even on the HTTPS version.

Cyan4973:
> it looks like xxhash.com still redirects to http://cyan4973.github.io/xxHash/ 

This might be an issue related cache.
I also noticed that effect on some systems, while others would properly redirect to the `https://` version.

These systems are not only (too) complex to setup, the impact of changes also takes time, as it needs to propagate throughout the network, up to the local terminal which is also full of cached information ...

easyaspi314:
Oh yeah, seems like it is just cache. 

Nevermind then.

PR #330, opened 2020-03-09
[HOTFIX] Align malloc, use createState/freeState in sanity check

easyaspi314:
Thanks, @RedSpah for noticing this.

Previously, the 64-byte alignment of XXH3_state_t was only enforced on the stack, and the lower `malloc` alignment requirements caused crashes with the aligned AVX2/SSE2 loads and stores we assumed we could use.

This change ensures that XXH_malloc returns a 64 byte aligned pointer by manually aligning it. This avoids ugly feature tests and, besides, manual alignment would already be required as a portable fallback.

Dynamically allocated states are now used in the sanity check — this would have immediately pointed out this bug.

Due to the rather serious nature of this bug, it is suggested that this be patched into master and released as a hotfix for v0.7.3.

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -814,63 +814,8 @@ XXH_PUBLIC_API XXH128_hash_t XXH128_hashFromCanonical(const XXH128_canonical_t* 
  */
 #include <stdlib.h>
 
-/*
- * Malloc's a pointer that is always 64 byte aligned, to match the alignment
- * of XXH3_state_t.
- *
- * This must be freed with `XXH_free()`.
- *
- * malloc typically guarantees 16 byte alignment on 64-bit systems and 8 byte
- * alignment on 32-bit. This isn't enough for the 32 byte aligned loads in AVX2
- * or on 32-bit, the 16 byte aligned loads in SSE2 and NEON.
- *
- * This underalignment previously caused a rather obvious crash which went
- * completely unnoticed due to XXH3_createState() not actually being tested.
- * Credit to RedSpah for noticing this bug.
- *
- * The alignment is done manually: Functions like posix_memalign or _mm_malloc
- * are avoided: To maintain portability, we would have to write a fallback
- * like this anyways, and besides, testing for the existence of library
- * functions without relying on external build tools is impossible.
- *
- * The method is simple: Overallocate, manually align, and store the offset
- * to the original behind the returned pointer.
- */
-static void* XXH_malloc(size_t s)
-{
-    /* Overallocate to make room for manual realignment and an offset byte */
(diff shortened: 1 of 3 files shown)


easyaspi314:
smh learn how to math, gcc
```c
xxhash.h:839:32: error: conversion to 'unsigned char' from 'unsigned int' may alter its value [-Werror=conversion]
         unsigned char offset = 64 - ((size_t)base % 64);
                                ^
cc1.exe: all warnings being treated as errors
```

xxhash.h
```diff
@@ -813,8 +813,65 @@ XXH_PUBLIC_API XXH128_hash_t XXH128_hashFromCanonical(const XXH128_canonical_t*
...
+    /* Overallocate to make room for manual realignment and an offset byte */
+    unsigned char* base = (unsigned char*)malloc(s + 64);
+    if (base != NULL) {
+        /*
+         * Get the offset needed to align this pointer to 64 bytes.
+         *
+         * Even if the returned pointer is 64-byte aligned, there will always
+         * be at least one byte to store the offset to the original pointer.
+         */
+        size_t offset = 64 - ((size_t)base % 64);
```
Cyan4973: It seems this will generalize 64-bytes alignment for all XXH states, including the XXH32 and XXH64 ones, which don't need it.

Maybe we could have an intermediate layer, like `XXH_malloc_align()`, that would be invoked by the XXH3 state constructor, while XXH32 and XXH64 continue to use the normal `XXH_malloc()`, which works well for them ?

Also, small detail, we can go the safe route to always align on 64-bytes, like the current static `struct` does,
or use the required alignment of the current target, which is supposed to be set by `XXH_ACC_ALIGN`, which would slightly reduce the amount of over-allocated memory.

easyaspi314: Not a bad idea.

I think we should generalize it to 32 bytes. Especially when we implement the dispatcher, we can't tell what the optimal alignment is. 

Unless we plan to add AVX512 (literally can't say it with a straight face 😛), 32 bytes is a good enough alignment.

easyaspi314:
Removed hardcoded 64 byte alignment, renamed it and moved to xxh3.h, and restored XXH_malloc and XXH_free.

Also added a few asserts.

Cyan4973:
Nice fix !
And very appropriate update of the test suite !

easyaspi314:
@Cyan4973 can you patch this into `master` as well? Some people were requesting it in a different project

Cyan4973:
Yes, this is certainly possible,
though I'm not sure if there is a big difference between fetching `dev` or `master`.

easyaspi314:
Mostly psychological. ¯\_(ツ)_/¯

PR #332, opened 2020-03-09
More documentation and cleanup

easyaspi314:
- Mark RC status in xxh3 comment
 - Add XXH128 summary
 - Shill XXH3 in XXH32 and XXH64 comments :stuck_out_tongue: 
 - Start wrapping the absurdly long lines in xxh3.h
 - Minor rearrangements
 - ~~Use `xxh_u32` instead of `U32` in `xxhsum.c`~~
   - ~~Thought this was already the case...~~
   - ~~Don't redefine on XXH_INLINE_ALL~~
 - ~~Cleaned up xxhsum bench~~
   - ~~Remove copy/paste mess~~
   - ~~Make output long enough so we don't cut off on xxHash names~~
   - ~~Make it easier to expand~~

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -469,7 +469,7 @@ XXH_ALIGN(64) static const xxh_u8 kSecret[XXH_SECRET_DEFAULT_SIZE] = {
 };
 
 /*
- * Does a 32-bit to 64-bit long multiply.
+ * Calculates a 32-bit to 64-bit long multiply.
  *
  * Wraps __emulu on MSVC x86 because it tends to call __allmul when it doesn't
  * need to (but it shouldn't need to anyways, it is about 7 instructions to do
@@ -1069,8 +1069,8 @@ XXH3_accumulate_512(      void* XXH_RESTRICT acc,
 #else   /* scalar variant of Accumulator - universal */
 
     XXH_ALIGN(XXH_ACC_ALIGN) xxh_u64* const xacc = (xxh_u64*) acc; /* presumed aligned */
-    const xxh_u8* const xinput = (const xxh_u8*) input;  /* no alignment restriction */
-    const xxh_u8* const xsecret  = (const xxh_u8*) secret;   /* no alignment restriction */
+    const xxh_u8* const xinput  = (const xxh_u8*) input;  /* no alignment restriction */
+    const xxh_u8* const xsecret = (const xxh_u8*) secret;   /* no alignment restriction */
     size_t i;
     XXH_ASSERT(((size_t)acc & (XXH_ACC_ALIGN-1)) == 0);
     for (i=0; i < ACC_NB; i++) {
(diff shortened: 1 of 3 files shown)


xxh3.h
```diff
@@ -1663,21 +1672,36 @@ XXH_PUBLIC_API XXH64_hash_t XXH3_64bits_digest (const XXH3_state_t* state)
...
+ * even without counting the significantly larger output size.
+ *
+ * That strength naturally comes at the cost of speed, especially on short
+ * lengths. Note that longer hashes are about as fast as the 64-bit version
+ * due to it using only a slight modification of the 64-bit loop.
+ *
+ * XXH128 is also more oriented towards 64-bit machines. It is still extremely
+ * fast for a _128-bit_ hash on 32-bit (it usually clears XXH64), but until the
+ * hashLong loop starts, expect roughly half the speed of XXH3_64bits.
+ */
```
easyaspi314: @Cyan4973 got any suggestions for me to add here?

easyaspi314:
Updated XXH128 blurb.

xxh3.h
```diff
@@ -1663,21 +1672,42 @@ XXH_PUBLIC_API XXH64_hash_t XXH3_64bits_digest (const XXH3_state_t* state)
...
 
 /* ==========================================
- * XXH3 128 bits (=> XXH128)
- * ========================================== */
+ * XXH3 128 bits (a.k.a XXH128)
+ * ==========================================
+ * XXH3's 128-bit variant has better mixing and strength than the 64-bit variant,
+ * even without counting the significantly larger output size.
+ *
+ * Unlike XXH3_64bits which aims for excellency, XXH128 aims for perfection.
```
Cyan4973: This might be a tad too much, let's not bait trolls.
The first paragraph is good enough I believe.

xxh3.h
```diff
@@ -1663,21 +1672,42 @@ XXH_PUBLIC_API XXH64_hash_t XXH3_64bits_digest (const XXH3_state_t* state)
...
+ * For example, extra steps are taken to avoid the seed-dependent collisions
+ * in 17-240 byte inputs (See XXH3_mix16B and XXH128_mix32B).
+ *
+ * This strength naturally comes at the cost of some speed, especially on short
+ * lengths. Note that longer hashes are about as fast as the 64-bit version
+ * due to it using only a slight modification of the 64-bit loop.
+ *
+ * XXH128 is also more oriented towards 64-bit machines. It is still extremely
+ * fast for a _128-bit_ hash on 32-bit (it usually clears XXH64), but until the
+ * hashLong loop starts, expect roughly half the speed of XXH3_64bits.
```
Cyan4973: I should measure it again. 
Last time I checked, while it was undeniably slower, I did not remember a x2 speed difference.

easyaspi314: Wait, I think Clang has an issue with XXH128 that makes it slower than GCC which I was planning on looking into

easyaspi314: Removed speed comment, it isn't really required.

xxh3.h
easyaspi314: Ok

easyaspi314:
Cleaned up the xxhsum bench. Now uses a table and a loop.

Not only does this eliminate the ugly copy/paste, but it makes it super easy to add your own hashes to the benchmark.

xxh3.h
easyaspi314: So like this?
```c
/* ==========================================
 * XXH3 128 bits (a.k.a XXH128)
 * ==========================================
 * XXH3's 128-bit variant has better mixing and strength than the 64-bit variant,
 * even without counting the significantly larger output size.
 *
 * For example, extra steps are taken to avoid the seed-dependent collisions
 * in 17-240 byte inputs (See XXH3_mix16B and XXH128_mix32B).
 * ...
 */
```

xxhsum.c
```diff
@@ -578,82 +623,61 @@ static void BMK_benchHash(hashFunction h, const char* hName, const void* buffer,
...
  * specificTest: 0 == run all tests, 1+ runs specific test
- * buffer: Must be 8-byte aligned (if malloc'ed, it should be)
+ * buffer: Must be 16-byte aligned.
  * The real allocated size of buffer is supposed to be >= (bufferSize+3).
  * returns: 0 on success, 1 if error (invalid mode selected)
  */
-static int BMK_benchMem(const void* buffer, size_t bufferSize, U32 specificTest)
+static int BMK_benchMem(const void* buffer, size_t bufferSize, xxh_u32 specificTest)
 {
-    assert((((size_t)buffer) & 8) == 0);  /* ensure alignment */
```
easyaspi314: This error went completely unnoticed because we manually align to *16* bytes

xxhsum.c
```diff
@@ -578,82 +626,61 @@ static void BMK_benchHash(hashFunction h, const char* hName, const void* buffer,
...
+            assert(g_hashesToBench[i].name != NULL);
+            /* aligned */
+            if (specificTest == 0 || specificTest == 2 * i + 1) {
+                BMK_benchHash(g_hashesToBench[i].func, g_hashesToBench[i].name, buffer, bufferSize);
+            }
+            /* unaligned */
+            if (specificTest == 0 || specificTest == 2 * i + 2) {
+                /* Append "unaligned". */
+                char hashNameBuf[HASHNAME_MAX + 1];
+                snprintf(hashNameBuf, sizeof(hashNameBuf), "%s unaligned", g_hashesToBench[i].name);
```
easyaspi314: snprintf usage ok?

easyaspi314:
Actually, I'll move the xxhsum bench to a different PR. 

Just let me know if you want me to keep that line in the xxh128 comment and I'll rebase.

xxhsum.c
easyaspi314: Nevermind, apparently MSVC 2013 lacks `snprintf`…guess we are using strncat.

xxh3.h
Cyan4973: Yep, that's fine

easyaspi314:
Rebased, xxhsum bench/typedef changes removed.

easyaspi314:
Should be ready to merge once the tests (hopefully) pass.

Cyan4973:
Looks great !
Thanks @easyaspi314 !

PR #333, opened 2020-03-09
xxhsum bench improvements

easyaspi314:
- Clean up xxhsum benchmark code
   - Reformat
   - Use a table and a loop to abstract it
      - Removes copy/paste mess
      - Makes adding new hashes to test trivial
   - Fix an incorrect `assert` which just happened to work 
 - Add `withSecret` to the benchmark
   - Uses a static buffer because it is easier than passing around a secret (is that ok?)
   - Generated in the same way as the sanity buffer (it is now in a function)
 - Make the output lines long enough to fit the xxHash names.
 - Fix some minor issues and remove some unused macros

### DIFF
--- a/xxhsum.c
+++ b/xxhsum.c
@@ -583,17 +583,17 @@ typedef struct {
 } hashInfo;
 
 static const hashInfo g_hashesToBench[] = {
-    { "XXH32",            &localXXH32 },
-    { "XXH64",            &localXXH64 },
-    { "XXH3_64b",         &localXXH3_64b },
-    { "XXH3_64b seeded",  &localXXH3_64b_seeded },
-    { "XXH3_64b secret",  &localXXH3_64b_secret },
-    { "XXH128",           &localXXH3_128b },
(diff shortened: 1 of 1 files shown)


easyaspi314:
Also I checked the MSVC 2013 docs.

It seriously *does* have `vsnprintf`, but not `snprintf`. (It has `_snprintf` which is non-conforming AFAICT)

xxhsum.c
```diff
@@ -1927,18 +1922,18 @@ static void errorOut(const char* msg)
  * Will also modify `*stringPtr`, advancing it to position where it stopped reading.
  * @return 1 if an overflow error occurs
  */
-static int readU32FromCharChecked(const char** stringPtr, unsigned* value)
```
easyaspi314: Forgot to mention in the commit message :slightly_frowning_face:

Cyan4973:
Thanks @easyaspi314 , it's a nice code cleanup !

PR #334, opened 2020-03-10
Minor Makefile fixes

easyaspi314:
- Don't use `-mavx2 -maes` in `collisions`
   - Prevents `make test` from passing on non-x86 targets
 - ~~Default `$(CC)` to `gcc`~~
   - ~~classic mingw32 has `gcc` but not `cc`~~
 - Use the more proper `-std` flag instead of `--std`

TODO: proper Windows library flags: bfd accepts Linux flags, but
ld.lld errors on `-Wl,-soname`. Might be a later PR.

### DIFF
--- a/tests/collisions/Makefile
+++ b/tests/collisions/Makefile
@@ -28,10 +28,8 @@ VPATH = $(SRC_DIRS)
 CPPFLAGS += $(addprefix -I ,$(SRC_DIRS))
 CFLAGS   ?= -std=c99 \
             -Wall -Wextra -Wconversion
-CFLAGS   += -maes -mavx2
 CXXFLAGS ?= -Wall -Wextra -Wconversion --std=c++11
 LDFLAGS  += -pthread
-LDFLAGS  += -maes -mavx2
 TESTHASHES = 110000000
 
 HASH_SRC := $(sort $(wildcard allcodecs/*.c allcodecs/*.cc))
--- a/tests/collisions/Makefile
+++ b/tests/collisions/Makefile
@@ -18,9 +18,9 @@
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 #
-#  You can contact the author at :
+#  You can contact the author at:
 #  - xxHash homepage: https://www.xxhash.com
-#  - xxHash source repository : https://github.com/Cyan4973/xxHash
+#  - xxHash source repository: https://github.com/Cyan4973/xxHash
 #
 
 SRC_DIRS = ./ ../../ allcodecs/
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -34,6 +34,8 @@ LIBVER_MINOR := $(shell echo $(LIBVER_MINOR_SCRIPT))
 LIBVER_PATCH := $(shell echo $(LIBVER_PATCH_SCRIPT))
 LIBVER := $(LIBVER_MAJOR).$(LIBVER_MINOR).$(LIBVER_PATCH)
 
+CC     ?= gcc
+CC     ?= g++
```
Cyan4973: I presume you meant `CXX` ?

Cyan4973:
Wow, `$(CC)` is problematic on mingw32 ?

Makefile
easyaspi314: Yup 🤦‍♂️

easyaspi314:
Yep:

```
~/xxHash $ make
cc -O3  -c xxhash.c -o xxhash.o
make: cc: No such file or directory
make: *** [Makefile:87: xxhash.o] Error 127
$
```

Cyan4973:
Now that you mention it, I vaguely remember having to specify `CC=` on the command line while testing compilation under `mingw32`.

That being said, it feels kind of disproportionate to set a non-standard "default" for _all_ platforms just to avoid having to specify a `CC=gcc` argument on a single relatively minor one.

easyaspi314:
> That being said, it feels kind of disproportionate to set a non-standard "default" for all platforms just to avoid having to specify a CC=gcc argument on a single relatively minor one.

Name one platform where `cc` isn't a symlink to `gcc`. Besides, people have been using `CC ?= gcc` for years without a problem.

easyaspi314:
Wait…I'm dumb. `CC := gcc`

Cyan4973:
> Name one platform where cc isn't a symlink to gcc.

FreeBSD and macosx default to `clang`. For `macosx`, it's actually their own flavor of `clang`, with their own numbering.

I wouldn't be surprised if some IBM and some Oracle systems default to their own compiler.

The standard allows for some flexibility on this topic, and I believe it's better to respect that. That's part of the little details that make the library and its default build system more "portable".

> Wait…I'm dumb. CC := gcc

This is actually worse. At least, with `?=`, `CC` can be set from environment. While with `:=`, `CC` is unconditionally set, and the only remaining choice is to override it by setting it as an argument of `make` itself.

easyaspi314:
> FreeBSD and macosx default to clang. For macosx, it's actually their own flavor of clang, with their own numbering.

Well guess what `gcc` is a link to? `clang`

This is actually worse. At least, with ?=, CC can be set from environment. While with :=, CC is unconditionally set, and the only remaining choice is to override it by setting it as an argument of make itself. Some cross-compilation tests fail due to this.

Oof, true. Yeah, totally didn't think of it, I'll revert the change.

easyaspi314:
Ok, `$(CC)` is no longer set to `gcc`.

Cyan4973:
> Well guess what gcc is a link to? clang

Yeah, I know.
They do that because many build scripts default to `gcc`, and they (correctly) guess that it's better to ease the transition, rather than overflowing support with a ton of outcry.
It doesn't mean it's right to set `gcc` in the build script.

Let's be clear that `CC ?= gcc` isn't "horrible". Quite frankly, it's within acceptable range.
Just if we have the choice, I _prefer_ we stay on the side of the standard. And I believe we do have such a choice, as `mingw32` is simply not an important enough platform to deserve carving an exception for.

PR #335, opened 2020-03-10
New Windows Unicode solution, works on XP

easyaspi314:
Now uses `WriteConsoleW` instead of `vfwprintf`. This is codepage independent and works on Windows XP.

Doing this also avoids the `_O_U8TEXT` hack.

Also use `_vsnprintf` instead of `vsnprintf`, as it is more consistent than `vsnprintf` in that it doesn't write the null terminator on truncation. Weird.

Some minor cleanup was done as well.

### DIFF
--- a/xxhsum.c
+++ b/xxhsum.c
@@ -126,10 +126,15 @@ static __inline int IS_CONSOLE(FILE* stdStream) {
 
 /* Unicode helpers for Windows */
 #if defined(_WIN32)
-/* Converts a UTF-8 string to UTF-16. Acts like strdup. The string must be freed afterwards. */
-static wchar_t *utf8_to_utf16(const char *str)
+/*
+ * Converts a UTF-8 string to UTF-16. Acts like strdup. The string must be freed afterwards.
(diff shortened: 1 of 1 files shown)


easyaspi314:
Truly fixes #293. 

unicode fixed

sergeevabc:
Err… I’m a mere user without compiler. Could I have a binary to verify the fix, please?

easyaspi314:
[xxhsum-windows-7328e217.zip](https://github.com/Cyan4973/xxHash/files/4315081/xxhsum-windows-7328e217.zip)

sergeevabc:
Oh, so many binaries, it makes me confused all the time. Right, I confirm that
```xxhsum-i686-mingw-w64.exe``` and ```xxhsum-i686-mingw32.exe``` finally work as expected,
and report that ```xxhsum-i686-msvc2019.exe``` cannot be launched (XP says “not Win32 app”).

easyaspi314:
I might have used the wrong filename, or MSVC builds just don't work with XP. I'll test in my VM later.

PR #336, opened 2020-03-12
update master

Cyan4973:
with the xxh3 state alignment fix


easyaspi314:
Release notes:

```md
`v0.7.4` is a bugfix release for `v0.7.3`.

No changes to `XXH3`'s algorithm have been made.

It primarily exists to fix a `malloc` underalignment issue in `XXH3_createState()`, causing crashes with SIMD aligned loads.

If you only use the single shot functions or use `XXH3_state_t` directly (e.g. on the stack), there is no reason to update.

There are also some tweaks to `xxhsum`'s benchmark, and some formatting improvements.
```

Cyan4973:
Release note ?
Do you imply you would like a release, just for this fix ?

easyaspi314:
Well you changed the version number so I naturally assumed that was what you were planning on doing. :confused:

easyaspi314:
> updated version number in order to distinguish users of the release from potential users of the development version

…but what if we don't go straight to `v0.8.0` and need to do a real `v0.7.4`? Then we have confusion.

Cyan4973:
ah no, don't worry, I believe that's fine.

When people report a bug, we will be able to ask the version nb.
If they say `v0.7.3`, we know it's the release version,
if they say `v0.7.4`, we know it's the development version.

That's all there is.

PR #339, opened 2020-03-16
Reduce a XXH_mult32to64 in scalar version XXH_mult64to128

gzm55:
With Karatsuba technique, call `XXH_mult32to64()` 3 times in scalar version of `XXH_mult64to128()`. This may help on some 32bit CPU with very high latency for 32x32-->64 mutiplication.

I've no 32bit CPU, but on modern Intel 64bit CPU, this method may sometimes improve performance a little if force to use scalar version of `XXH_mult64to128()`.

It could be better to keep both the original one and this method, and select the best for some proper CPUs.

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -594,16 +594,32 @@ XXH_mult64to128(xxh_u64 lhs, xxh_u64 rhs)
      *     of 32-bit ADD/ADCs.
      */
 
-    /* First calculate all of the cross products. */
-    xxh_u64 const lo_lo = XXH_mult32to64(lhs & 0xFFFFFFFF, rhs & 0xFFFFFFFF);
-    xxh_u64 const hi_lo = XXH_mult32to64(lhs >> 32,        rhs & 0xFFFFFFFF);
-    xxh_u64 const lo_hi = XXH_mult32to64(lhs & 0xFFFFFFFF, rhs >> 32);
-    xxh_u64 const hi_hi = XXH_mult32to64(lhs >> 32,        rhs >> 32);
+    xxh_u64 const lhs_lo32 = (xxh_u32)lhs;
+    xxh_u64 const lhs_hi32 = lhs >> 32;
+    xxh_u64 const lhs_fold33 = lhs_lo32 + lhs_hi32;
+    xxh_u64 const lhs_fold_lo32 = (xxh_u32)lhs_fold33;
+    xxh_u64 const lhs_fold_hi1 = lhs_fold33 >> 32;
 
-    /* Now add the products together. These will never overflow. */
-    xxh_u64 const cross = (lo_lo >> 32) + (hi_lo & 0xFFFFFFFF) + lo_hi;
-    xxh_u64 const upper = (hi_lo >> 32) + (cross >> 32)        + hi_hi;
-    xxh_u64 const lower = (cross << 32) | (lo_lo & 0xFFFFFFFF);
+    xxh_u64 const rhs_lo32 = (xxh_u32)rhs;
+    xxh_u64 const rhs_hi32 = rhs >> 32;
+    xxh_u64 const rhs_fold33 = rhs_lo32 + rhs_hi32;
+    xxh_u64 const rhs_fold_lo32 = (xxh_u32)rhs_fold33;
+    xxh_u64 const rhs_fold_hi1 = rhs_fold33 >> 32;
+
+    xxh_u64 const L   = XXH_mult32to64(lhs_lo32, rhs_lo32);
+    xxh_u64 const H   = XXH_mult32to64(lhs_hi32, rhs_hi32);
+    xxh_u64 const C_L = XXH_mult32to64(lhs_fold_lo32, rhs_fold_lo32);
+    xxh_u64 const C_K = ((0-lhs_fold_hi1) & rhs_fold_lo32) + ((0-rhs_fold_hi1) & lhs_fold_lo32);
+    xxh_u64 const C_H = lhs_fold_hi1 & rhs_fold_hi1;
+
+    // C = C_H * 2^64 + C_K * 2^32 + C_L
+    // K = C - H - L
+    xxh_u64 const K_lo33 = (xxh_u64)(xxh_u32)C_L + 0x200000000ULL - (xxh_u32)H - (xxh_u32)L;
+    xxh_u64 const K_hi33 = (C_L >> 32) + C_K + (C_H << 32) - 2 - (H >> 32) - (L >> 32);
+
+    // lhs * rhs = (H + K_hi33) * 2^64 + (K_lo33) * 2^32 + L
+    xxh_u64 const lower = (K_lo33 << 32) + L;
+    xxh_u64 const upper = (((L>>32) + K_lo33)>>32) + K_hi33 + H;
 
     XXH128_hash_t r128;
     r128.low64  = lower;


easyaspi314:
Eh, I doubt it will be faster.

With the exception of like Pentium 4, long multiplies are pretty cheap. 

Even Pentium M has a 5 cycle latency and 1 cycle throughput.

The other thing is that this version compiles to 4 instructions on ARMv6 with the DSP extension, and 32-bit ARM is a very important target due to there still being a market for those chips. Cortex-M3 and M4 are rather popular in the embedded market, and there are still a lot of ARMv7-A phones out there.

```asm
@ r0: XXH128_hash_t *r128, r2:r3: u64 lhs, sp[0-7]: u64 rhs
XXH_mult64to128:
        push    {r4, lr}                @ save caller saved registers
        ldr     r12, [sp, #8]           @ load rhs from stack
        ldr     r4, [sp, #12]           @ ditto
        umull   r1, lr, r12, r2         @ lo_lo = rhs.lo * lhs.lo
        str     r1, [r0]                @ store r128.low64.lo
        umull   r1, r12, r12, r3        @ hi_lo = rhs.lo * lhs.hi
        umaal   r1, lr, r4, r2          @ cross = rhs.hi * lhs.lo + (hi_lo & 0xffffffff) + (lo_lo >> 32)
        umaal   r12, lr, r4, r3         @ upper = rhs.hi * lhs.hi + (hi_lo >> 32)        + (cross >> 32)
        str     r1, [r0, #4]            @ store r128.low64.hi
        str     lr, [r0, #12]           @ store r128.high64.hi
        str     r12, [r0, #8]           @ store r128.high64.lo
        pop     {r4, pc}                @ pop caller saved registers and return
```

Besides, one of the requirements for XXH3 is a decent long multiply. 🙂

easyaspi314:
For the reference, here is what the same version of Clang generates on ARMv6:
```asm
XXH_mult64to128:
        push    {r4, r5, r6, r7, r8, r9, r10, r11, lr}
        sub     sp, sp, #4
        ldr     r8, [sp, #44]
        ldr     r12, [sp, #40]
        mov     r9, #0
        umull   r5, r6, r8, r3
        umull   r10, lr, r12, r2
        subs    r7, r5, r6
        sbc     r6, r6, #0
        subs    r7, r7, lr
        sbc     r11, r6, #0
        adds    r4, r12, r8
        adc     r12, r9, #0
        rsb     r6, r12, #0
        adds    r2, r2, r3
        and     r3, r2, r6
        adc     r6, r9, #0
        rsb     r1, r6, #0
        and     r1, r4, r1
        umull   r2, r4, r4, r2
        adds    r3, r7, r3
        adc     r7, r11, #0
        adds    r1, r3, r1
        adc     r3, r7, #0
        adds    r1, r1, r4
        adc     r3, r3, #0
        mov     r4, #2
        subs    r2, r2, r5
        sbc     r4, r4, #0
        subs    r2, r2, r10
        sbc     r4, r4, #0
        adds    r2, r2, lr
        str     r2, [r0, #4]
        adc     r2, r4, #0
        and     r7, r12, r6
        adds    r1, r1, r2
        adc     r2, r3, r7
        subs    r1, r1, #2
        str     r10, [r0]
        str     r1, [r0, #8]
        sbc     r1, r2, #0
        str     r1, [r0, #12]
        add     sp, sp, #4
        pop     {r4, r5, r6, r7, r8, r9, r10, r11, pc}
```

Cyan4973:
Are there ways to compare speed on some 32-bit targets ?
Maybe some target would prefer one variant over the other ?

easyaspi314:
I can see if my other Pentium 4 still works (we got rid of the one I tested earlier, but I have another one in my basement).

The other thing is that the primary use of the i686 version is going to be 32-bit Windows binaries on 64-bit CPUs, whether it is because 32-bit Windows is needed for some specific program (some businesses need 16-bit programs), devs only building 32-bit binaries for ease of distribution, or because a large portion of the technologically illiterate don't know which version to download for their CPU.

Cyan4973:
I agree that 32-bit Window is a decreasing presence. 
I had in mind some other "active" 32-bit systems, such as ARM, though I understand it can be more difficult to test.

easyaspi314:
ARM is definitely going to prefer the current `umaal` path. It was literally designed for multiprecision multiplication.

With the exception of the "smol" Thumb 1 only Cortex-M chips (which lack a long multiply anyways and therefore are incompatible), most ARM chips have a long multiply that is 3-4 cycles. Some higher-end Cortex-M and Cortex-R chips have single cycle multiplies, and even the ARM7TDMI is only 6 cycles.

Even without `umaal`, the current path is pretty efficient (not optimal, though):
```asm
XXH_mult64to128:
        push    {r4, r5, r6, lr}
        ldr     r12, [sp, #16]
        ldr     r5, [sp, #20]
        mov     r6, #0
        umull   r1, lr, r12, r2
        str     r1, [r0]
        umull   r1, r4, r12, r3
        mov     r12, #0
        umlal   r1, r6, r5, r2
        umlal   r4, r12, r5, r3
        adds    r1, r1, lr
        str     r1, [r0, #4]
        adc     r1, r6, #0
        adds    r1, r4, r1
        str     r1, [r0, #8]
        adc     r1, r12, #0
        str     r1, [r0, #12]
        pop     {r4, r5, r6, pc}
```

Cyan4973:
So, since the situation on ARM looks pretty clear,
is there any other 32-bit target where Karatsuba could stand a chance ?

easyaspi314:
Highly doubt it. The only time where Karatsuba is worth it is in carryless multiplication ([example](https://github.com/DLTcollab/sse2neon/blob/fabdcb5d7d8d503865f7be999030015aac89d9db/tests/impl.cpp#L2178)) where a single scalar subproduct requires multiple steps and there is no carry tracking, or with huge numbers, roughly 32 subproducts or more.

Cyan4973:
OK, so to summarize, 
thanks for implementing and suggesting this Karatsuba method @gzm55 , it's a nice effort that could be worthwhile, 
now the next stage is to prove that it's indeed as beneficial as it sounds, for at least some real-world targets, by measuring / comparing it using benchmarks.
Apparently, it's a tough battle on `ARM`, but maybe other targets might prove more favorable ?Does that sound reasonable ?

easyaspi314:
The only thing that *might* be faster on x86 is SSE2, but it'll be for 32-bit only and would look incredibly ugly. [Sound familiar?](https://github.com/easyaspi314/xxHash/blob/b0f8e1c25fc0cc93a00d902690495ac706f0963f/xxhash-vec.h#L613) :nauseated_face:

gzm55:
@easyaspi314 thanks for deeply analyzing the codes. This method should be faster only if a long multiplication is about 10 cycles and the processor has less than 10 common registers, since the carry tracing is very complex especially when calulating the `C` and `K`. But this kind of processors is rare found in the real world if we limit the processors having 32x32->64 "decent long multiply". It is very interesting that the JVM [`Math.multiplyHigh()`](https://github.com/AdoptOpenJDK/openjdk-jdk14/blob/master/src/java.base/share/classes/java/lang/Math.java#L1107) use a simplified K-method with 3 64x64->L64 multiplications for positives multiplicators.

What about considering some IoT chips with only 32x32->L32 multiply, including Cortex-M0,M0+,M1, M23, and the `MUL` latencies for these chips may [very high](https://en.wikipedia.org/wiki/ARM_Cortex-M#Instruction_sets). Although this may 'break' the requirement, we can still do limited optimization for these processors. That also means most 32-bit processors will be covered. (PS, for non-decent 32-bit `MUL`, `XXH_mult32to64()` also needs to be rewritten as 3 or 4 `MUL`.)

easyaspi314:
> It is very interesting that the JVM Math.multiplyHigh() use a simplified K-method with 3 64x64->L64 multiplications for positives multiplicators.

Well actually, it is doing a 63x63 multiply. With the `if (x < 0 || y < 0)` test, it is testing if bit 64 is set. Without that, the arithmetic will overflow. That is the main issue with Karatsuba — it was made for pen and paper (not computers), and without a bit to overflow, you would have to do a 65-bit multiply.


> What about considering some IoT chips with only 32x32->L32 multiply, including Cortex-M0,M0+,M1, M23, and the MUL latencies for these chips may very high. Although this may 'break' the requirement, we can still do limited optimization for these processors. That also means most 32-bit processors will be covered. (PS, for non-decent 32-bit MUL, XXH_mult32to64() also needs to be rewritten as 3 or 4 MUL.)

I am fully aware of the M0 incompatibility. 

The problem isn't just the multiply, Thumb has a huge issue with register pressure in XXH3. ([read my comment](https://github.com/Cyan4973/xxHash/blob/7fc2b540470de98d5ad70e327d4400a5c9420cbf/xxh3.h#L131))

Pretending that Thumb has unaligned access and `umull` (not pushing it with `umlal`), the 64-bit `XXH3_accumulate_512` loop barely squeezes into the registers:
```asm
XXH3_accumulate_512_64b:
        push    {r4-r7, lr}
        movs    r3, #8  @ ACC_NB
.Lloop:
        mov     r12, r3
        ldr     r4, [r0]
        ldr     r5, [r0, #4]
        ldr     r6, [r1]
        ldr     r7, [r1, #4]
        adds    r4, r6
        adcs    r5, r7
        ldr     r3, [r2]
        eors    r6, r3
        ldr     r3, [r2, #4]
        eors    r7, r3
        umull   r3, r7, r6, r7
        adds    r4, r3
        adcs    r5, r7
        stmia   r0!, {r4, r5} 
        adds    r1, #8
        adds    r2, #8
        mov     r3, r12
        subs    r3, #1
        bne     .Lloop
.Lloop_end:
        pop     {r4-r7, pc}
```
However, M0 lacks unaligned access — so it will be trying to do byteshift loads. 
And of course, `umull` doesn't exist.

Clang generates this if I write a basic 32->64-bit multiply routine with `-Oz`
```asm
XXH3_accumulate_512:
        push    {r4, r5, r6, r7, lr}
        sub     sp, #28
        str     r2, [sp, #4]            @ 4-byte Spill
        str     r1, [sp, #24]           @ 4-byte Spill
        str     r0, [sp]                @ 4-byte Spill
        movs    r4, #0
.LBB0_1:                                @ =>This Inner Loop Header: Depth=1
        cmp     r4, #64
        beq     .LBB0_3
@ %bb.2:                                @   in Loop: Header=BB0_1 Depth=1
        ldr     r0, [sp, #24]           @ 4-byte Reload
        adds    r6, r0, r4
        ldrb    r1, [r6, #4]
        ldrb    r2, [r6, #5]
        lsls    r2, r2, #8
        adds    r1, r2, r1
        ldrb    r2, [r6, #6]
        ldrb    r3, [r6, #7]
        lsls    r3, r3, #8
        adds    r2, r3, r2
        lsls    r2, r2, #16
        adds    r2, r2, r1
        str     r2, [sp, #20]           @ 4-byte Spill
        ldr     r0, [sp, #4]            @ 4-byte Reload
        adds    r3, r0, r4
        ldrb    r1, [r3, #4]
        ldrb    r5, [r3, #5]
        lsls    r5, r5, #8
        adds    r1, r5, r1
        ldrb    r5, [r3, #6]
        ldrb    r7, [r3, #7]
        lsls    r7, r7, #8
        adds    r5, r7, r5
        lsls    r5, r5, #16
        adds    r7, r5, r1
        eors    r7, r2
        lsrs    r5, r7, #16
        ldr     r1, [sp, #24]           @ 4-byte Reload
        ldrb    r1, [r1, r4]
        ldrb    r2, [r6, #1]
        lsls    r2, r2, #8
        adds    r1, r2, r1
        ldrb    r2, [r6, #2]
        ldrb    r6, [r6, #3]
        lsls    r6, r6, #8
        adds    r2, r6, r2
        lsls    r2, r2, #16
        adds    r6, r2, r1
        str     r6, [sp, #16]           @ 4-byte Spill
        ldrb    r1, [r0, r4]
        ldrb    r2, [r3, #1]
        lsls    r2, r2, #8
        adds    r1, r2, r1
        ldrb    r2, [r3, #2]
        ldrb    r3, [r3, #3]
        lsls    r3, r3, #8
        adds    r2, r3, r2
        lsls    r2, r2, #16
        adds    r1, r2, r1
        eors    r1, r6
        uxth    r3, r1
        mov     r2, r5
        mov     r6, r5
        muls    r2, r3, r2
        lsrs    r1, r1, #16
        uxth    r5, r7
        mov     r7, r1
        muls    r7, r5, r7
        movs    r0, #0
        adds    r2, r7, r2
        str     r2, [sp, #8]            @ 4-byte Spill
        adcs    r0, r0
        lsrs    r2, r2, #16
        lsls    r0, r0, #16
        adds    r0, r0, r2
        str     r0, [sp, #12]           @ 4-byte Spill
        muls    r6, r1, r6
        muls    r3, r5, r3
        ldr     r0, [sp]                @ 4-byte Reload
        adds    r2, r0, r4
        ldr     r1, [r2, #4]
        ldr     r5, [r0, r4]
        ldr     r7, [sp, #16]           @ 4-byte Reload
        adds    r5, r5, r7
        ldr     r7, [sp, #20]           @ 4-byte Reload
        adcs    r1, r7
        adds    r3, r3, r5
        adcs    r1, r6
        ldr     r5, [sp, #8]            @ 4-byte Reload
        lsls    r5, r5, #16
        adds    r3, r3, r5
        str     r3, [r0, r4]
        ldr     r0, [sp, #12]           @ 4-byte Reload
        adcs    r1, r0
        str     r1, [r2, #4]
        adds    r4, #8
        b       .LBB0_1
.LBB0_3:
        add     sp, #28
        pop     {r4, r5, r6, r7, pc}
```
Yes, that is one iteration, no unrolling here.

Having to load *two* unaligned 64-bit integers _and_ do a 32-bit to 64-bit multiply is too much. That is why I explicitly warn on `-mthumb` on non-cortex-M0 devices.

There are more practical hash functions to run on it.

gzm55:
> it was made for pen and paper (not computers), and without a bit to overflow, you would have to do a 65-bit multiply.

Yes, common usage of K-method is big number arithmetics which have no overflow problem. The 65-bit multiply is the slowest part to workaround.

For 32-bit ARM processors, as @easyaspi314 analyzed, it seems no suitable candidates.

Cyan4973:
Trying to settle on this PR,
my understanding of this thread is that
the Karatsuba technique is worth investigating, but effectively failed to deliver a benefit for our use case.
Is that correct ?

gzm55:
> Trying to settle on this PR,
> my understanding of this thread is that
> the Karatsuba technique is worth investigating, but effectively failed to deliver a benefit for our use case.
> Is that correct ?

Yes, right now we still cannot find a proper target hardware benefit from this solution, so I will close this PR.

PR #342, opened 2020-03-22
always skip scrambling for the last block

gzm55:
This will reduce a scrambling when length is align to block. On AVX2, 1K input will improve about 6% performance, 10K will improve about 3%.

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -1302,7 +1302,7 @@ XXH3_hashLong_internal_loop( xxh_u64* XXH_RESTRICT acc,
 {
     size_t const nb_rounds = (secretSize - STRIPE_LEN) / XXH_SECRET_CONSUME_RATE;
     size_t const block_len = STRIPE_LEN * nb_rounds;
-    size_t const nb_blocks = len / block_len;
+    size_t const nb_blocks = (len - 1) / block_len;
 
     size_t n;
 
--- a/xxh3.h
+++ b/xxh3.h
@@ -1315,14 +1315,13 @@ XXH3_hashLong_internal_loop( xxh_u64* XXH_RESTRICT acc,
 
     /* last partial block */
     XXH_ASSERT(len > STRIPE_LEN);
-    {   size_t const nbStripes = (len - (block_len * nb_blocks)) / STRIPE_LEN;
+    {   size_t const nbStripes = (len - (block_len * nb_blocks) - 1) / STRIPE_LEN;
         XXH_ASSERT(nbStripes <= (secretSize / XXH_SECRET_CONSUME_RATE));
         XXH3_accumulate(acc, input + nb_blocks*block_len, secret, nbStripes, accWidth);
 
         /* last stripe */
-        if (len & (STRIPE_LEN - 1)) {
-            const xxh_u8* const p = input + len - STRIPE_LEN;
-            /* Do not align on 8, so that the secret is different from the scrambler */
+        {   const xxh_u8* const p = input + len - STRIPE_LEN;
+            /* May not align on 8, so that the secret is different from the scrambler */
 #define XXH_SECRET_LASTACC_START 7
             XXH3_accumulate_512(acc, p, secret + secretSize - STRIPE_LEN - XXH_SECRET_LASTACC_START, accWidth);
     }   }


Cyan4973:
Unfortunately, this modification will also change the return value of the hash for inputs which are multiple of block size (1 KB by default), thus making it a breaking change, failing the validation test.

At this stage, we are trying to stabilize the output value of the hash function, significantly raising the bar to justify another change, as it would delay the finalization of the algorithm by a few more months.

gzm55:
@Cyan4973 Indeed it is, but only the hashes on the 1k boundary is changed. If someone don't care this length, there is no change for him; on the other hand, if he cares this length, the improvement should not be ignored. So although xxh3 is in rc state, i still post this pr for discussion.

easyaspi314:
Yeah, unless we find something huge (like a massive collision mess), we want to avoid changing it.

We would have to thoroughly test it, which means Yann has to borrow a server to run `tests/collisions`, and it is just a domino effect.

gzm55:
ok, i'll leave this pr here, and close it after v0.8 is released.

Cyan4973:
Changed target branch.
All PRs impacting return values will be merged into `newOutput074`, for proper validation, before joining `dev`.

Cyan4973:
Note on this specific PR : 
the same change must be reflected on both the one-shot and the streaming API. Both must remain in sync.

Cyan4973:
I'm experimenting with this candidate modification in `v0.7.4`,
one difficult part is to ensure that both the one-shot and streaming API remain synchronized.
So far, it's not the case in this PR.

I can attempt to do it on my side, though it's not clear yet if the logic can be replicated simply on the streaming side without inserting some kind of "exception code".

Cyan4973:
Here is my understanding of what this patch is doing.

Using default `secret` (implying a block size of 1 KB), we currently have :

| input len | M iter/s | comment |
| --- | --- | --- |
| 959 | 17.88 | 15 stripes
| 960 | 17.88 | 15 stripes
| 961 | 17.06 | 16  stripes
| 1023 | 17.06 | 16 stripes
| 1024 | 16.39 | 16 stripes + scrambler (1 full block) **<==** 
| 1025 | 15.71 | 1 block + 1 stripe

The patch proposes to optimize speed for input lengths which are _precisely_ multiple of the block size. As a side-effect, it unconditionally uses the unaligned "last secret" for the last stripe, saving one branch, but also impacting the return value of any input length multiple of `STRIPE_LEN` (==`64`). Which gives : 

| input len | M iter/s | comment |
| --- | --- | --- |
| 959 | 17.88 | _unmodified_
| 960 | 17.88 | **different return value**
| 961 | 17.06 | _unmodified_
| 1023 | 17.06 | _unmodified_
| 1024 | **17.06** | **faster, and different return value**
| 1025 | 15.71 | _unmodified_

And that's basically it, the whole point is to get this `1024` input length a bit faster (and then N*`1024` multiples, though the relative savings become proportionally smaller). Let's be honest, the benefit is a bit narrow.

Anyway, as long as the patch is associated to the "one-shot` API, it is well localized, and doesn't impact complexity much, so it's a reasonable candidate.

Problem is, the _streaming_ variant must now apply the same logic, for results to remain comparable.

And here is the issue : 
the streaming API buffers input data by blocks of 4 stripes (256 bytes). On reaching that amount, it consumes the internal buffer, and then try to consume input directly whenever it's large enough (larger than internal buffer).
The decision to use the scrambler is made inside the inner loop. Any time it consumes the nb of stripes required to make a block (not necessarily a multiple of 4, since `secret` can have custom sizes), it unconditionally triggers a scrambler.

So the logic is relatively clear : reach a certain amount of bytes, trigger the scrambler.
Changing this logic is possible, though it requires some care : rather than triggering the scrambler on consuming an entire block, this action should be considered on starting a new block. There is an additional complexity at the beginning of the stream, but let's assume a solution for it can be found.

A remaining problem is the second part of the change : the last stripe always use the "last secret".
Let's take an example, an input length is a clean multiple of 256 bytes.
Every 256 bytes, the internal loop consumes 4 stripes, using the "normal" secret.
At the very end, after consuming the last buffer,
it discovers that, actually, there is no additional byte.
So the last stripe was incorrectly calculated, using the wrong part of the `secret`.

In order to avoid this issue, the streaming function shall never trigger the internal loop as long as it's not sure that there is at least another byte to consume. Kind of a "lazy" decision.

This can probably be solved too. But noticed how we have been staking up additional little details, one complexity after another. The risk is that such details are not necessarily clear in the code, so they should be well documented. 

I will now see if the streaming logic can be adapted using a reasonable amount of complexity. Considering that the speed benefit is relatively narrow, the complexity cost should also remain reasonably low.

easyaspi314:
Well we could probably simplify a majority of the streaming logic already.

Cyan4973:
In #403, I went ahead, and applied the content of this PR,
plus additional changes required for the streaming API.

Good news is : it's not complex.
At the cost of one inherent fragility (`consumeStripes()` can only be invoked after guaranteeing that there is at least one more byte to consume afterwards, a condition the function itself cannot verify), which is fully internal and well documented, the rest of the streaming API is modified but not complexified. Mostly, a few test conditions must be `<` instead of `<=`.

So this is indeed good news. I believe this modification is a good candidate for a merge.

Cyan4973:
> Well we could probably simplify a majority of the streaming logic already.

If you believe the streaming logic can be made way simpler, by all means, this is good news, and I'll be certainly glad to review a PR on this topic.

Cyan4973:
Replaced by #403, currently merged into staging branch `newOutput074`

Issue #341, opened 2020-03-20
No wildcards on Windows CMD.EXE


easyaspi314:
### Hooray, another Windows bug.

The current `GetCommandLineW` method does not expand wildcards on `cmd.exe`:

v0.7.2:
```
C:\Code\xxHash>xxhsum.exe *.c
04504bbceb58a519  test.c
1eed5ddd8e02d18b  xxhash.c
45f1dab554385494  xxhsum.c
```
v0.7.3:
```
C:\Code\xxHash>xxhsum.exe *.c
Error: Could not open '*.c': Invalid argument.
```

Note that this is not a problem with `wmain`. 

I think the solution is to use [`__wgetmainargs`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/getmainargs-wgetmainargs?view=vs-2019) with `_DoWildCard = 1`.

This also avoids linking against `shell32.dll`, which is a lot of bloat:
`CommandLineToArgvW`:
```
$ ldd xxhsum.exe
        ntdll.dll => /c/Windows/SYSTEM32/ntdll.dll (0x7ffbf8140000)
        KERNEL32.DLL => /c/Windows/System32/KERNEL32.DLL (0x7ffbf7ef0000)
        KERNELBASE.dll => /c/Windows/System32/KERNELBASE.dll (0x7ffbf5ad0000)
        msvcrt.dll => /c/Windows/System32/msvcrt.dll (0x7ffbf7e50000)
        SHELL32.dll => /c/Windows/System32/SHELL32.dll (0x7ffbf69e0000)
        ucrtbase.dll => /c/Windows/System32/ucrtbase.dll (0x7ffbf5d80000)
        cfgmgr32.dll => /c/Windows/System32/cfgmgr32.dll (0x7ffbf5e80000)
        shcore.dll => /c/Windows/System32/shcore.dll (0x7ffbf79b0000)
        RPCRT4.dll => /c/Windows/System32/RPCRT4.dll (0x7ffbf77e0000)
        combase.dll => /c/Windows/System32/combase.dll (0x7ffbf6540000)
        bcryptPrimitives.dll => /c/Windows/System32/bcryptPrimitives.dll (0x7ffbf5ed0000)
        windows.storage.dll => /c/Windows/System32/windows.storage.dll (0x7ffbf5120000)
        msvcp_win.dll => /c/Windows/System32/msvcp_win.dll (0x7ffbf5f50000)
[...]

PR #346, opened 2020-03-28
Fix Unicode hopefully for the last time

easyaspi314:
Use `__wgetmainargs()` instead of `GetCommandLineW()` and `CommandLineToArgvW()`. While it is an internal API, aside from declaring the function, it actually simplifies things a lot, as we can setup a `wmain()` equivalent in a few lines.

It also avoids linking to `Shell32.dll`, which is a lot of bloat for a simple command line utility.

Most importantly, this supports wildcards on `cmd.exe`, fixing #341.

Also improved comments and did minor cleanup.

### DIFF
--- a/xxhsum.c
+++ b/xxhsum.c
@@ -124,8 +124,8 @@ static __inline int IS_CONSOLE(FILE* stdStream) {
 #  define S_ISREG(x) (((x) & S_IFMT) == S_IFREG)
 #endif
 
-/* Unicode helpers for Windows */
-#if defined(_WIN32)
+/* Unicode helpers for Windows to make UTF-8 act as it should. */
+#ifdef _WIN32
 /*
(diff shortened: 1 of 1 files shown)


easyaspi314:
Sorry about the slow response.

Cyan4973:
Nice effort @easyaspi314 !

I note that we have no test to check this feature in CI, hence it _may_ silently break in some future.
Not a deal breaker, but one property to consider when integrating this patch.

I believe you made everything in your power to deliver this capability as cleanly as possible.
Yet the code complexity it now requires is non-trivial, and may become source for concern for some future maintenance efforts.

I will nonetheless accept this PR, just keep in mind that if it has to become even more complex in the future, we may want to reconsider the complexity / benefit trade-off. Not all features are worth supporting. Maintenance drag is a very real effect for long-live project.

easyaspi314:
I hope it doesn't.

PR #347, opened 2020-04-06
add optimized implementation for AVX512 targets

gzm55:
on avx512 targets, variant with default secrets will improve 45%. The following result is compiled with `-Ofast -mavx512f`, and there is also no need to add `-mno-avx256-split-unaligned-load` with gcc.

compiling option:
- avx512 gcc-9/clang-9:  `-Ofast -mavx512f`
- avx2 gcc-9: `-Ofast -mavx2 -mno-avx256-split-unaligned-load`
- avx2 clang-9: `-Ofast -mavx2`

## compiled with Clang 9
```bash
./xxhsum 0.7.4 (64-bit x86_64 + AVX512 little endian), Clang 9.0.0 (tags/RELEASE_900/final), by Yann Collet
Sample of 100 KB...
XXH3_64b                      :     102400 ->   565524 it/s (55227.0 MB/s)
XXH3_64b unaligned            :     102400 ->   565409 it/s (55215.7 MB/s)
XXH3_64b w/seed               :     102400 ->   561795 it/s (54862.8 MB/s)
XXH3_64b w/seed unaligned     :     102400 ->   561556 it/s (54839.5 MB/s)
XXH3_64b w/secret             :     102400 ->   332811 it/s (32501.1 MB/s)
XXH3_64b w/secret unaligned   :     102400 ->   332829 it/s (32502.8 MB/s)
XXH128                        :     102400 ->   508060 it/s (49615.3 MB/s)
XXH128 unaligned              :     102400 ->   507962 it/s (49605.6 MB/s)
XXH128 w/seed                 :     102400 ->   505046 it/s (49320.9 MB/s)
XXH128 w/seed unaligned       :     102400 ->   504939 it/s (49310.5 MB/s)
XXH128 w/secret               :     102400 ->   345985 it/s (33787.6 MB/s)
XXH128 w/secret unaligned     :     102400 ->   345963 it/s (33785.4 MB/s)
```

```bash
./xxhsum 0.7.4 (64-bit x86_64 + AVX2 little endian), Clang 9.0.0 (tags/RELEASE_900/final), by Yann Collet
Sample of 100 KB...
XXH3_64b                      :     102400 ->   387469 it/s (37838.8 MB/s)
XXH3_64b unaligned            :     102400 ->   387931 it/s (37883.9 MB/s)
XXH3_64b w/seed               :     102400 ->   365748 it/s (35717.6 MB/s)
XXH3_64b w/seed unaligned     :     102400 ->   365647 it/s (35707.7 MB/s)
XXH3_64b w/secret             :     102400 ->   341313 it/s (33331.4 MB/s)
XXH3_64b w/secret unaligned   :     102400 ->   341394 it/s (33339.2 MB/s)
XXH128                        :     102400 ->   361423 it/s (35295.2 MB/s)
XXH128 unaligned              :     102400 ->   361373 it/s (35290.4 MB/s)
XXH128 w/seed                 :     102400 ->   331072 it/s (32331.3 MB/s)
XXH128 w/seed unaligned       :     102400 ->   331056 it/s (32329.7 MB/s)
XXH128 w/secret               :     102400 ->   314772 it/s (30739.4 MB/s)
XXH128 w/secret unaligned     :     102400 ->   314827 it/s (30744.8 MB/s)
```

## compiled with GCC 9

```bash
./xxhsum 0.7.4 (64-bit x86_64 + AVX512 little endian), GCC 9.2.1 20191102, by Yann Collet
Sample of 100 KB...
XXH3_64b                      :     102400 ->   512896 it/s (50087.5 MB/s)
XXH3_64b unaligned            :     102400 ->   512808 it/s (50078.9 MB/s)
XXH3_64b w/seed               :     102400 ->   521637 it/s (50941.1 MB/s)
XXH3_64b w/seed unaligned     :     102400 ->   520925 it/s (50871.6 MB/s)
XXH3_64b w/secret             :     102400 ->   341146 it/s (33315.0 MB/s)
XXH3_64b w/secret unaligned   :     102400 ->   341093 it/s (33309.8 MB/s)
XXH128                        :     102400 ->   462556 it/s (45171.5 MB/s)
XXH128 unaligned              :     102400 ->   462441 it/s (45160.3 MB/s)
XXH128 w/seed                 :     102400 ->   446254 it/s (43579.5 MB/s)
XXH128 w/seed unaligned       :     102400 ->   446616 it/s (43614.8 MB/s)
XXH128 w/secret               :     102400 ->   404667 it/s (39518.3 MB/s)
XXH128 w/secret unaligned     :     102400 ->   404753 it/s (39526.6 MB/s)
```

```bash
./xxhsum 0.7.4 (64-bit x86_64 + AVX2 little endian), GCC 9.2.1 20191102, by Yann Collet
Sample of 100 KB...
XXH3_64b                      :     102400 ->   365130 it/s (35657.2 MB/s)
XXH3_64b unaligned            :     102400 ->   363159 it/s (35464.7 MB/s)
XXH3_64b w/seed               :     102400 ->   364409 it/s (35586.8 MB/s)
XXH3_64b w/seed unaligned     :     102400 ->   364189 it/s (35565.3 MB/s)
XXH3_64b w/secret             :     102400 ->   330790 it/s (32303.7 MB/s)
XXH3_64b w/secret unaligned   :     102400 ->   330939 it/s (32318.3 MB/s)
XXH128                        :     102400 ->   333179 it/s (32537.0 MB/s)
XXH128 unaligned              :     102400 ->   333267 it/s (32545.6 MB/s)
XXH128 w/seed                 :     102400 ->   331603 it/s (32383.1 MB/s)
XXH128 w/seed unaligned       :     102400 ->   331731 it/s (32395.6 MB/s)
XXH128 w/secret               :     102400 ->   303619 it/s (29650.3 MB/s)
XXH128 w/secret unaligned     :     102400 ->   303526 it/s (29641.2 MB/s)
```

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -1346,7 +1346,7 @@ XXH3_accumulate(     xxh_u64* XXH_RESTRICT acc,
     for (n = 0; n < nbStripes; n++ ) {
         const xxh_u8* const in = input + n*STRIPE_LEN;
 #if (XXH_VECTOR == XXH_AVX512)
-	if (accWidth == XXH3_acc_64bits) XXH_PREFETCH(in + XXH_PREFETCH_DIST_AVX512_64);
+        if (accWidth == XXH3_acc_64bits) XXH_PREFETCH(in + XXH_PREFETCH_DIST_AVX512_64);
         else                             XXH_PREFETCH(in + XXH_PREFETCH_DIST_AVX512_128);
 #else
         XXH_PREFETCH(in + XXH_PREFETCH_DIST);
--- a/xxh3.h
+++ b/xxh3.h
@@ -1327,8 +1327,14 @@ XXH3_scrambleAcc(void* XXH_RESTRICT acc, const void* XXH_RESTRICT secret)
 }
 
 #define XXH_PREFETCH_DIST 384
-#define XXH_PREFETCH_DIST_AVX512_64  496
-#define XXH_PREFETCH_DIST_AVX512_128 384
+
+#ifdef __clang__ // for clang
+#  define XXH_PREFETCH_DIST_AVX512_64  320
+#  define XXH_PREFETCH_DIST_AVX512_128 320
+#else // for gcc
+#  define XXH_PREFETCH_DIST_AVX512_64  640
+#  define XXH_PREFETCH_DIST_AVX512_128 512
+#endif
 
 /*
  * XXH3_accumulate()
--- a/xxh3.h
+++ b/xxh3.h
@@ -163,9 +163,12 @@
 #define XXH_AVX2   2 /* AVX2 for Haswell and Bulldozer */
 #define XXH_NEON   3 /* NEON for most ARMv7-A and all AArch64 */
 #define XXH_VSX    4 /* VSX and ZVector for POWER8/z13 */
+#define XXH_AVX512 5 /* AVX512 for Skylake and Icelake */
 
 #ifndef XXH_VECTOR    /* can be defined on command line */
-#  if defined(__AVX2__)
+#  if defined(__AVX512F__)
+#    define XXH_VECTOR XXH_AVX512
+#  elif defined(__AVX2__)
 #    define XXH_VECTOR XXH_AVX2
 #  elif defined(__SSE2__) || defined(_M_AMD64) || defined(_M_X64) || (defined(_M_IX86_FP) && (_M_IX86_FP == 2))
 #    define XXH_VECTOR XXH_SSE2
@@ -198,6 +201,8 @@
 #     define XXH_ACC_ALIGN 16
 #  elif XXH_VECTOR == XXH_VSX   /* vsx */
 #     define XXH_ACC_ALIGN 16
+#  elif XXH_VECTOR == XXH_AVX512 /* avx512 */
+#     define XXH_ACC_ALIGN 64
 #  endif
 #endif
 
@@ -930,7 +935,37 @@ XXH3_accumulate_512(      void* XXH_RESTRICT acc,
                     const void* XXH_RESTRICT secret,
                     XXH3_accWidth_e accWidth)
 {
-#if (XXH_VECTOR == XXH_AVX2)
+#if (XXH_VECTOR == XXH_AVX512)
+
+    XXH_ASSERT((((size_t)acc) & 63) == 0);
+    XXH_STATIC_ASSERT(STRIPE_LEN == sizeof(__m512i));
+    {   XXH_ALIGN(64) __m512i* const xacc    =       (__m512i *) acc;
+
+        /* data_vec    = input[0]; */
+        __m512i const data_vec    = _mm512_loadu_si512   (input);
+        /* key_vec     = secret[0]; */
+        __m512i const key_vec     = _mm512_loadu_si512   (secret);
+        /* data_key    = data_vec ^ key_vec; */
+        __m512i const data_key    = _mm512_xor_si512     (data_vec, key_vec);
+        /* data_key_lo = data_key >> 32; */
+        __m512i const data_key_lo = _mm512_shuffle_epi32 (data_key, _MM_SHUFFLE(0, 3, 0, 1));
+        /* product     = (data_key & 0xffffffff) * (data_key_lo & 0xffffffff); */
+        __m512i const product     = _mm512_mul_epu32     (data_key, data_key_lo);
+        if (accWidth == XXH3_acc_128bits) {
+            /* xacc[0] += swap(data_vec); */
+            __m512i const data_swap = _mm512_shuffle_epi32(data_vec, _MM_SHUFFLE(1, 0, 3, 2));
+            __m512i const sum       = _mm512_add_epi64(*xacc, data_swap);
+            /* xacc[0] += product; */
+            *xacc = _mm512_add_epi64(product, sum);
+        } else {  /* XXH3_acc_64bits */
+            /* xacc[0] += data_vec; */
+            __m512i const sum = _mm512_add_epi64(*xacc, data_vec);
+            /* xacc[0] += product; */
+            *xacc = _mm512_add_epi64(product, sum);
+        }
+    }
+
+#elif (XXH_VECTOR == XXH_AVX2)
 
     XXH_ASSERT((((size_t)acc) & 31) == 0);
     {   XXH_ALIGN(32) __m256i* const xacc    =       (__m256i *) acc;
@@ -1118,7 +1153,29 @@ XXH3_accumulate_512(      void* XXH_RESTRICT acc,
 XXH_FORCE_INLINE void
 XXH3_scrambleAcc(void* XXH_RESTRICT acc, const void* XXH_RESTRICT secret)
 {
-#if (XXH_VECTOR == XXH_AVX2)
+#if (XXH_VECTOR == XXH_AVX512)
+
+    XXH_ASSERT((((size_t)acc) & 63) == 0);
+    XXH_STATIC_ASSERT(STRIPE_LEN == sizeof(__m512i));
+    {   XXH_ALIGN(64) __m512i* const xacc = (__m512i*) acc;
+        const __m512i prime32 = _mm512_set1_epi32((int)PRIME32_1);
+
+        /* xacc[0] ^= (xacc[0] >> 47) */
+        __m512i const acc_vec     = *xacc;
+        __m512i const shifted     = _mm512_srli_epi64    (acc_vec, 47);
+        __m512i const data_vec    = _mm512_xor_si512     (acc_vec, shifted);
+        /* xacc[0] ^= secret; */
+        __m512i const key_vec     = _mm512_loadu_si512   (secret);
+        __m512i const data_key    = _mm512_xor_si512     (data_vec, key_vec);
+
+        /* xacc[0] *= PRIME32_1; */
+        __m512i const data_key_hi = _mm512_shuffle_epi32 (data_key, _MM_SHUFFLE(0, 3, 0, 1));
+        __m512i const prod_lo     = _mm512_mul_epu32     (data_key, prime32);
+        __m512i const prod_hi     = _mm512_mul_epu32     (data_key_hi, prime32);
+        *xacc = _mm512_add_epi64(prod_lo, _mm512_slli_epi64(prod_hi, 32));
+    }
+
+#elif (XXH_VECTOR == XXH_AVX2)
 
     XXH_ASSERT((((size_t)acc) & 31) == 0);
     {   XXH_ALIGN(32) __m256i* const xacc = (__m256i*) acc;
@@ -1270,6 +1327,8 @@ XXH3_scrambleAcc(void* XXH_RESTRICT acc, const void* XXH_RESTRICT secret)
 }
 
 #define XXH_PREFETCH_DIST 384
+#define XXH_PREFETCH_DIST_AVX512_64  496
+#define XXH_PREFETCH_DIST_AVX512_128 384
 
 /*
  * XXH3_accumulate()
@@ -1286,7 +1345,12 @@ XXH3_accumulate(     xxh_u64* XXH_RESTRICT acc,
     size_t n;
     for (n = 0; n < nbStripes; n++ ) {
         const xxh_u8* const in = input + n*STRIPE_LEN;
+#if (XXH_VECTOR == XXH_AVX512)
+	if (accWidth == XXH3_acc_64bits) XXH_PREFETCH(in + XXH_PREFETCH_DIST_AVX512_64);
+        else                             XXH_PREFETCH(in + XXH_PREFETCH_DIST_AVX512_128);
+#else
         XXH_PREFETCH(in + XXH_PREFETCH_DIST);
+#endif
         XXH3_accumulate_512(acc,
                             in,
                             secret + n*XXH_SECRET_CONSUME_RATE,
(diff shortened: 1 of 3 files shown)


easyaspi314:
> on avx512 targets, variant with default secrets will improve 45%. The following result is compiled with `-Ofast -mavx512`, and there is also no need to add `-mno-avx256-split-unaligned-load` with gcc.

Those benchmarks are made with Clang, which likes to impersonate `gcc` and MSVC. 

Notably, FreeBSD, macOS, the Android NDK, and Termux all do this by default.

gzm55:
the previous benchmarks are compiled via `CC=clang-9 make` on a Ubuntu 18.04 machine, the GCC compiled version is also tested with a similar result.

For the `-mno-avx256-split-unaligned-load` option, the real GCC is not affected any more since that is an `avx2 only` option.

gzm55:
The benchmark results compiled by gcc is updated.

It is wired that custom secret will significantly affects performance, the clang-compiled avx512 `XXH3_64b w/secret` even slower than avx2.

easyaspi314:
Clang doesn't appear to even *try* to unroll the `withSecret` variants.

gzm55:
gcc also has the similar performance on the `withSecret` ones.

gzm55:
@easyaspi314 the main factor seems to be the runtime `secretSize` which blocks unrolling in the internal loop. A `switch` table works around for 6 align size between [136,256]:

```c
XXH_NO_INLINE XXH64_hash_t
XXH3_hashLong_64b_withSecret(const xxh_u8* XXH_RESTRICT input, size_t len,
                             const xxh_u8* XXH_RESTRICT secret, size_t secretSize)
{
    switch (secretSize) {
    case XXH3_SECRET_SIZE_MIN + 0 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN + 0 * 8);
    case XXH3_SECRET_SIZE_MIN + 1 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN + 1 * 8);
    case XXH3_SECRET_SIZE_MIN + 2 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN + 2 * 8);
    case XXH3_SECRET_SIZE_MIN + 3 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN + 3 * 8);
    case XXH3_SECRET_SIZE_MIN + 4 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN + 4 * 8);
    case XXH3_SECRET_SIZE_MIN + 5 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN + 5 * 8);
    case XXH3_SECRET_SIZE_MIN + 6 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN + 6 * 8);
    case XXH3_SECRET_SIZE_MIN + 7 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN + 7 * 8);
    case XXH3_SECRET_SIZE_MIN + 8 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN + 8 * 8);
    case XXH3_SECRET_SIZE_MIN + 9 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN + 9 * 8);
    case XXH3_SECRET_SIZE_MIN +10 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN +10 * 8);
    case XXH3_SECRET_SIZE_MIN +11 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN +11 * 8);
    case XXH3_SECRET_SIZE_MIN +12 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN +12 * 8);
    case XXH3_SECRET_SIZE_MIN +13 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN +13 * 8);
    case XXH3_SECRET_SIZE_MIN +14 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN +14 * 8);
    case XXH3_SECRET_SIZE_MIN +15 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN +15 * 8);
    }
    return XXH3_hashLong_64b_internal(input, len, secret, secretSize);
}
```

easyaspi314:
What about this?

```c
#if XXH_VECTOR == XXH_AVX512 && !XXH_REROLL
    if (XXH_likely(nbStripes % 4 == 0)) {
        for (n = 0; n < nbStripes; n+=4 ) {
            const xxh_u8* const in = input + n*STRIPE_LEN;
            if (accWidth == XXH3_acc_64bits) XXH_PREFETCH(in + XXH_PREFETCH_DIST_AVX512_64);
            else                             XXH_PREFETCH(in + XXH_PREFETCH_DIST_AVX512_128);
            XXH3_accumulate_512(acc,
                                in,
                                secret + n*XXH_SECRET_CONSUME_RATE,
                                accWidth);
            XXH3_accumulate_512(acc,
                                in + 1 * STRIPE_LEN,
                                secret + (n+1)*XXH_SECRET_CONSUME_RATE,
                                accWidth);
            XXH3_accumulate_512(acc,
                                in + 2 * STRIPE_LEN,
                                secret + (n+2)*XXH_SECRET_CONSUME_RATE,
                                accWidth);
            XXH3_accumulate_512(acc,
                                in + 3 * STRIPE_LEN,
                                secret + (n+3)*XXH_SECRET_CONSUME_RATE,
                                accWidth);
        }
    } else
#endif
    for (n = 0; n < nbStripes; n++) {
        ...
```

xxh3.h
```diff
@@ -1270,6 +1327,8 @@ XXH3_scrambleAcc(void* XXH_RESTRICT acc, const void* XXH_RESTRICT secret)
 }
 
 #define XXH_PREFETCH_DIST 384
+#define XXH_PREFETCH_DIST_AVX512_64  496
```
Cyan4973, line 1330: I'm surprised this distance, 496, is not a multiple of 64.
What about 512 ? or 448 ?

My understanding is that, for a distance 496, and presuming a perfectly aligned input on 64-bytes boundaries, 496 is effectively equivalent to 448. Now, for the more general case of non-aligned input, 496 will sometimes be equivalent to 448, sometimes equivalent to 512, depending on the specific `input_address % 64` value.
And I'm not sure if this behavior is really a benefit.

Cyan4973:
Thanks for this great submission @gzm55 !

I understand it noticeably improves speed when using the default `secret` for CPU with AVX512 instruction set. +40% definitely matches the bar in order to introduce a new target-specific code path in the library.

I also understand that optimizing the `_withSecret` variant is more complex.
That being said, as a way to keep this PR clean and simple, I'm happy to merge it if you wish, and accept a follow-up with improvements on the `_withSecret` variant.

I merely have a small question on the new prefetch distance.

Cyan4973:
> Clang doesn't appear to even try to unroll the withSecret variants, so there is a lot of branching overhead.

Was it tested with `XXH_INLINE_ALL` ?
I naively assumed that, if the `secret` is known at compile time, then the compiler would find a way to use the `secret`'s length to correctly optimize the inner loop, even if it's not the default one.

xxh3.h
gzm55: according to [this](https://www.cc.gatech.edu/~hyesoon/lee_taco12.pdf), the prefetch distance have to be profiled on the benchmark machine (2 core ECS on Aliyun, Cannon lake CPU). I tested from 128 to 512, it seems 496 is the best.

gzm55:
@Cyan4973 agree to separate the `_withSecret` discussion and optimization to another PR.

the distance of prefetching is got from profiling each one byte for 100K length input (xxhsum default). The benchmark machine is a ECS server, not a desktop. Do you have some experience or environments to settle down a better distance?

xxh3.h
Cyan4973: > I tested from 128 to 512, it seems 496 is the best.

Well, I have no access to the program you used to measure this value.
But I would expect the outcome of this program to be subtly impacted by input's alignment.
I would actually expect this input to feature `start_address % 64 == 16`.

Try multiple different `start_address`.

Cyan4973:
> Do you have some experience or environments to settle down a better distance?

There will never be a "one size fits all" best prefetch distance for all systems.
The "better" distance is necessarily system dependent.

Some generic rules of thumb to establish a distance : 
- Currently, this is a "better than nothing" approach; we want a "generic" value which is good enough to improve performance across the board. This is different from looking for the _best_ value for each system.
- In the future, we might have to offer some programmability on this value. First compile time, then maybe runtime. We don't have to do that now, this is ground for some future patch.
- Be wary of synthetic benchmark for this topic. A benchmark will basically assume that it has the whole system for itself. This will lead to over-use of memory bandwidth and cache size, on the ground that they improve the synthetic benchmark. But in a real application, bandwidth will be shared with other parts of the application, and over-using it with over-demanding prefetch requests will lead to system-wide performance degradation. When selecting a prefetching distance, one wants to lean towards _smaller_ distances, and basically stop early, as soon as benefits from larger prefetching distances become "small". That's because it preserve a general resource (memory bandwidth) for the benefit of the larger system, be it other parts of the application or other concurrent applications.

xxh3.h
gzm55: ok, i will try this today.

gzm55:
@Cyan4973 the prefetch distances are updated to 64 aligned for clang and gcc. The results:

### Clang
```bash
./xxhsum 0.7.4 (64-bit x86_64 + AVX512 little endian), Clang 9.0.0 (tags/RELEASE_900/final), by Yann Collet
Sample of 100 KB...
XXH3_64b                      :     102400 ->   564617 it/s (55138.4 MB/s)
XXH3_64b unaligned            :     102400 ->   564709 it/s (55147.3 MB/s)
XXH3_64b w/seed               :     102400 ->   561776 it/s (54860.9 MB/s)
XXH3_64b w/seed unaligned     :     102400 ->   561770 it/s (54860.4 MB/s)
XXH3_64b w/secret             :     102400 ->   332788 it/s (32498.8 MB/s)
XXH3_64b w/secret unaligned   :     102400 ->   332700 it/s (32490.2 MB/s)
XXH128                        :     102400 ->   508311 it/s (49639.7 MB/s)
XXH128 unaligned              :     102400 ->   508230 it/s (49631.9 MB/s)
XXH128 w/seed                 :     102400 ->   505657 it/s (49380.5 MB/s)
XXH128 w/seed unaligned       :     102400 ->   505754 it/s (49390.1 MB/s)
XXH128 w/secret               :     102400 ->   346498 it/s (33837.7 MB/s)
XXH128 w/secret unaligned     :     102400 ->   346553 it/s (33843.1 MB/s)
```

### Gcc
```bash
./xxhsum 0.7.4 (64-bit x86_64 + AVX512 little endian), GCC 9.2.1 20191102, by Yann Collet
Sample of 100 KB...
XXH3_64b                      :     102400 ->   528095 it/s (51571.8 MB/s)
XXH3_64b unaligned            :     102400 ->   528163 it/s (51578.4 MB/s)
XXH3_64b w/seed               :     102400 ->   528725 it/s (51633.3 MB/s)
XXH3_64b w/seed unaligned     :     102400 ->   528844 it/s (51645.0 MB/s)
XXH3_64b w/secret             :     102400 ->   341283 it/s (33328.4 MB/s)
XXH3_64b w/secret unaligned   :     102400 ->   341335 it/s (33333.5 MB/s)
XXH128                        :     102400 ->   472524 it/s (46144.9 MB/s)
XXH128 unaligned              :     102400 ->   472535 it/s (46146.0 MB/s)
XXH128 w/seed                 :     102400 ->   465276 it/s (45437.1 MB/s)
XXH128 w/seed unaligned       :     102400 ->   465038 it/s (45413.8 MB/s)
XXH128 w/secret               :     102400 ->   404590 it/s (39510.7 MB/s)
XXH128 w/secret unaligned     :     102400 ->   404518 it/s (39503.7 MB/s)
```

easyaspi314:
After some testing, I don't think prefetching on aarch64 is worth it.

It ends up being slower.

xxh3.h
```diff
@@ -1271,6 +1328,14 @@ XXH3_scrambleAcc(void* XXH_RESTRICT acc, const void* XXH_RESTRICT secret)
 
 #define XXH_PREFETCH_DIST 384
 
+#ifdef __clang__ // for clang
```
Cyan4973, line 1331: As a quick comment,
I'm surprised to notice that the optimal distance seems to depend on the compiler.
I guess it implies that generated assembly is different enough to warrant such a distinction.

Cyan4973:
> After some testing, I don't think prefetching on aarch64 is worth it.
> It ends up being slower.

This is a surprising outcome, though one certainly worth investigating.

I don't know how `aarch64` prefetching works, but I guess one mechanism probably exists, and specificities are supposed to be abstracted behind `gcc` intrinsics.
Maybe there are not.

Anyway, selecting different prefetching policy depending on target hardware is certainly within the scope of this project.

xxh3.h
gzm55: yes, i think each compiler will have very different loop cycles, the slower loop will correspond the larger distance, so clang < gcc 64b < gcc 128b.

gzm55:
> Anyway, selecting different prefetching policy depending on target hardware is certainly within the scope of this project.

While searching the prefetch dist for avx512, I think we should define a matrix over (target, compiler, accWidth) for the dist values or disable prefetch.

PR #350, opened 2020-04-08
refine prefetch for internal loop

gzm55:
Refine the prefetch distances skeleton, preserve the previous results and open the slots for other compilers and archs. Also disable prefetch for `aarch64` according to @easyaspi314's [testing](https://github.com/Cyan4973/xxHash/pull/347#issuecomment-610542930).

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -1336,6 +1336,54 @@ XXH3_scrambleAcc(void* XXH_RESTRICT acc, const void* XXH_RESTRICT secret)
 #  define XXH_PREFETCH_DIST_AVX512_128 512
 #endif
 
+/* The prefetch feature will be affected my many factors. For internal loop, we control it by three
+ * macros:
+ *   - XXH_PREFETCH
+ *   - XXH_PREFETCH_DIST_64
+ *   - XXH_PREFETCH_DIST_128
+ *
+ * Given XXH_PREFETCH is enabled, here we can select proper prefetch distances at compile time over
+ * (compiler, vector modes, target hardwares). Each group distances should be selected over
+ * profiling experiments. And we still have some principles:
+ *   - the base line comes from disabling prefetch
+ *   - since the internal loop step is forward, the distances should be non-negative
+ *   - distances shoule be aligned to the common cache line size of the target hardwares
+ *   - lean towards smaller distances, and basically stop early, as soon as benefits from larger
+ *     prefetching distances start to become "small". That's because it preserves a general resource
+ *     (memory bandwidth) for the benefit of the larger system, be it other parts of the application
+ *     or other concurrent applications.
+ */
+#if defined(_MSC_VER) // MSVC
+// use common settings
+#  define XXH_PREFETCH_DIST_64  384
+#  define XXH_PREFETCH_DIST_128 384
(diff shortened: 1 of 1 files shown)


easyaspi314:
Too complex, and the fact that the numbers differ between compilers makes little sense aside from one generating better code. 

As for aarch64:
```c
#ifndef XXH_NO_PREFETCH
#  if defined(__aarch64__)
#    define XXH_NO_PREFETCH 1
#  else
#    define XXH_NO_PREFETCH 0
#  endif
#endif
```

gzm55:
@easyaspi314 prefetch is disabled for all aarch64. And I think the result should be based on the testing results in each environment.

Cyan4973:
I'm still surprised that prefetching hurts performance on `aarch64`.
We have good results using prefetching with `zstd` on `aarch64`, so we know it can be beneficial.
Furthermore, `xxHash` access pattern is very simple, and the benefits of prefetching look very straightforward for this scenario. Hence I don't see how it could "hurt".

That being said, if prefetching crashes the algorithm on `aarch64`, then it's certainly ground for disabling it. That's really a strange outcome. Prefetching is presumed always valid, even when the memory address is invalid. This should basically rule out the possibility of a crash by prefetching...

https://gcc.gnu.org/projects/prefetch.html

easyaspi314:
> That being said, if prefetching crashes the algorithm on aarch64, then it's certainly ground for disabling it. That's really a strange outcome. 

No, it is only on the `avr` and `msp430` backends. However, they have other issues, primarily due to being 8-bit and 16-bit respectively, and, well, experimental.

kevattor:
> aarch64

xxh3.h
```diff
@@ -1336,6 +1336,48 @@ XXH3_scrambleAcc(void* XXH_RESTRICT acc, const void* XXH_RESTRICT secret)
...
+#  if (XXH_VECTOR == XXH_AVX512) // avx implies x86
+#    define XXH_PREFETCH_DIST_64  320
+#    define XXH_PREFETCH_DIST_128 320
+#  else
+#    define XXH_PREFETCH_DIST_64  384
+#    define XXH_PREFETCH_DIST_128 384
+#  endif
+#elif defined(__GNUC__) && ! defined(__clang__) // GCC
+#  if (XXH_VECTOR == XXH_AVX512) // avx implies x86
+#    define XXH_PREFETCH_DIST_64  640
```
Cyan4973, line 1370: Is it still valuable to separate `XXH_PREFETCH_DIST_64` from `XXH_PREFETCH_DIST_128` if the only difference is here ?
It seems this complexifies the code, for little gains.

Cyan4973:
Thanks for the patch @gzm55 .
I've no doubt that it works and is well thought.
At this stage in the development process though, I will rather aim at reducing some of the complexities.
To this end, I will simplify the part related to prefetch distance, conflicting with this PR.

One lesson I keep from this proposal is that optimal prefetch distance optimal varies depending on systems and compilation settings. Therefore, prefetch distance should be a controllable value, allowing experiments on various systems. So I'll make the prefetch distance optionally selectable at compilation time. Eventually, in the future, this will allow some finer grained selection, and it will make sense to optimize that part again.

PR #351, opened 2020-04-10
vectorize XXH3_initCustomSecret for x86 arch

gzm55:
Vectorize `XXH3_initCustomSecret()` function  on sse2/avx2/avx512 arch, improve the seeded variants especially for shorter keys.

### Test config

The pre-processor output for scalar and non-x86 variants should not change, so only test  sse2, avx2 and avx512 vector modes. `XXH3_initCustomSecret()` is called once at the begin of xx_withSeed public APIs, that is, the shorter keys are more benefit, therefore the benchmark covers the length of 256, 521 and 100K. The unaligned input benchmarks are omitted, only include aligned `XXH3_64b w/seed` and `XXH128 w/seed` methods.

Benchmark environments:

- Machine A: X64 avx2, CentOS 8.1, gcc 9.1.1 and clang 8.0.1
- Machine B: X64 avx512, Ubuntu 18.04, gcc 9.2.1 and clang 9.0.0

Compiling options:
 - `-m<arch>` for arch in sse2, avx2 and avx512f
 - `-Ofast`
 - `-DXXH_INLINE_ALL`
 - `-mno-avx256-split-unaligned-load` for gcc + avx2

Testing methods: run `xxhsum -b -B<length> -i5` three or four times, select the stable high results (roughly P95).

Baseline commit: Cyan4973/xxHash@dda230e

### Benchmarks:

|Box|CC|SIMD|XXH3 Len=256|XXH128 Len=256|XXH3 Len=521|XXH128 Len=521|XXH3 Len=100K|XXH128 Len=100K|
| :---: | :---: | :---: | --- | --- | --- | --- | --- | --- |
|A|GCC 9|SSE2|`7533.9 -> 7593.5` +0.8% | `6008.2 -> 6041.3` +0.6% | `9690.5 -> 9979.8` +3%|`8233.9 -> 8237.5` + 0.04%|`14737.7 -> 14666.3` -0.5%|`13730.6 -> 13919.9` +1.4%|
|A|CLANG8|SSE2|`6660.5 -> 8578.5` +28.8% | `5186.0 -> 6262.9` +20.8% | `9609.6 -> 11280.6` +17.4% |`7574.2 -> 8414.6` +11.1%|`15661.3 -> 15630.8` -0.2%|`14037.6 -> 13957.1` -0.6%|
|B|GCC 9|SSE2|`9529.9 -> 9777.5` +2.6% | `7383.4 -> 7466.9` +1.1% | `12380.0 -> 12525.3` +1.2%|`9865.7 -> 9840.8` -0.3%|`18083.9 -> 18089.7` +0.03%|`16991.9 -> 16995.4` +0.02%|
|B|CLANG9|SSE2|`9450.0 -> 10957.7` +16.0% | `7192.5 -> 8006.8` +11.3% | `12780.4 -> 13836.9` +8.3%|`10017.5 -> 10890.1` +8.7%|`20895.7 -> 20923.1` +0.1%|`18567.8 -> 18629.2` +0.3%|
|A|GCC 9|AVX2|`6951.1 -> 8059.3` +15.9% | `5712.1 -> 6536.8` +14.4% | `10740.0 -> 12142.0` +13.1% | `9107.4 -> 10099.7` +10.9%| `23156.7 -> 23252.1` +0.4% | `20931.7 -> 21056.4` +0.6%|
|A|CLANG8|AVX2|`7902.0 -> 9041.8` +14.4% | `6156.8 -> 6851.4` +11.3% | `12672.5 -> 14216.6` +12.2% | `10049.8 -> 11247.1` + 11.9%| `25870.1 -> 25759.3` -0.4% | `23988.6 -> 23871.1` -0.5%|
|B|GCC 9|AVX2|`8068.3 -> 10047.4` +24.5% | `6797.7 -> 8149.1` +19.9% | `13487.6 -> 15777.9` +17.0%|`11144.8 -> 12741.4` +14.3%|`34737.1 -> 34738.7` +0.005%|`31703.3 -> 31630.2` -0.2%|
|B|CLANG9|AVX2|`11070.9 -> 11027.0` -0.4% | `8480.9 -> 8528.9` +0.6% | `18046.5 -> 18153.8` +0.6%|`14345.6 -> 14399.9` +0.4%|`36038.9 -> 35802.3` -0.7%|`32581.3 -> 32495.7` -0.3%|
|B|GCC 9|AVX512|`7913.8 -> 8171.4` +3.3% | `6576.6 -> 8149.1` +23.9% | `12404.5 -> 12712.8` +2.5%|`10943.0 -> 11119.4` +1.6%|`51631.9 -> 51542.4` -0.2%|`44612.3 -> 45509.9` +2%|
|B|CLANG9|AVX512|`9151.3 -> 9656.0` +5.5% | `7259.4 -> 7542.9` +3.9% | `14812.3 -> 15593.5` +5.3% |`11742.4 -> 12494.4` +6.4%|`54842.7 -> 53182.4` -0.3%|`49400.0 -> 48802.1` -1.2%|

#### Note

Clang 9 will auto vectorize the scalar code, so the improvement from Clang 9 on AVX2 arch is indistinctive.


xxh3.h
```diff
@@ -1509,18 +1506,47 @@ XXH_FORCE_INLINE void XXH3_initCustomSecret(xxh_u8* XXH_RESTRICT customSecret, x
...
+        XXH_ALIGN(64)       __m256i* const dest = (      __m256i*) customSecret;
+                            __m256i  const seed = { (long)seed64, -(long)seed64, (long)seed64, -(long)seed64 };
+        for (i=0; i < nbRounds; ++i) dest[i] = _mm256_add_epi64(src[i], seed);
+    }
+#elif (XXH_VECTOR == XXH_SSE2)
+    XXH_STATIC_ASSERT((XXH_SECRET_DEFAULT_SIZE & 15) == 0);
+    {   int const nbRounds = XXH_SECRET_DEFAULT_SIZE / sizeof(__m128i);
+        XXH_ALIGN(64) const __m128i* const src  = (const __m128i*) kSecret;
+        XXH_ALIGN(64)       __m128i* const dest = (      __m128i*) customSecret;
+                            __m128i  const seed = { (long)seed64, -(long)seed64 };
```
easyaspi314, line 1531: Non-standard extension, will truncate seed64 on Windows or 32-bit

Cyan4973: Indeed, the `long` type is problematic, as its size, 32 or 64-bit, can be different depending on platform.
So it is best avoided.

xxh3.h
```diff
@@ -1463,16 +1463,13 @@ XXH_FORCE_INLINE void XXH_writeLE64(void* dst, xxh_u64 v64)
...
 {
-    int const nbRounds = XXH_SECRET_DEFAULT_SIZE / 16;
     int i;
     /*
      * We need a separate pointer for the hack below.
      * Any decent compiler will optimize this out otherwise.
      */
     const xxh_u8 *kSecretPtr = kSecret;
 
-    XXH_STATIC_ASSERT((XXH_SECRET_DEFAULT_SIZE & 15) == 0);
```
Cyan4973, line 1474: Why removing this assertion ?

xxh3.h
```diff
@@ -1559,7 +1585,7 @@ XXH3_hashLong_64b_withSecret(const xxh_u8* XXH_RESTRICT input, size_t len,
 XXH_NO_INLINE XXH64_hash_t
 XXH3_hashLong_64b_withSeed(const xxh_u8* input, size_t len, XXH64_hash_t seed)
 {
-    XXH_ALIGN(8) xxh_u8 secret[XXH_SECRET_DEFAULT_SIZE];
```
Cyan4973, line 1562: Note : I remember having measured some unexpected performance loss when increasing alignment requirement of this array. I could not explain it. But that's the reason why alignment remained at `8`.
_Maybe_ situation has changed, and this problem is no longer present, 
just present a set of benchmark to illustrate that outcome.

Cyan4973:
Please provide some benchmark results to back up performance claims.

Beyond the cases where performance is improved, notably AVX variants, 
please also benchmark variants which benefit less (SSE, Scalar code, ARM platform if possible). 
We want to ensure that the proposed change is not detrimental to them.

xxh3.h
easyaspi314: I want to go back in time and slap whoever thought up with making type sizes platform-dependent.

Java got it right.

C did not.

| Type | \*nix 64-bit | Windows 64-bit | 32-bit | 8-bit/16-bit |
|:----|---:|---:|---:|----:|
| `char` | 8 | 8 | 8 | 8 |
| `short` | 16 | 16 | 16 | 16 |
| `int` | 32 | 32 | 32 | 16 |
| `long` | 64 | 32 | 32 | 32 |
| `long long` | 64 | 64 | 64 | 64 |

Also add `char` being signed on x86 and unsigned on ARM/PPC/MIPS, and the standard type system is useless.

`stdint.h` or `limits.h` for C89 are the only reliable options.

gzm55: yes, c type system is so messy~ I add a `xxh_i64` for portable `int64_t`.

gzm55:
> Please provide some benchmark results to back up performance claims.
> 
> Beyond the cases where performance is improved, notably AVX variants,
> please also benchmark variants which benefit less (SSE, Scalar code, ARM platform if possible).
> We want to ensure that the proposed change is not detrimental to them.


The shorter input will benefit from this patch since the initial part is call once, i will sort out the benchmark results.

easyaspi314:
MSVC doesn't support brace initialization for SSE types, that is a GCC extension.

gzm55:
@easyaspi314 OK, will change it to _mm_load_si128.

xxh3.h
gzm55: on x86/x64, alignment greater than native stack alignment (16-byte for x64) will introduce an `and` instruction in the prolog of a function.

easyaspi314: That shouldn't make any measurable difference, aside from maybe requiring a frame pointer, which x86 is optimized to set up quickly

gzm55: on x64, alignment greater than 16 will also introduce `push rsp` and `pop rsp` instructions, these mem read/write operations seems measurable on short keys.

gzm55:
> Please provide some benchmark results to back up performance claims.
> 
> Beyond the cases where performance is improved, notably AVX variants,
> please also benchmark variants which benefit less (SSE, Scalar code, ARM platform if possible).
> We want to ensure that the proposed change is not detrimental to them.

@Cyan4973 @easyaspi314 the benchmark is update in the description. Scalar and ARM platform is not tested since the changes do not affect those parts.

easyaspi314:
I have tested previously, and NEON only improves things slightly on 32-bit, while being detrimental on 64-bit. Clang already autovectorizes it on 32-bit (albeit not in the most efficient way), so I am not too worried about that.

gzm55:
> I have tested previously, and NEON only improves things slightly on 32-bit, while being detrimental on 64-bit. Clang already autovectorizes it on 32-bit (albeit not in the most efficient way), so I am not too worried about that.

Clang 9 is really clever, with the some hint, it generate efficient SSE2/AVX2 instructions(add+mov sequence) for scalar loops.

```c
    __m128i const seeds = { (long long)seed64, -(long long)seed64 };
    const xxh_u64 *seedsPtr = (const void*)&seeds;
    __asm__("" : "+r" (seedsPtr));
    for (i=0; i < nbRounds; i++) {
        xxh_u64 lo = XXH_readLE64(kSecretPtr + 16*i)     + seedsPtr[0];
        xxh_u64 hi = XXH_readLE64(kSecretPtr + 16*i + 8) + seedsPtr[1];
        XXH_writeLE64(customSecret + 16*i,     lo);
        XXH_writeLE64(customSecret + 16*i + 8, hi);
    }
```

The older clang or gcc have to be vectorized by hand.

Cyan4973:
I'm slightly surprised that the benchmark seems to report a performance _loss_ on some _big_ inputs (100 KB). I would expect this patch to be more or less insensible at such length. Is that just a measurement noise ?

Cyan4973:
FYI, as I'm benchmarking this patch,

currently using `gcc` v9.3.0 with default `SSE2`,
and I'm observing something strange : speed of `XXH32`, and even `XXH64`, are _reduced_ after applying the patch.

`XXH32` : 8900 -> 7350
`XXH64` : 13400 -> 12900

This is strange, as the patch does not seem to impact directly `XXH32` nor `XXH64`. But I can reproduce it pretty consistently, and the drop is much larger than noise level.

gzm55:
another possible cause: `__asm__("" : "+r" (dest));` reduces occupied registers, this can avoid saving registers to stack, but have to read the `secret` from memory, so loss some performance for long input.

For `XXH32` and `XXH64`, is it possible affected by the executable size?

xxhash.h
```diff
@@ -1452,6 +1452,14 @@ XXH_PUBLIC_API XXH32_hash_t XXH32_hashFromCanonical(const XXH32_canonical_t* src
 
 typedef XXH64_hash_t xxh_u64;
 
+#if !defined (__VMS) \
+  && (defined (__cplusplus) \
+  || (defined (__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L) /* C99 */) )
+    typedef int64_t xxh_i64;
```
Cyan4973, line 1458: I presume this type is declared solely for the purpose of compatibility with `_mm512_set1_epi64()` and `_mm256_set_epi64x()`.
But the official type (in Intel documentation) is `__int64`, and I presume it's necessarily defined when the intrinsics are defined. 
Consequently, I believe there is no need to create an intermediate type, just cast to `__int64` directly in the function invocation.

Cyan4973:
> another possible cause: `__asm__("" : "+r" (dest));` 

But this line is not supposed to be applicable for `XXH32` nor `XXH64` ...

Cyan4973:
I'm considering that the performance impact on `XXH32` and `XXH64` might be indirect, related to instruction alignment. This is inherently unpredictable (that is, as long as compilers do not pay better attention to this topic), as it changes with compiler version, and with the rest of the program.

Cyan4973:
Performance benchmark, focusing on `_withSeed()` variant : 

Aligned input of 256 bytes, `XXH3_64bits_withSeed()`, `SSE2` : 
`clang` v9.0.1 : 14.2 -> 16.3
`clang` v10.0.0 : 13.5 -> 16.3
`gcc` v7.5.0 : 12.9 -> 13.9
`gcc` v8.4.0 : 13.0 -> 13.9
`gcc` v9.3.0 : neutral
`gcc` v10.0.0 : neutral

xxh3.h
```diff
@@ -1509,18 +1512,101 @@ XXH_FORCE_INLINE void XXH3_initCustomSecret(xxh_u8* XXH_RESTRICT customSecret, x
...
+        }
+    }
+#elif (XXH_VECTOR == XXH_AVX2)
+    XXH_STATIC_ASSERT((XXH_SECRET_DEFAULT_SIZE & 31) == 0);
+    XXH_STATIC_ASSERT((XXH_SECRET_DEFAULT_SIZE / sizeof(__m256i)) == 6);
+    XXH_STATIC_ASSERT(XXH_SEC_ALIGN <= 64);
+    (void)kSecretPtr;
+    (void)(&XXH_writeLE64);
+    (void)i;
+    XXH_PREFETCH(customSecret);
```
Cyan4973, line 1542: minor : what's the benefit of this prefetch ?
`customSecret` is merely written into, and not read from.

xxhash.h
gzm55: On macos, clang declare `_mm512_set1_epi64` and `_mm256_set_epi64x` with type `long long`, and does not define a type of `__int64`.

xxh3.h
gzm55: `customSecret` will be read later when perform the hashing, so when writing the secret, we should not use non-temporal instructions, then the `customSecret` will always be mapped into cache even for write-only access. The prefetch here makes the some of the following assignment a little faster.

gzm55:
Yes, this patch benefits more for shorter inputs.

Gcc seems to generate `SSE2` instructions for the original scalar branch, but does not use a better method, while `clang` can only auto vectorize instructions for `AVX2`. This should explain this phenomenon.

> Interestingly, the impact of `SSE2` / `AVX2` seem reversed by compiler :
> `clang` does not benefit much on `AVX2`, but does get substantial gains with `SSE2`.
> Conversely, `gcc` sometimes see no gains with `SSE2`, but see great gains with `AVX2`.
> 
> Anyway, I'm inclined to believe that this patch does the job, as advertised.

xxh3.h
Cyan4973: I note that the `prefetch` instruction is used on the `AVX2` path, but not the other ones (neither `SS2`, `AVX512` nor `scalar`). Is there a reason for that difference ?

gzm55: I have tried to enable prefetching with different lengths for `SSE2` and `AVX512`, and do not get reproducible improvements, while `AVX2` did.

For `scalar` branch, it produces exactly the same code after this patch as before, so do not enable prefetch.

baryluk: BTW. For AVX512 if you use non-temporal loads (NTA, aka stream loads), it is probably also useful to use non-temporal prefetch instruction. This way you actually protect the cache against evictions, otherwise you will nullify benefits of non-temporarl loads by using wrong prefetch. Just my thoughts. Performance advantages of nta prefetch and loads are very hard to measure, because it impacts both the code doing loads (it could be a bit faster, or maybe use less power), but even if it has the same speed, it might impact positively other code, processes or kernel (by not trashing L1 / L2 caches), which is harder to measure.

Cyan4973:
The PR looks good to me.
There are a number of conflicts though, due to changes introduced during the PR.

Cyan4973:
PR transferred into #377

PR #360, opened 2020-05-09
Contribute a Cryptol specification of XXH32 and XXH64 to the documentation

weaversa:
I created a specification for xxHash in [Cryptol](https://github.com/GaloisInc/cryptol). I thought it might be nice to include it in the xxHash documentation so developers could test against it. 

Cryptol, if you're not familiar, is an interpreted domain specific language for cryptography. The interpreter allows one to quickly generate test vectors at any part of the algorithm. Cryptol also has a tight binding to SMT solvers, useful for proving properties about such specifications.

This also provides the groundwork for potential future work --- on the off chance someone ever wanted to _prove_ that the optimized xxhash implementation here is correct with regard to a formal specification (à la Amazon's [s2n](http://www0.cs.ucl.ac.uk/staff/b.cook/CAV18_s2n.pdf)), they'd need some formal specification to start with.

### DIFF
--- a/doc/xxhash.cry
+++ b/doc/xxhash.cry
@@ -1,0 +1,206 @@
+module xxhash where
+
+/**
+ * The 32-bit variant of xxHash. The first argument is the sequence
+ * of L bytes to hash. The second argument is a seed value.
+ */
+XXH32 : {L} (fin L) => [L][8] -> [32] -> [32]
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks @weaversa !

weaversa:
You're welcome @Cyan4973. I've used xxHash for a few projects and wanted to contribute something back.

PR #367, opened 2020-05-20
Clean up namespace

easyaspi314:
- Every symbol, typedef, and macro is prefixed with `XXH` or `xxh`.
 - For compatibility with legacy code that messes with the internals, `XXH_OLD_NAMES` can be defined to wrap old names in macros. This also brings back `U32` and friends.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -290,7 +290,7 @@ preview-man: man
 	man ./xxhsum.1
 
 .PHONY: test
-test: DEBUGFLAGS += -DDEBUGLEVEL=1
+test: DEBUGFLAGS += -DXXH_DEBUGLEVEL=1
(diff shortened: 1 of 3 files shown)


easyaspi314:
```c
$ clang -DXXH_NO_INLINE_HINTS -c -Dstatic= -fno-inline xxhash.c
$ nm xxhash.o | grep -vi xxh
                 U free
                 U malloc
                 U memcmp
                 U memcpy
                 U memset
```

Cyan4973:
Thanks @easyaspi314, 
indeed, it's a good idea. 
With `xxhash` more likely to be inlined now, it's important to ensure proper namespace isolation, so that user programs are less likely to trigger a name collision.

PR #373, opened 2020-05-27
Remove "Loading..." from source

baryluk:
This makes it way easier to use xxhsum in scripts as a drop-in replacement for sha256sum, md5sum, etc, and stops polluting log files after such replacement. The Loading message also serves no real and demonstrably useful purpose, and messes up with console sometimes, or significantly slows down processing when processing big number of files (10k+).

I am aware of `--quiet` option, but sometimes some tools expect single binary name, not binary with flags, (i.e. `find -exec "xxh64sum --quiet"` will not work). And again, it should be default. All existing popular hash functions don't print such unneeded diagnostics.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -199,7 +199,7 @@ test-xxhsum-c: xxhsum
 	# xxhsum does not display "Loading" message into stderr either
 	! ./xxhsum xxh* 2>&1 | grep Loading
 	# Check that xxhsum do display filename that it failed to open.
-	LC_ALL=C xxhsum noneexistent 2>&1 | grep "Error: Could not open 'nonexistent'"
+	LC_ALL=C ./xxhsum noneexistent 2>&1 | grep "Error: Could not open 'nonexistent'"
 	# xxhsum to/from file, shell redirection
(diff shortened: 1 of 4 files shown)


Cyan4973:
In its early days, `xxhsum` was used to checksum large files, on actual rotating disks, so the loading time wasn't necessarily trivial, and the message served as a feedback signal to the user, telling that the program is not just stuck idle, but is actually processing requested data.

Now, this approach is useless or detrimental when loading a lot of small files, especially on very fast medium like nowadays' nvme. Besides, making `xxhsum` interchangeable with similar utilities such as `md5sum` is a valuable goal in itself.

So I'm fine with this patch once it passes all tests.

baryluk:
> So I'm fine with this patch once it passes all tests.

Done. Silly mistakes.

Cyan4973:
minor detail, indirectly related : I'm surprised that travisCI tests did not fire up to validate this PR.
They should have. We only have appveyor tests right now.

Hopefully, no new issue will appear after the merge (should be fine, this patch is well contained).

PR #377, opened 2020-05-28
Vectorized initCustomSecret()

Cyan4973:
initial PR from @gzm55,
fix some conflicts 
added minor style changes and tests

supersedes #351

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -206,6 +206,12 @@
 #  endif
 #endif
 
+#if XXH_VECTOR == XXH_SSE2 || XXH_VECTOR == XXH_AVX2 || XXH_VECTOR == XXH_AVX512
+#  define XXH_SEC_ALIGN XXH_ACC_ALIGN
+#else
+#  define XXH_SEC_ALIGN 8
+#endif
+
 /*
  * UGLY HACK:
  * GCC usually generates the best code with -O3 for xxHash.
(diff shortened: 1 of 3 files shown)


xxh3.h
```diff
@@ -1464,19 +1470,117 @@ XXH_FORCE_INLINE void XXH_writeLE64(void* dst, xxh_u64 v64)
...
+
+#else  /* scalar code path */
+
     /*
      * We need a separate pointer for the hack below.
      * Any decent compiler will optimize this out otherwise.
      */
-    const xxh_u8 *kSecretPtr = XXH3_kSecret;
-
+    const xxh_u8* const kSecretPtr = XXH3_kSecret;
```
easyaspi314, line 1583: This not being const was intentional for the inline asm hack.

easyaspi314: Here's a diagram of what I am doing.
asm hack kSecretPtr

Cyan4973: I'm surprised it doesn't just fail at compilation time.

I understand that this hack is only valid for `aarch64`, but it should have impact [this test](https://travis-ci.org/github/Cyan4973/xxHash/jobs/692011930) for example.

Cyan4973: So the reason it was not failing was that the `aarch64` test was done with `gcc` rather than `clang`, but the hack is only active for `clang`.

Just added `clang` tests, and now the issue is properly detected.

PR #381, opened 2020-05-29
Make xxhsum check if fread() failed

WayneD:
If fread() fails reading a file, output an error and exit.  The new code
also avoids an update call on a 0-length buffer at end of file.


Cyan4973:
Thanks @WayneD !

baryluk:
Thanks. We probably also should check `fclose`, and emit a warning (and continue with other files). Technically it should never fail if the file opened for read, but it is still a good practice to check it.

Issue #361, opened 2020-05-14
XXH3 Keyed Variant long secrets


koraa:
Hi,

thanks again for creating XXH3! Very impressive!

I just finally realized that the keyed variant of XXH3 requires a 192 byte secret (I always assumed it would be 192 bits). Should we make this point extra clear in the doc since 192 bit is a common key size? I'll be happy to open a PR…

192 byte also seems a bit excessive for many use cases; in terms of entropy a much shorter key length should suffice for a non cryptographic hash function…For my use case I am probably going to use chacha20 to expand a 32 byte key, or maybe SHAKE to support other key lengths, but maybe you have a better idea?

Could a specialized KDF be added to xxhash to support shorter key lengths natively?

easyaspi314:
Maybe we should make `XXH3_initCustomSecret` public.

Cyan4973:
The minimum size for a secret is [`136` bytes](https://github.com/Cyan4973/xxHash/blob/dev/xxhash.h#L518), 
but indeed, that's still a too large a length for the described use case.

At the other extreme, it's possible to initialize using a `seed`, which must be exactly 8-bytes long.

Therefore, there is a gap between those two variants, which do not allow initialization with a 32-byte key. This is a capability `XXH3` doesn't have yet, but indeed, it could be added.

When it comes to very large inputs, there's no problem : it's certainly possible to generate a "large enough" secret from a too small one, and could be added easily.

But for API completeness, it doesn't seem right to offer this option only for "long" inputs.
One also has to consider "small inputs" (<= 240 bytes). And now it becomes a more complex story case.

A simple generic solution would always generate the full internal secret, and work from there. This would work fine, but there is a "startup cost" for each hash, when generating the internal secret, and this fixed cost becomes non trivial as input length becomes small.

For example, presuming the input is still 5 bytes long, and the custom secret is 32-bytes long, the naive initialization method would still generate a full internal secret (likely 192-bytes long) before starting calculation of the hash on the 5-bytes input, adding some non-negligible latency.

Since only a portion of the secret will be used, an alternative could be to generate only the portion of the secret that will be actually used for hash calculation. For example, for a 5-byte input, it is only 16-bytes long. However, this would add branches at the beginning, potentially harming performance more than a straightforward initialization operating on a stable length.

[...]

PR #386, opened 2020-06-08
Key Derivation function

Cyan4973:
created `XXH3_generateSecret()`,
answering #361 .

updated documentation and added corresponding test.

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -2036,14 +2036,31 @@ XXH3_generateSecret(void* secretBuffer, const void* customSeed, size_t customSee
 
     {   size_t const segmentSize = sizeof(XXH128_hash_t);
         size_t const nbSegments = XXH_SECRET_DEFAULT_SIZE / segmentSize;
-        XXH128_hash_t scrambler = XXH128(customSeed, customSeedSize, 0);
+        XXH128_canonical_t scrambler;
+        XXH64_hash_t seeds[12];
         size_t segnb;
+        XXH_ASSERT(nbSegments == 12);
         XXH_ASSERT(segmentSize * nbSegments == XXH_SECRET_DEFAULT_SIZE); /* exact multiple */
+        XXH128_canonicalFromHash(&scrambler, XXH128(customSeed, customSeedSize, 0));
+
+        /* prepare seeds */
+        {   size_t toFill = (customSeedSize > sizeof(seeds)) ? sizeof(seeds) : customSeedSize;
+            size_t filled = toFill;
+            memcpy(secretBuffer, customSeed, toFill);
+            while (filled < sizeof(seeds)) {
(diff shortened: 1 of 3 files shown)


koraa:
Couple of comments regarding the KDF:

First, I still believe using LE instead of BE/canonical representation would be preferable; this is an internal functions and pretty much all architectures that matter are using LE, so they all need to reverse the bytes at the moment, worsening performance.

Then I would like to call into question that the best API for this is exposing XXH3_generateSecret; it seems to me that this basically ended up being a variant of XXH3 with a 128 bit seed and a more secure key derivation function; I propose that the API exposes the function like that `withSeed128`. Users can invoke XXH3 themselves if they need to compress a longer secret.

Given the current definition of the KDF I believe it would even be possible to define a short-cut variant for short strings using a 128bit seed similar to the 64bit seed variant…

Finally I believe it is worth considering, whether the withSecret variant should not simply be based on `withSeed128`. Given that there would be a relatively secure/long seed variant of XXH3, the only real use case for `withSecret` is performance optimization, which could still be achieved with the following pattern:

```C
// Precompute secret
XXH3_state_t state;
XXH3_128bits_reset_withSeed128(&state, mySeed);

// Use it
XXH3_state_t myHasher = state;
XXH3_128bits_update(&myHasher, ...);
```

Defining `withSecret` in terms of `withSeed128` could have other advantages: (1) improved robustness against low quality secrets (2) reduces the memory management headache, since the entropy pool would never point at a user defined buffer.

This change would not impact API compatibility, but it may introduce a performance regression in some setups; given that xxh3 is currently still in beta I believe that this is acceptable.

Cyan4973:
> it seems to me that this basically ended up being a variant of XXH3 with a 128 bit seed and a more secure key derivation function

This might have been lost in the noise, but this version of `generateSecret()` offers way more than 128-bit of variance, due to employing portions of `customSeed` as additional `seed` when generating segments of the `secret`.

Anyway, in order to preserve room for future changes, I'm fine with moving that function in the experimental section of the API, so that its outcome becomes non-binding on stabilizing the main algorithm.

> I propose that the API exposes the function like that withSeed128

The existence of such a variant was publicly asked as an opened question last year. It's unfortunate there was no identified usage back then.

Introducing it now, while possible, has the potential to slow down the stabilization stage. The problem is mostly concentrated on the small inputs, where it will be necessary to introduce the new variant in a way which remains compatible with existing `_withSeed()`.

Also, the `secret` initialization planed for `_withSeed128()` is basically a variant of `_withSeed()`, hence nowhere near as complex as `generateSecret()`.

> I believe it is worth considering, whether the `withSecret` variant should not simply be based on `withSeed128`

The level of freedom offered by `_withSecret()` cannot be matched by `_withSeed128()`. Consequently, the amount of unknown introduced by `_withSecret()` is much larger, making it way more difficult for an external actor to predictably tamper with internal accumulators.

PR #387, opened 2020-06-10
Intel vector instruction dispatcher

Cyan4973:
Started from a baseline developed by @easyaspi314 (#358),
`xxh_x86dispatch.c` makes it possible to dynamically detect
which vector instruction set is present on the active system
and use the (presumable) best available one,
using this priority order : `avx512` > `avx2` > `sse2` > `scalar`.

3rd party applications willing to employ this ability
just have to `#include "xxh_x86dispatch.h"`.
This header automatically includes `xxhash.h`,
and swap relevant `XXH3`/`XXH128` symbols
with their vector dispatch counterpart.

Only runs on `x86`/`x64`.
Only compiles with `gcc` and `clang`.

added : `make dispatch` target, which compiles `xxhsum` with this new ability.
added : `make xxhsum DISPATCH=1`, which is equivalent to `make dispatch`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ xxhsum
 xxhsum32
 xxhsum_privateXXH
 xxhsum_inlinedXXH
+dispatch
 tests/generate_unicode_test
 
 # compilation chain
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,8 @@ matrix:
             - cppcheck
       script:
         - make -B test-all
+        - make clean
+        - make dispatch
 
     - name: Check results consistency on x64
       arch: amd64
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ matrix:
         - make clean
         - CPPFLAGS="-mavx2 -DXXH_VECTOR=2" make check   # AVX2 code path
         - make clean
-        - CPPFLAGS="-mavx512f -DXXH_VECTOR=5" make check   # AVX512 code path
+        - CPPFLAGS="-mavx512f -DXXH_VECTOR=3" make check   # AVX512 code path
         - make clean
         - CPPFLAGS=-DXXH_REROLL=1 make check   # reroll code path (#240)
         - make -C tests/bench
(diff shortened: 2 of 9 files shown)


Cyan4973:
It happens that my wife has one those almost rare icelake laptop, which supports `avx512f`.
So it's a nice playground to test this code path.

Good news : `xxhsum` compiles and runs fine with `-mavx512f`.
Bad news : `dispatch` compiles, with both `gcc` and `clang`, but then fails at runtime.

Exact reason is unclear yet. 
`gcc` just immediately `segfault`. 
`clang` fails during the self-test (64-bit hash test 100, likely related to `xxh3`).

This feels like some out-of-bound access.

No problem with a standard `avx2` laptop or desktop, tested with `macosx`, `linux` and Windows. So, one needs the specific hardware to observe ...

easyaspi314:
Weird...Does my version also segfault?

(On my branch you can just run `make DISPATCH=1`).

Also, I believe Intel's SDE might be useful for testing AVX512.

Cyan4973:
OK, this is fixed. This was an index issue, now checked with a `static_assert()`.
Dispatcher works fine on `avx512` too now.

easyaspi314:
Shouldn't we use the official names for `-DXXH_VECTOR`? e.g. `XXH_SSE2`, `XXH_AVX2`, etc?

Cyan4973:
Are you referring to the `README.md` documentation on `XXH_VECTOR` build macro ?

easyaspi314:
I mean in places like `travis.yml` where the bug was caused by mixing up the magic ID numbers for `XXH_VECTOR`.
If we just used the textual names, we wouldn't have had the issue you just had.

Cyan4973:
In reality, this wasn't the source of the problem. The change in vector ID numbers is merely a consequence of the selected fix.

But I do nonetheless agree that expressing vector instruction set in name rather than ID would be beneficial for readability.

Thing is, I don't know how to do that without swamping the `Makefile` with a ton a lines of code just for this (barely used and mostly debug) feature.

Anyway, if you believe you have a good solution, you're welcomed to give it a try.
It could be part of a separate PR, as I guess it doesn't impact the current one.

easyaspi314:
> Thing is, I don't know how to do that without swamping the Makefile with a ton of lines of code just for this (barely used and mostly debug) feature.

Huh? I am just talking about using `-DXXH_VECTOR=XXH_AVX2` instead of `-DXXH_VECTOR=2`.

Cyan4973:
My bad, misunderstood,
yes you are right, this is cleaner

PR #394, opened 2020-06-17
Some Makefile improvements

WayneD:
- Fix identification of Cygwin/MSYS/MINGW "OS" that have $(OS) of
  Windows% but shouldn't obey the Windows rules.
- Add Cygwin to the list of validated installs. I ran install as:
  `PREFIX=/usr make install`
- The uninstall target now removes the xxh3.h file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,7 +117,7 @@ libxxhash.a: xxhash.o
 	$(AR) $(ARFLAGS) $@ $^
 
 $(LIBXXH): LDFLAGS += -shared
-ifneq ($(UNAME), Windows)
+ifeq (,$(filter Windows%,$(OS)))
 $(LIBXXH): CFLAGS += -fPIC
 endif
 $(LIBXXH): xxhash.c
--- a/Makefile
+++ b/Makefile
@@ -339,7 +339,7 @@ trailingWhitespace:
 # =========================================================
 # make install is validated only for the following targets
 # =========================================================
-ifneq (,$(filter Linux Darwin GNU/kFreeBSD GNU OpenBSD FreeBSD NetBSD DragonFly SunOS CYGWIN% , $(shell uname)))
+ifneq (,$(filter Linux Darwin GNU/kFreeBSD GNU OpenBSD FreeBSD NetBSD DragonFly SunOS CYGWIN% , $(UNAME)))
 
 DESTDIR     ?=
 # directory variables: GNU conventions prefer lowercase
--- a/Makefile
+++ b/Makefile
@@ -44,8 +44,17 @@ CFLAGS += $(DEBUGFLAGS) $(MOREFLAGS)
 FLAGS   = $(CFLAGS) $(CPPFLAGS)
 XXHSUM_VERSION = $(LIBVER)
 
+ifeq '$(findstring ;,$(PATH))' ';'
+    UNAME := Windows
+else
+    UNAME := $(shell uname 2>/dev/null || echo Unknown)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -44,16 +44,25 @@ CFLAGS += $(DEBUGFLAGS) $(MOREFLAGS)
 FLAGS   = $(CFLAGS) $(CPPFLAGS)
 XXHSUM_VERSION = $(LIBVER)
 
+ifeq '$(findstring ;,$(PATH))' ';'
```
Cyan4973, line 47: What's the logic used here ?
If `PATH` is not defined then it's necessarily Windows ?

Makefile
```diff
@@ -44,16 +44,25 @@ CFLAGS += $(DEBUGFLAGS) $(MOREFLAGS)
...
+    UNAME := Windows
+else
+    UNAME := $(shell uname 2>/dev/null || echo Unknown)
+    UNAME := $(patsubst CYGWIN%,Cygwin,$(UNAME))
```
Cyan4973, line 51: In a recent patch, I modified the OS detection below so that variable strings such as `CYGWIN%` can be used directly within the `filter()`

Makefile
```diff
@@ -410,6 +419,7 @@ uninstall:  ## uninstall libraries, CLI, links and man page
 	@$(RM) $(DESTDIR)$(LIBDIR)/libxxhash.$(SHARED_EXT_MAJOR)
 	@$(RM) $(DESTDIR)$(LIBDIR)/$(LIBXXH)
 	@$(RM) $(DESTDIR)$(INCLUDEDIR)/xxhash.h
+	@$(RM) $(DESTDIR)$(INCLUDEDIR)/xxh3.h
```
Cyan4973, line 422: :+1:

Makefile
easyaspi314: An ugly hack that takes advantage of `%PATH%` using semicolons as a delimiter instead of colons. 

Honestly I think there are better options though.

WayneD: I think it's a pretty good differentiation since the $(OS) value for a Cygwin build is just Windows_NT and only Windows uses a semicolon as a PATH separator. However, if you can think of a better way to figure out that a Posix-compatible build on Windows is in progress, feel free to improve that.

Makefile
WayneD: Yup, but that puts all the match complexity down into every ifeq. I think it's nicer to simplify the values and thus make the follow-up conditionals easier. However, either way works fine.

Cyan4973: I agree for things which are used multiple times.
For example, defining `UNAME :=` once and reusing it everywhere is a neat idea.

But for the larger pattern test enabling the `make install` section, it's only invoked once.
So is it really better to define it elsewhere, rather than at the one place where it's used ?

Makefile
Cyan4973: What about the existing test : `ifneq (,$(filter Windows%,$(OS)))` ? (eventually modified to make it a variable)

WayneD: My tweak of the .exe suffix was unneeded, since even cygwin likes .exe suffixes. The lack of the -fPIC option also seems to work fine, so your change for make install seems good.

Makefile
WayneD: My latest merge just defines UNAME and fixes up the uninstall of xxh3.h.

Cyan4973:
Thanks @WayneD !

PR #395, opened 2020-06-17
fixed demerphq avalanche tests for len [0-3]

### LINKED ISSUE #344: XXH3 Fail Avalanche Tests from demerphq/smhasher
# How Reproduce
```bash
git clone https://github.com/demerphq/smhasher.git demerphq-smhasher
git clone https://github.com/Cyan4973/xxHash.git
git clone https://github.com/demerphq/BeagleHash.git # optional
cp xxHash/{xxhash.*,xxh3.h} demerphq-smhasher/
cd demerphq-smhasher

# check source version
git log -n 1 | grep 0762887
sha1sum xxh3.h | grep e41cf39a164c491ae86deb85c10051c5826e8f44
sha1sum xxhash.c | grep f1a286236f359bd1f42942025ea9a9067d42ebd7 
sha1sum xxhash.h | grep 14b6b68b53dd6bda4d12db6d304862b04a9447b4

# edit Hash.h main.cpp
~/demerphq-smhasher$ git diff Hashes.h main.cpp
diff --git a/Hashes.h b/Hashes.h
index 0484116..25ac5fa 100644
--- a/Hashes.h
+++ b/Hashes.h
@@ -10,6 +10,7 @@
 #include "MurmurHash3.h"

 #if defined(__x86_64__)
+#define XXH_INLINE_ALL
 #include "xxhash.h"
 #include "metrohash.h"
 #include "cmetrohash.h"
@@ -166,6 +167,10 @@ inline void xxHash64_with_state_test( const void * key, int len, const void *sta
   *(uint64_t*)out = (uint64_t) XXH64(key, (size_t) len, *((unsigned long long *)state));
 }

+inline void xxh3_with_state_test( const void * key, int len, const void *state, void * out ) {
+  *(uint64_t*)out = (uint64_t) XXH3_64bits_withSeed(key, (size_t) len, *((uint64_t *) state));
+}
+
 inline void metrohash64_1_with_state_test ( const void * key, int len, const void *state, void * out ) {
   metrohash64_1((const uint8_t *)key,(uint64_t)len,*((uint64_t*)state),(uint8_t *)out);
 }
diff --git a/main.cpp b/main.cpp
index 7d887a1..9670198 100644
--- a/main.cpp
+++ b/main.cpp
@@ -367,6 +367,9 @@ HashInfo g_hashes[] =
   { "xxHash64", "xxHash, 64-bit",
     64, 64, 64, 0x22B287C9,
     NULL, xxHash64_with_state_test },
+  { "xxh3", "XXH3, 64-bit",
+    64, 64, 64, 0x8710F469,
+    NULL, xxh3_with_state_test },
 #endif
 #if defined(__x86_64__)
   { "jodyhash32", "JodyHash for 32 bit (v5)",


./make_smhasher || : # generate VERSION.h
mkdir -p build
pushd build
cmake ..
make -j4
cp SMHasher ..
popd

# Avalanche Test
./SMHasher --test=Avalanche xxh3
./SMHasher --test=Avalanche xxHash64
```

# Expected Result
All tests OK, xxHash64 passed.

# Actual Result
xxh3 fails for len in [0-3] bytes.

Cyan4973:
fix #344

This modification fixes `demerphq`'s avalanche tests for len [0-3],
(_outdated_) by switching to stronger `rrmxmx` avalanche, already in use for len [4-8].

(_outdated_ ) This is not without consequences : the new variant is ~10% slower in latency, and ~20% slower in bandwidth. That being said, the impact is only felt in the len [0-3] range, so it's pretty limited. Also, the previous function was focused on providing good mixing for (up to) 24-bit of input, while the new one also mixes well the 64-bit `seed`, which is kind of an extended scope.

__Update__ : previous paragraph is no longer correct, new version produces identical perf, by switching to `XXH64` avalanche instead of `rrmxmx`.

The real issue is that this change makes the result of `XXH3()` no longer comparable to previous version. I know that we are still in development stage, and that technically it's still possible to make such changes, but that's still not great news now that we are entering "release candidate" stage.

Well, anyway, in order to pass the more stringent `demerphq` avalanche test, there is no magic, any solution will have to change the result of the function.

Benchmark on a core i7-9700k, using `gcc v9.3.0` : 
```
Len,     1 ,     2 ,     3 ,     4
Throughput small inputs of fixed size :
PR , 403.5M, 401.0M, 403.6M, 441.0M
dev, 401.5M, 402.8M, 402.5M, 437.9M
Latency for small inputs of fixed size :
PR , 194.2M, 193.3M, 193.9M, 176.3M
dev, 193.8M, 191.4M, 193.9M, 177.8M
```

basically, identical performance

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -651,9 +651,8 @@ XXH_FORCE_INLINE xxh_u64 XXH_xorshift64(xxh_u64 v64, int shift)
 }
 
 /*
- * We don't need to (or want to) mix as much as XXH64.
- *
- * Short hashes are more evenly distributed, so it isn't necessary.
+ * This is a fast avalanche stage,
+ * suitable when input bits are already partially mixed
  */
 static XXH64_hash_t XXH3_avalanche(xxh_u64 h64)
 {
@@ -661,6 +660,21 @@ static XXH64_hash_t XXH3_avalanche(xxh_u64 h64)
     h64 *= 0x165667919E3779F9ULL;
     h64 = XXH_xorshift64(h64, 32);
     return h64;
+}
+
+/*
+ * This is a stronger avalanche,
+ * inspired by Pelle Evensen's rrmxmx
+ * preferable when input has not been previously mixed
+ */
+static XXH64_hash_t XXH3_rrmxmx(xxh_u64 h64, xxh_u64 len)
+{
+    /* this mix is inspired by Pelle Evensen's rrmxmx */
+    h64 ^= XXH_rotl64(h64, 49) ^ XXH_rotl64(h64, 24);
+    h64 *= 0x9FB21C651E98DF25ULL;
+    h64 ^= (h64 >> 35) + len ;
+    h64 *= 0x9FB21C651E98DF25ULL;
+    return XXH_xorshift64(h64, 28);
 }
 
 
@@ -708,15 +722,14 @@ XXH3_len_1to3_64b(const xxh_u8* input, size_t len, const xxh_u8* secret, XXH64_h
      * len = 2: combined = { input[1], 0x02, input[0], input[1] }
      * len = 3: combined = { input[2], 0x03, input[0], input[1] }
      */
-    {   xxh_u8 const c1 = input[0];
-        xxh_u8 const c2 = input[len >> 1];
-        xxh_u8 const c3 = input[len - 1];
+    {   xxh_u8  const c1 = input[0];
+        xxh_u8  const c2 = input[len >> 1];
+        xxh_u8  const c3 = input[len - 1];
         xxh_u32 const combined = ((xxh_u32)c1 << 16) | ((xxh_u32)c2  << 24)
                                | ((xxh_u32)c3 <<  0) | ((xxh_u32)len << 8);
         xxh_u64 const bitflip = (XXH_readLE32(secret) ^ XXH_readLE32(secret+4)) + seed;
         xxh_u64 const keyed = (xxh_u64)combined ^ bitflip;
-        xxh_u64 const mixed = keyed * XXH_PRIME64_1;
-        return XXH3_avalanche(mixed);
+        return XXH3_rrmxmx(keyed, 0);
     }
 }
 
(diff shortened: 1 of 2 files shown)


tommyettinger:
I'm on the fence about this one. For most users it will have minimal performance penalty, but changes the results for hashes of small inputs, which could break compatibility where those hashes mattered. But, we don't know where the avalanche might matter for some users who rely on it with similar seeds... This is both a safer approach because it improves quality in a corner case, and a less-stable approach because it changes some results. I guess this needs asking the question, could another set of result-changes (say, for XXH4) happen at some point, and how would they be handled? Or alternately, should "randomizing the seed" just not be considered a priority for XXH3 when the inputs are so small?

easyaspi314:
My hot take on this:

On the plus side:
 - We pass an extra test
 - We provide a little extra protection for seeds
On the minus side:
 - We change the algorithm *again*
 - If I am reading this correctly, the test is failing because of changes in the *seed* — 99% of the time, the seed will be effectively constant, so these changes would not happen.

Cyan4973:
I agree. This change, alone, seems to bring too little improvements for the cost of breaking compatibility with `v0.7.3`.
So it can stay here, un-merged, and be re-considered if there are other more significant breaking changes required for next release.

Cyan4973:
Note, for the record :
it seems that using `XXH64_avalanche()` instead of `XXH3_rrmxmx()` is good enough to pass the `demerphq` avalanche test, but doesn't suffer from any noticeable speed drop.
Of course, it doesn't change the main problem that results are no longer comparable to `v0.7.3`, 
but if it comes to it, such a variant would be better from a performance perspective.

easyaspi314:
> Also, switched all len [0-3] avalanche to `XXH64_avalanche()`, since it's faster, while avoiding the performance cost of `rrmxmx`.

Have you benchmarked this?

Cyan4973:
> Have you benchmarked this?

Yes. As mentioned, the performance measured with `benchHash` is very close to current `dev` branch.

Issue #382, opened 2020-06-01
bsd output style


grylem:
Hello
please add options for bsd otput style for calc and checking like
XXHASH32 (filename) sum
XXHASH64 (filename) sum
XXHASH128 (filename) sum
Thanks.

easyaspi314:
This isn't a bad idea.
The only potential issue is that *older* versions of xxhsum wouldn't be able to read it.
For the reference, it would go like this
```
XXHASH128 (checksum.xxh128) = 9eb7a54d24dbf6a2eb9f7ce7a1853cd0
```

Note that coreutils provides the `--tag` option which switches to the BSD format.

Cyan4973:
Preserving consistency with `coreutils` is important.
Adding support for `--tag` format is a good idea.

grylem:
sorry i made a mistake in the title should be "bsd output style".

Cyan4973:
Adding a new output mode will be easy.

What will be more difficult is to add a matching consumption mode.
`xxhsum -c` must read files produced by `xxhsum`, to control listed checksums.
For consistency, it will also have to read files produced by `xxhsum --tag`.

This will cost code, complexity and time.

easyaspi314:
Well if we are going to match coreutils, we are also missing the backslash modifier.

If the first non-space character is a backslash, it will escape `\n` and `\\`.

So
```shell
$ export FILENAME=$'foo\n\\bar.txt'
$ echo "$FILENAME" > "$FILENAME"
$ md5sum "$FILENAME"
\dc7fd66c4fc94c46b7c91d95c2c30914  foo\n\\bar.txt
$ md5sum --tag "$FILENAME"
\MD5 (foo\n\\bar.txt) = dc7fd66c4fc94c46b7c91d95c2c30914
$ xxhsum "$FILENAME"
cbf4f722b723ac65  foo
\bar.txt
```

If we really want to do that, it will be a pain unless we just copy from coreutils.

Cyan4973:
That's probably a different topic, unrelated to the `--tag` option.

grylem:
[RHash](https://github.com/rhash/RHash) can work with both types of output format - BSD and md5sum. For calculating and checking.

baryluk:
[...]

PR #396, opened 2020-06-18
introduced --tag for BSD-style output

Cyan4973:
Follows #382
`--tag` is an hidden option for the time being (undocumented)
as it's only an output format,
there is no ability to `-c` check it yet.

The effort was also useful in refactoring part of `xxhsum` code base.

One advantage of the `--tag` BSD format is
that it's a lot safer to introduce additional formats.
For example, little and big-endian variants can co-exist safely,
since endianess is clearly expressed as part of the algorithm name.
It would also make it possible to introduce XXH3 (later),
even though it's 64-bit, which would be confusing with XXH64.

Also :
added ability to select algorithm by specifying bit width.
Hence :
H0 == H32 ==> XXH32
H1 == H64 ==> XXH64
H2 == H128==> XXH128

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -227,6 +227,20 @@ test-xxhsum-c: xxhsum
 	./xxhsum -c < .test.xxh32
 	# check variant with '*' marker as second separator
 	$(SED) 's/  / \*/' .test.xxh32 | ./xxhsum -c
+	# check bsd-style output
+	./xxhsum --tag xxhsum* | $(GREP) XXH64
+	./xxhsum --tag -H0 xxhsum* | $(GREP) XXH32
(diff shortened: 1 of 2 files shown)


easyaspi314:
Why did you bother changing `argv` to `const char **`? It just complicates things due to the nested `const` rule. 

`main` is either `int main(void)` or `int main(int argc, char **argv)`, not `int main(int argc, const char **argv)`.

Cyan4973:
I like `const` that much  🤷 

Besides, the impact was not limited to `main()` : all downstream functions were in turn forced to take a mutable `char**` parameter due to the impossibility to cast `char**` into `const char**` on Windows. This is precisely what I want to avoid.

Issue #398, opened 2020-06-19
Clang 9.0.0 x86_64 performance regression in non-dispatched code


easyaspi314:
It seems that there has been a large performance regression on Clang for x86_64 on the non-dispatched path.
This does not affect GCC, only Clang apparently.

MacBookPro8,2 (15-inch, Early 2011)
2.0GHz Intel Core i7-2635QM (Sandy Bridge)

Clang 9.0.0 (macOS uses `-msse4.1` by default)

`dev` (dispatch disabled)
```      
 5#XXH3_64b                      :     102400 ->    87853 it/s ( 8579.4 MB/s)   
 6#XXH3_64b unaligned            :     102400 ->    82803 it/s ( 8086.2 MB/s)   
 7#XXH3_64b w/seed               :     102400 ->   134703 it/s (13154.6 MB/s)   
 8#XXH3_64b w/seed unaligned     :     102400 ->   126697 it/s (12372.8 MB/s)   
 9#XXH3_64b w/secret             :     102400 ->   123520 it/s (12062.5 MB/s)   
10#XXH3_64b w/secret unaligned   :     102400 ->   114818 it/s (11212.7 MB/s)   
11#XXH128                        :     102400 ->    77291 it/s ( 7548.0 MB/s)   
12#XXH128 unaligned              :     102400 ->    70927 it/s ( 6926.5 MB/s)   
13#XXH128 w/seed                 :     102400 ->   125048 it/s (12211.7 MB/s)   
14#XXH128 w/seed unaligned       :     102400 ->   118428 it/s (11565.2 MB/s)   
15#XXH128 w/secret               :     102400 ->    84696 it/s ( 8271.1 MB/s)   
16#XXH128 w/secret unaligned     :     102400 ->    74465 it/s ( 7272.0 MB/s) 
```
1b14f648d63ddbf66a99e043c472789575c3673e
```
XXH3_64b                      :     102400 ->   135562 it/s (13238.5 MB/s)      
XXH3_64b unaligned            :     102400 ->   125060 it/s (12212.9 MB/s)      
XXH3_64b w/seed               :     102400 ->   135817 it/s (13263.4 MB/s)      
XXH3_64b w/seed unaligned     :     102400 ->   128615 it/s (12560.1 MB/s)      
XXH3_64b w/secret             :     102400 ->   124632 it/s (12171.0 MB/s)      
XXH3_64b w/secret unaligned   :     102400 ->   115404 it/s (11269.9 MB/s)      
[...]

PR #399, opened 2020-06-19
fixed performance issue reported in #398

Cyan4973:
note :
I was unable to reproduce the speed issue on my own platforms (laptop, desktop, macos, ubuntu).
Therefore, @easyaspi314, could you test that this patch solves your performance issue ?

The fix feels logical enough, and doesn't negatively impact performance either, so even if I can't observe performance benefits directly, it's good to go.

Perhaps more importantly : 
this patch also fixes `XXH128_withSecret()`,
which was messed up with default `kSecret`.
The error remained unnoticed because there was no self-test for this variant.
Therefore, also added a test in `xxhsum` checking results of `XXH128_withSecret()`

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -1770,6 +1770,21 @@ XXH3_hashLong_64b_withSecret(const xxh_u8* XXH_RESTRICT input, size_t len,
 }
 
 /*
+ * It's important for performance that XXH3_hashLong is not inlined.
+ * Since the function is not inlined, the compiler may not be able to understand that,
+ * in some scenarios, its `secret` argument is actually a compile time constant.
+ * This variant enforces that the compiler can detect that,
+ * and uses this opportunity to streamline the generated code for better performance.
+ */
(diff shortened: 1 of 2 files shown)


easyaspi314:
I will test performance tomorrow.

I did realize that the performance difference happened when adding XXH32 and XXH64's streaming variants to the table...which somehow made a difference to XXH3. 🤔

Cyan4973:
> I did realize that the performance difference happened when adding XXH32 and XXH64's streaming variants to the table...which somehow made a difference to XXH3.

Instructions alignment ?...

easyaspi314:
Possibly, I believe my MacBook might have the loop alignment errata. It is really strange though. 

Either way, nice find on the XXH128 bug…imagine if that got to v0.7.4...

Cyan4973:
I'm still interested to know the impact of this PR on your system.
In the meantime, since I believe there is no downside to this PR, and it also fixes another nasty problem related to `XXH128_withSecret()`, I'll merge it.

PR #405, opened 2020-06-23
Dispatch defines

erijo:
Make it possible to build xxh_x86dispatch.c with `XXH_DISPATCH_AVX` and/or `XXH_DISPATCH_AVX512` undefined.


Cyan4973:
Thanks @erijo !

Issue #348, opened 2020-04-07
custom secrets of XXH3 blocks loop unrolling


gzm55:
In discussion of #347, we found that the runtime size of custom secret seems to be the main issue of benchmark results of `_withSecret()` variants, which blocks the internal loop unrolling. Now there are 2 known workarounds:

### switch table

Use switch table to convert some known length as compile time constants. This works for multiple targets including scalar, sse2, avx2 and avx512, but may bloat the executable size.

```c
XXH_NO_INLINE XXH64_hash_t
XXH3_hashLong_64b_withSecret(const xxh_u8* XXH_RESTRICT input, size_t len,
                             const xxh_u8* XXH_RESTRICT secret, size_t secretSize)
{
    switch (secretSize) {
    case XXH3_SECRET_SIZE_MIN + 0 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN + 0 * 8);
    case XXH3_SECRET_SIZE_MIN + 1 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN + 1 * 8);
    case XXH3_SECRET_SIZE_MIN + 2 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN + 2 * 8);
    case XXH3_SECRET_SIZE_MIN + 3 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN + 3 * 8);
    case XXH3_SECRET_SIZE_MIN + 4 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN + 4 * 8);
    case XXH3_SECRET_SIZE_MIN + 5 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN + 5 * 8);
    case XXH3_SECRET_SIZE_MIN + 6 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN + 6 * 8);
    case XXH3_SECRET_SIZE_MIN + 7 * 8: return XXH3_hashLong_64b_internal(input, len, secret, XXH3_SECRET_SIZE_MIN + 7 * 8);
    ......
    }
    return XXH3_hashLong_64b_internal(input, len, secret, secretSize);
}
```

### manually unrolling

[...]

PR #406, opened 2020-06-23
reverted default secret to Farsh

Cyan4973:
I launched several quality tests overnight, 
testing branch `newOutput074` using the new `Pi` decimals constant, 
and the results are not so good.
To be clear, they are not "horrible" either. They are just a bit less good than previous default `secret` in a few tests.
I quickly attempted another experiment, using `e` the Euler constant, 
and it was a bit better, but still not as good as previous default `kSecret`.

I don't want to spend too much time on the topic, 
and at this stage, the simplest and quickest solution is to just use previous constant which has proven to work best so far.


gzm55:
Since the hash quality is bound to the quality of secrets, and it is not easy to discover a high quality secret, do the developers really need an api accepting a custom secret?

easyaspi314:
> Since the hash quality is bound to the quality of secrets, and it is not easy to discover a high quality secret, do the developers really need an api accepting a custom secret?

What about if you discover a *higher* quality secret?

gzm55:
> > Since the hash quality is bound to the quality of secrets, and it is not easy to discover a high quality secret, do the developers really need an api accepting a custom secret?
> 
> What about if you discover a _higher_ quality secret?

The custom-secret APIs can be development-only (enabled by a special macro) for searching new  _higher_ quality secrets. Once found, he then should use an internal copy of `xxhash.h` and `xxh3.h`, update the default secret to the new set.

Cyan4973:
There are a few points here : 

1) The quality of the hash function depends on the "randomness quality" of the `secret` : yes, this is nothing new. We already know that feeding `secret` with a blank page full of `\0` is pretty bad. Even less obvious examples, such as a repeating sequence, a repeated non-zero character, or even a text file with ascii character set, are also bad. The `secret` must be "random looking", at a binary level, otherwise the quality of the hash function suffers. This is true, this is documented, I can make it even more clear by emphasizing the warning. The point is, this is not new.

2) "it is not easy to discover a high quality secret" : well, it depends. Any "random looking" sequence is expected to be good enough for this purpose. If the point is raised in reaction of sticking to existing Farsh secret, instead of switching to `Pi` decimals, let's be clear that the `Pi` decimals is good enough too. I'm just a bit picky here, thinking that if I have to choose between 2 _default_ `secret`, I'd better pick the one with better metrics attached. But let's be clear, it's not a "game changer", both `Pi` and `e` decimals are also fit for the job.

3) If previous sentence is merely stating that finding a good random-looking source is not that trivial to begin with, well this is indeed correct. Any kind of man-made source tends to feature some kind of "structure", which makes it "not random". In essence, that's the reason why a new function `generateSecret()` is proposed, which should fix the issue of having to "find" a random-looking source by transforming any kind of `seed` content into a `secret` of high entropy quality. I believe it answers this concern completely.

4) "do the developers really need an api accepting a custom secret?"
I believe it's useful. I also believe it's not a "basic" feature : it's a fairly advanced API, targeting expert users.
If the point is that some non-expert users could mis-use it, just because it exists, I would say that yes, that's true, but I don't think it's a good enough reason to completely remove the entry point from the API, since it can also have legitimate usages. A more appropriate mitigation is to better document the feature. It can also be pushed into a kind of "expert only" section, to increase warning signals.

5) "The custom-secret APIs can be development-only (enabled by a special macro) for searching new higher quality secrets. Once found, he then should use an internal copy of xxhash.h and xxh3.h, update the default secret to the new set."
Well, if one uses the `_withSecret()` variant in combination with `XXH_INLINE_ALL`, this essentially achieves the same objective, without the need to create and maintain a separate code, thanks to the power of inlining : the custom secret will be inserted in-place. This is, I believe, way more powerful and secure than creating a separate code.

6) Worst case, if the `_withSecret()` API is deemed too dangerous, or need some refactor or rethinking, I could isolate it into the "experimental" section of the API, alongside `generateSecret()`. In which case, this API will not be considered "stabilized" by `v0.8.0`. It will have to be unlocked by using `XXH_STATIC_LINKING_ONLY` macro. It's also implied that the function return value could still change in a future version, which may be a bit too much.
At this stage though, I have no idea how such a variant could be made any "better", and pushing it into that section will surely hamper its broader adoption. Still, if that's the point, it could be done. I would need a few arguments in this direction to take such a decision.

gzm55:
> I believe it's useful. I also believe it's not a "basic" feature : it's a fairly advanced API, targeting expert users.

Agree, APIs with `secret` should be used with caution, and even be recommended not to use directly in production environments. Then, maybe there is very little benefit to fix #348 .  We should consider a safer and compile-time method for expert to change the contents of XXH3_kSecret, such as a defined macro to replace the default contents of XXH3_kSecret.

easyaspi314:
…wut

"Secret" is merely a way to separate the 64-bit seeded option from a long set of keys by name.

PR #407, opened 2020-06-24
Switch mixer to XXH128

Cyan4973:
So far, `XXH3` and `XXH128` have been using a related yet slightly different mixer.
The main reason is that `XXH128` must spread ingestion over a 128-bit accumulator width, 
while `XXH3` only needs 64-bit, which is easier and slightly faster.

This PR simplifies the mixer, by always using `XXH128` mixer.
This choice comes at the cost of a bit of top speed performance for `XXH3` (~10%).

However, it also provides the following benefits : 
- There was a small gradual degradation in `XXH3` collision ratio, from 64-bit towards 62-bit, when input length increases beyond > 1 KB (it was previously believed that collision ratio would remain stable whatever the input length beyond 256). This PR fixes it.
- Simpler code
- Smaller produced binary (variable impact, depending on compiler settings).
- Improves performance for small binaries (`-Oz`), by removing a branch within the critical loop (most visible in `arm`).

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -1754,7 +1754,7 @@ XXH3_hashLong_64b_internal(const xxh_u8* XXH_RESTRICT input, size_t len,
 {
     XXH_ALIGN(XXH_ACC_ALIGN) xxh_u64 acc[XXH_ACC_NB] = XXH3_INIT_ACC;
 
-    XXH3_hashLong_internal_loop(acc, input, len, secret, secretSize, XXH3_acc_64bits, f_acc512, f_scramble);
+    XXH3_hashLong_internal_loop(acc, input, len, secret, secretSize, XXH3_acc_128bits, f_acc512, f_scramble);
 
     /* converge into final hash */
     XXH_STATIC_ASSERT(sizeof(acc) == 64);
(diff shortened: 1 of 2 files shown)


easyaspi314:
Suggestion: Change the breakoff point to `XXH3_hashLong_internal_loop` in the dispatcher. This allows us to reduce the number of dispatch functions we need.

Cyan4973:
> Change the breakoff point to `XXH3_hashLong_internal_loop` in the dispatcher

I've looked at it, and I believe I don't fully understand how it can be done (or maybe I misinterpret the objective).

`XXH3_hashLong_internal_loop()` is invoked by both `XXH3_hashLong_64b_internal()` and `XXH3_hashLong_128b_internal()`.
But these 2 functions are not the starting point of the dispatch chain. They are themselves invoked multiple times, hence instantiated multiple times, depending on caller's choice (default `secret` ? `seed` ? dynamic instruction set dispatcher ?). 
I don't see an easy way to break that chain in order to generate a single (or just less) function(s).

Maybe you have a better way or something else in mind.
That being said, as we are getting close to `v0.7.4`, I'm eager to keep it simple, to ensure the release date.
I will be happy to review a patch on this suggested optimization after this is done.

PR #411, opened 2020-06-27
Fix compiler warning for 32 bit architectures

ellert:
In file included from xxhash.h:2065,
                 from xxhash.c:43:
xxh3.h: In function 'XXH3_len_9to16_128b':
xxh3.h:2278: warning: integer constant is too large for 'long' type


Cyan4973 (approved):

Cyan4973:
Thanks @ellert !

On a side note, I'm surprised we did not found this issue earlier, as we are testing 32-bit compilation in CI.
That being said, neither `clang` nor `gcc` seem to complain, at least with our set of flags.
Which compiler are you using ?

PR #413, opened 2020-06-28
Fix C99isms in XXH3_initCustomSecret_avx512

easyaspi314:
Removed `//` comments and a compound literal.


Cyan4973 (approved):

Cyan4973:
Thanks @easyaspi314 .

Maybe a meta-lesson here : we might be missing a test able to catch those.

PR #414, opened 2020-06-28
xxh*sum: print correct default algo for symlinked helpers

martinetd:
Printing xxh128sum's help for example would say default algo is xxh64,
which is not correct. Adjust g_defaultAlgo accordingly as well as the
currently selected algorithm.

-----
When doing a PR is just as fast as opening an issue I tend to just go ahead with it, but feel free to close this and fix it otherwise (or tell me it's a feature :P) -- removing the const attribute is a bit of a shame.

Thanks! Can't wait for 0.8 to be out.


Cyan4973 (approved):

Cyan4973:
Thanks @martinetd , 
I agree, it feels proper that `xxh128sum` displays `xxh128` as its default.

PR #416, opened 2020-07-01
WIP: new -withSeed API variant for reusing the custom secret of the long input

gzm55:
In many production scenarios using non-zero seeds, the seeds are almost compile time or runtime constants for consistent results. The new experiment api `XXH3_hashLong_64b_withSeed2` tries to improve performance of the long inputs in these cases by reusing the internal result of `initCustomSecret`.

Usage:
```c
XXH3_state_t seed_state;
XXH3_64bits_reset_withSeed(&seed_state, seed); // init once

while (len = read(data)) {
    result = XXH3_64bits_withSeed2(data, len, &seed_state); // reuse seed_state
}
```

I roughly bench the new api compared to the original one, on a CentOS 8 machine (SSE2/AVX2), compiling with gcc-9 and clang-9, with/without XXH_INLINE_ALL switch, the performance can be steadily  improved over input length from 256B to 100K. So open this WIP pr for further discussion.

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -1832,6 +1832,24 @@ XXH3_64bits_withSeed(const void* input, size_t len, XXH64_hash_t seed)
     return XXH3_64bits_internal(input, len, seed, XXH3_kSecret, sizeof(XXH3_kSecret), XXH3_hashLong_64b_withSeed);
 }
 
+XXH_NO_INLINE XXH64_hash_t
+XXH3_hashLong_64b_withSeed2(const xxh_u8* XXH_RESTRICT const input, const size_t len, const xxh_u8* XXH_RESTRICT const secret)
+{
+    return XXH3_hashLong_64b_internal(input, len, secret, XXH_SECRET_DEFAULT_SIZE, XXH3_accumulate_512, XXH3_scrambleAcc);
+}
+
+XXH_PUBLIC_API XXH64_hash_t
+XXH3_64bits_withSeed2(const void* const input, const size_t len, const XXH3_state_t* const seed_state)
+{
+    if (len <= 16)
+        return XXH3_len_0to16_64b((const xxh_u8*)input, len, XXH3_kSecret, seed_state->seed);
+    if (len <= 128)
+        return XXH3_len_17to128_64b((const xxh_u8*)input, len, XXH3_kSecret, sizeof(XXH3_kSecret), seed_state->seed);
+    if (len <= XXH3_MIDSIZE_MAX)
+        return XXH3_len_129to240_64b((const xxh_u8*)input, len, XXH3_kSecret, sizeof(XXH3_kSecret), seed_state->seed);
+    return XXH3_hashLong_64b_withSeed2((const xxh_u8*)input, len, seed_state->customSecret);
+}
+
 
 /* ===   XXH3 streaming   === */
 
--- a/xxhash.h
+++ b/xxhash.h
@@ -482,6 +482,7 @@ struct XXH64_state_s {
 #  define XXH3_64bits XXH_NAME2(XXH_NAMESPACE, XXH3_64bits)
 #  define XXH3_64bits_withSecret XXH_NAME2(XXH_NAMESPACE, XXH3_64bits_withSecret)
 #  define XXH3_64bits_withSeed XXH_NAME2(XXH_NAMESPACE, XXH3_64bits_withSeed)
+#  define XXH3_64bits_withSeed2 XXH_NAME2(XXH_NAMESPACE, XXH3_64bits_withSeed2)
 
 #  define XXH3_createState XXH_NAME2(XXH_NAMESPACE, XXH3_createState)
 #  define XXH3_freeState XXH_NAME2(XXH_NAMESPACE, XXH3_freeState)
@@ -509,6 +510,15 @@ XXH_PUBLIC_API XXH64_hash_t XXH3_64bits(const void* data, size_t len);
  * Note: seed==0 produces the same results as XXH3_64bits().
  */
 XXH_PUBLIC_API XXH64_hash_t XXH3_64bits_withSeed(const void* data, size_t len, XXH64_hash_t seed);
+
+/*
+ * XXH3_64bits_withSeed2():
+ * This variant try to reuse the custom secret for a seed.
+ * The seed_state is generated from XXH3_64bits_reset_withSeed().
+ * The produced results are same with XXH3_64bits_withSeed2().
+ */
+typedef struct XXH3_state_s XXH3_state_t;
+XXH_PUBLIC_API XXH64_hash_t XXH3_64bits_withSeed2(const void* data, size_t len, const XXH3_state_t *seed_state);
 
 /*
  * XXH3_64bits_withSecret():
--- a/xxhsum.c
+++ b/xxhsum.c
@@ -661,13 +661,21 @@ static U32 localXXH128_stream(const void* buffer, size_t bufferSize, U32 seed)
     return (U32)(XXH3_128bits_digest(&state).low64);
 }
 
+static U32 g_fixseed = 42;
+static XXH3_state_t g_fixseed_state;
+
+static U32 localXXH3_64b_fix_seeded(const void* buffer, size_t bufferSize, U32 seed)
+{
+    (void)seed;
+    return (U32)XXH3_64bits_withSeed2(buffer, bufferSize, &g_fixseed_state);
+}
 
 typedef struct {
     const char*  name;
     hashFunction func;
 } hashInfo;
 
-#define NB_HASHFUNC 10
+#define NB_HASHFUNC 11
 static const hashInfo g_hashesToBench[NB_HASHFUNC] = {
     { "XXH32",             &localXXH32 },
     { "XXH64",             &localXXH64 },
@@ -679,6 +687,7 @@ static const hashInfo g_hashesToBench[NB_HASHFUNC] = {
     { "XXH128 w/secret",   &localXXH3_128b_secret },
     { "XXH3_stream",       &localXXH3_stream },
     { "XXH128_stream",     &localXXH128_stream },
+    { "XXH3_64b_w/seed-2",     &localXXH3_64b_fix_seeded },
 };
 
 #define NB_TESTFUNC (1 + 2 * NB_HASHFUNC)
@@ -994,6 +1003,12 @@ void BMK_testXXH3(const void* data, size_t len, U64 seed, U64 Nresult)
     if (len>0) assert(data != NULL);
 
     {   U64 const Dresult = XXH3_64bits_withSeed(data, len, seed);
+        BMK_checkResult64(Dresult, Nresult);
+    }
+
+    {   XXH3_state_t seed_state;
+        XXH3_64bits_reset_withSeed(&seed_state, seed);
+        U64 const Dresult = XXH3_64bits_withSeed2(data, len, &seed_state);
         BMK_checkResult64(Dresult, Nresult);
     }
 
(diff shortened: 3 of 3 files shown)


easyaspi314:
Not a terrible idea.

I'd personally have it called `withState`.
```c
/*
 * XXH3_64bits_withState():
 *
 * Acts as a single shot function, but uses the initialization values
 * from `state`. Note that `state` is unmodified and it can be
 * called at any time (even during an incremental hashing session).
 *
 * This can be used as a cache for the withSeed variant, like so:
 *
 *    XXH3_state_t* state = XXH3_createState();
 *    XXH3_64bits_reset_withSeed(state, seed);
 *    XXH64_hash_t h1 = XXH3_64bits_withState("Hello", 5, state);
 *    XXH3_freeState(state);
 *    XXH64_hash_t h2 = XXH3_64bits_withSeed("Hello", 5, seed);
 *    assert(h1 == h2);
 */
XXH_PUBLIC_API XXH64_hash_t
XXH3_64bits_withState(const void* input, size_t length, const XXH3_state_t* state);
```

Alternatively, we could make the seed to secret algorithm public, but having two KDF functions is weird (unless we make a special case for size==8), and the seed would still be needed for the short lengths.

gzm55:
caching for seed's secret is not same with just reuse the secret via `withSecret` api. The short inputs for `withSeed` still uses default secret.

Cyan4973:
My understanding of this WIP is that its sole point is to bypass `secret`'s generation when using a `seed`, hence saving some work.

`secret` generation is not free, but it's reasonably fast, and a topic only for "large enough" inputs ( >= 241 bytes). I expect its one-time impact to represent a minor share of total runtime, even for "small" data (~256 bytes), steadily decreasing to insignificant levels as input size get larger.

I haven't tested this PR directly, but I presume it can be correctly estimated by just comparing `default` and `_withSeed` variants with multiple input lengths. Here are some results, using `xxhsum -b` : 

| input length | `default` GB/s | `_withSeed` GB/s | diff |
| --- | --- | --- | --- |
| 256 | 13.0 | 10.35 | 25 % |
| 512 | 16.1 | 13.7 | 17 % |
| 1 KB | 18.5 | 16.5 | 12 % |
| 2 KB | 19.1 | 18.2 | 5 % |
| 4 KB | 19.6 | 19.2 | 2 % |
| 8 KB | 19.6 | 19.4 | 1 % |

This investigation suggests that the difference becomes insignificant after 8 KB. It's also never "huge", with the maximum cost representing 25% speed at 256 bytes.

One has to wonder if this speed impact over a limited range of input lengths is large enough to justify the creation of new prototypes and a new programming pattern for the user to learn.

Cyan4973:
Food for thoughts : 
maybe the `state` could recognize that it was already initialized, using the same seed, and as a consequence, could skip the `secret` initialization stage, on the ground that it's already stored in `state`.

gzm55:
The improvement ratio on `avx512` is about double as @Cyan4973 posted. Build by clang-9 without `INLINE_ALL`, the results is:

| len  | default GB/s | _withSeed GB/s | _withSeed2 GB/s | diff Seed2 to Seed |
| -- | -- | -- | -- | -- |
| 256 | 16. 2| 10.0 | 16.2 | 62% |
| 512 | 23.2 | 16.6 | 24.1 | 45% |
| 1024 | 31.8 | 24.4 | 32.2 | 32% |
| 2048 | 33.5 | 29.9 | 37.4 | 25% |
| 4096 | 39.1 | 36.5 | 41.3 | 13% |
| 8192 | 42.6 | 41.0 | 43.7 | 7% |
| 100K | 45.7 | 45.5 | 45.6 | 0.2% |

gzm55:
> Food for thoughts :
> maybe the `state` could recognize that it was already initialized, using the same seed, and as a consequence, could skip the `secret` initialization stage, on the ground that it's already stored in `state`.

`XXH3_initCustomSecret` is not easily skipped within `XXH3_64bits_reset_withSeed`, which first `memset` the whole `state` to zero. To skip `memset` and `XXH3_initCustomSecret`, we have to reset many caching fields in `state`, such as bufferedSize, nbStripesSoFar. Such way would make the reset_ function more difficult to maintain with a little performance gain.

Cyan4973:
I tested this scenario again, using an `AVX512` laptop around.

> The improvement ratio on avx512 is about double as @Cyan4973 posted.

I can confirm this statement, performance difference between the "default" and `_withSeed()` variants is substantially wider on `AVX512`, remaining perceptible on an extended range of lengths.

| input length | `default` GB/s | `_withSeed` GB/s | diff |
| --- | --- | --- | --- |
| 256 | 6.7| 3.85 | 43 % |
| 512 | 10.8 | 6.7 | 38 % |
| 1 KB | 14.2 | 10.5 | 26 % |
| 2 KB | 16.0 | 13.7 | 15 % |
| 4 KB | 17.9 | 16.2 | 10 % |
| 8 KB | 18.8 | 18.0 | 5 % |
| 16 KB | 19.5 | 19.0 | 3 % |

If such a performance difference was present across most instruction variants, that would probably prove significant enough to require a solution, even one that needs additional function signatures. But as this magnitude seems specific to `AVX512`, I'm still not completely convinced that introducing new prototypes and new programming patterns is the right thing to do, if the benefit is mostly visible on one (still uncommon) instruction set.

> `XXH3_initCustomSecret` is not easily skipped within `XXH3_64bits_reset_withSeed` (...). Such way would make the reset_ function more difficult to maintain with a little performance gain.

I actually went ahead and tested the previously proposed reset variant, which skips `secret` initialization when it detects that the same `seed` was already used in previous initialization. And in my measurements, it fixed all the performance problems, with "default" and `_withSeed()` performing similarly at all input lengths when `seed` remains constant.
Also interesting, the resulting `reset` code doesn't increase in complexity imho.

There are 2 issues though : 
- This solves the performance gap for the _streaming_ variant, but does nothing for the "one shot" variant.
   + The "one shot" variant allocates its values on stack, and drop them across invocations. So it has to re-initialize them every time.
   + A tempting solution would be to use the _streaming_ variant, which has an explicit context parameter, for all usages, including one-shot ones. Unfortunately, it results in a substantial performance drop (compared to the "one shot" variant). This is because it's both more complex and introduces a hard separation between `_update()` and `_digest()` operations. The `_update()` function cannot ensure that it will necessarily be followed by a `_digest()`. So it has to preserve some context, which is a non-trivial operation compared to the hash itself.
- The `state` must be initialized before invoking `reset_withSeed()`. This is because the function now checks the `state`'s content to find out if it was already initialized with the same `seed` value. That's not a problem if the context is created with `XXH3_createState()`, but if the context is merely declared on stack, it's very easy for the user to forget initializing it (or assume that the `reset()` function will do it, since it was working fine so far on an uninitialized `state` area). I can properly document it, but this transfer of responsibility can become a source of issues in the future.

Presuming that the 2nd point can be answered by proper documentation, that still leaves the 1st and most important point opened : "one shot" variants are out of the picture, and cannot benefit from an already initialized `secret` as there is no state left between invocations.

gzm55:
> The "one shot" variant allocates its values on stack, and drop them across invocations. So it has to re-initialize them every time.

To overcome this, for "one shot" seed variants, how about _changing_ the signature by adding an additional `XXH3_state_t *` pointer?

```c
XXH_PUBLIC_API XXH64_hash_t
XXH3_64bits_withSeed(const void* input, size_t length, XXH64_hash_t seed, const XXH3_state_t* state);
```

Then there are three brief branches:
- `state == NULL`, allocates long secrets on stack, same with the current logic.
- `state != NULL && state is reset with the same seed`, re-use the `customSecret` in `state`.
- others, reset the `state` with `seed` before calculating the hash.

Migrating to new signature can simply add a `NULL` argument if do not want to reuse the long seeded secret.

easyaspi314:
That would make it ABI incompatible.

While we are fine adding new functions within reason, changing existing functions is to be avoided at all costs.

Cyan4973:
Indeed, changing the signature of such an important function, which is part of the API to stabilize for next release, isn't a good thing to do. Besides, the current function signature is as simple as it can and should be. Adding a context adds some unnecessary cognitive load to all users, while benefiting only a fraction of the scenarios.

At this stage, if the performance optimization scenario for input lengths in the 256-2K range has to be addressed through a dedicated function signature, I'm more in favor of @easyaspi314 's proposal, which better respects the established coding convention in this library.

Though even this proposal deserves a discussion.
In `XXH3_64bits_withState(const void* input, size_t length, const XXH3_state_t* state);`, the `state` is  a read-only argument, presumed to transport enough information to decide between "default", `_withSeed()` and `withSecret()` variants. And in the case of `_withSeed()`, re-use the pre-computed `secret` in-place.

However, such a prototype usefulness is basically limited to the `_withSeed()` scenario. Beyond that, I'm struggling to find a reason why it would be used. It also requires a full `XXH3_state_t` (~600 bytes), mostly for the purpose of storing a `seed` and a `customSecret` (~200 bytes), which could be considered a bit wasteful.

Another (close) alternative could be this one : 
`XXH3_64bits_withState(XXH3_state_t* state, const void* input, size_t length);`
This is about the same signature as the `update()` function, in which `state` is actively used by the function, not just read from.
This function would be able to use the `accumulator` space to store intermediate values. This would be useful to reduce stack usage, which is fairly important with `XXH3`. I know, accumulators use only 64 bytes (although these are aligned bytes, which can cost up to 128 bytes on `AVX512`). This may seem very little, and for a normal server application it likely is, but there are environments where such spending of stack space is frown upon, like kernel space for example. 
This would give an additional reason for such a variant to exist : not just to accelerate `_withSeed()` on medium-range input length, but also to get control over stack space usage, whenever it's needed. I would argue that it would make such a variant a bit more useful, hence more likely to occupy a slot in the API.

Even in this case, it is still using only half of the state space, so it could be considered a "waste".

Answering this last point, an entirely different alternative would cut the chase with something more directly applicable to the scenario at hand, like for example: 
`XXH3_64bits_withSeedAndSecret(const void* input, size_t length, XXH64_hash_t seed, const void* secret);`
This would achieve exactly what @gzm55 is looking for, while only requiring to allocate what's strictly needed (192 bytes for `secret`). `secret` could be produced by `generateSecret()`, by using a special rule when `customSeedLength == 8`, mimicking `secret` generation happening inside `_withSeed()`.

Of course, the main issue here is that it's primarily intended to help performance on `_withSeed()` variant over a small range of `inputLength` (roughly 256 - 2KB). And there is still a risk that someone would use a completely different `secret` unsynchronized with `seed`, intentionally or not, making the hash comparable with the `_withSecret()` variant on "long" inputs (> 240 bytes), and comparable with the `_withSeed()` variant for shorter ones.
In case it would be intentional, it also asks the question of an equivalence for the streaming variant, since otherwise the result of the function can't be replicated in a streaming scenario.
Of course, one has to wonder _why_ such a hybrid `seed + secret` variant should even exist, what does it offer "more" compared to existing `_withSecret()` (which also allows larger variety of `secret` length).
Finally, one thing I don't like about this direction is that it exposes `XXH3` internals a bit too much to my taste, with the separation between "short" and "large" input becoming highly visible and public (instead of being merely an "internal" implementation detail).

Another direction could be to change the way the one-shot `_withSeed()` variant works, by combining `seed` and `secret` "on-the-fly", instead of pre-calculating a full new `customSecret` at the beginning. This would spread-out the cost, likely being slower on large data, but _maybe_ faster on smaller data (<= 1 KB). However, I haven't even tested it yet as I'm concerned by both the limited speed improvement perspective and the increased complexity required within `accumulate_512()`, which would be unacceptable.

So there you have it, a list of potential follow-up. None of them look like a slam dunk, but several ones are possible, with their share of pros and cons. And they can be discussed.

Cyan4973:
deprecated by #579

PR #420, opened 2020-07-05
Add the ability to parse BSD style --check lines.

WayneD:
Here's a simple way for the --check code to parse BSD style lines. I made the --tag option no longer hidden, since --check works fine with the output.  I also fixed a bug where the --check parser accepted a line that ended after the first space (returning a bogus filename pointer) and I enabled the combining of  --check with --little-endian.

### DIFF
--- a/xxhsum.c
+++ b/xxhsum.c
@@ -1815,7 +1815,7 @@ static ParseLineResult parseLine(ParsedLine* parsedLine, char* line)
 {
     const char* const firstSpace = strchr(line, ' ');
     const char* hash_ptr;
-    int hash_len;
+    size_t hash_len;
     if (firstSpace == NULL || !firstSpace[1]) return ParseLine_invalidFormat;
 
     parsedLine->filename = NULL;
--- a/xxhsum.c
+++ b/xxhsum.c
@@ -1785,7 +1785,7 @@ static CanonicalFromStringResult canonicalFromString(unsigned char* dst,
     size_t i;
     for (i = 0; i < dstSize; ++i) {
         int h0, h1;
-        int j = reverseDigits ? dstSize - i - 1 : i;
+        size_t j = reverseDigits ? dstSize - i - 1 : i;
 
         h0 = charToHex(hashStr[j*2 + 0]);
         if (h0 < 0) return CanonicalFromString_invalidFormat;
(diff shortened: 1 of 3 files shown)


xxhsum.c
```diff
@@ -1811,19 +1813,39 @@ static CanonicalFromStringResult canonicalFromString(unsigned char* dst,
...
+        if (lastSpace - firstSpace < 5) return ParseLine_invalidFormat;
+        if (lastSpace[-1] != '=' || lastSpace[-2] != ' ' || lastSpace[-3] != ')') return ParseLine_invalidFormat;
+        lastSpace[-3] = '\0'; /* Terminate the filename */
+        *firstSpace = '\0';
+        rev = strstr(line, "_LE") != NULL; /* was output little-endian */
+        hash_ptr = lastSpace + 1;
+        hash_len = strlen(hash_ptr);
+        /* NOTE: This currently ignores the hash description at the start of the string.
+         * In the future we should parse it and verify that it matches the hash length.
+         * It could also be used to allow both XXH64 & XXH3_64bits to be differentiated. */
```
Cyan4973, line 1838: 👍

xxhsum.c
```diff
@@ -1811,19 +1813,39 @@ static CanonicalFromStringResult canonicalFromString(unsigned char* dst,
...
     parsedLine->filename = NULL;
     parsedLine->xxhBits = 0;
 
-    switch (firstSpace - line)
+    if (firstSpace[1] == '(') {
+        char* lastSpace = strrchr(line, ' ');
+        if (lastSpace - firstSpace < 5) return ParseLine_invalidFormat;
+        if (lastSpace[-1] != '=' || lastSpace[-2] != ' ' || lastSpace[-3] != ')') return ParseLine_invalidFormat;
+        lastSpace[-3] = '\0'; /* Terminate the filename */
+        *firstSpace = '\0';
```
Cyan4973, line 1832: These actions are modifying the input.
I'm wondering if there could be another way, preserving immutability of input ?
Like a local stack buffer for example ?

WayneD: Why keep it unchanged? It's just the input line. The code doesn't care that it got changed. It's just harvesting information out of it and discarding it for the next line from the file. However, we could create a static buf of the maximum input line size as one way to stash off the name. Or the code could return a strdup of the filename in the ParsedLine filename pointer (that the caller would then free). I'll go with the latter, but also set a flag so that the non-tag format doesn't need to do the malloc/free. (Update: since we need to free the filename pointer which was marked as const, and I just made it always allocated so it's always non-const.)

Cyan4973: Let's keep it at 7032ced649eefec83ad45fa923e72d17a3e99e9a then.

WayneD: OK. I discarded the commits past that point.

PR #423, opened 2020-07-06
Allow multiple hash types per file

WayneD:
Do we really need to reject the mixing of different hash types in a single --check file?  This branch removes that restriction, adds more xxhsum tests to the Makefile (including some --check --little-endian and --check of --tag output), and also fixes a grammatical issue when complaining about improperly formatted lines (It uses "line is" and "lines are" instead of "are" for both).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -220,14 +220,24 @@ test-xxhsum-c: xxhsum
 	LC_ALL=C ./xxhsum nonexistent 2>&1 | grep "Error: Could not open 'nonexistent'"
 	# xxhsum to/from file, shell redirection
 	./xxhsum xxh* > .test.xxh64
+	./xxhsum --tag xxh* > .test.xxh64_tag
+	./xxhsum --little-endian xxh* > .test.le_xxh64
+	./xxhsum --tag --little-endian xxh* > .test.le_xxh64_tag
 	./xxhsum -H0 xxh* > .test.xxh32
+	./xxhsum -H0 --tag xxh* > .test.xxh32_tag
+	./xxhsum -H0 --little-endian xxh* > .test.le_xxh32
+	./xxhsum -H0 --tag --little-endian xxh* > .test.le_xxh32_tag
 	./xxhsum -H2 xxh* > .test.xxh128
-	./xxhsum -c .test.xxh64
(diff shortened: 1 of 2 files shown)


Cyan4973:
Generally speaking, I believe it's more flexible to accept multiple checksum formats in the same file.
This seems even more relevant when adopting `--tag` line format.

I presume this restriction was in place in order to reproduce an identical limitation present in `md5sum`, `sha1sum`, etc.

Hence, a good follow-up question is : do we want to reproduce this limitation ?

Maybe it made sense when using the `GNU` format, where the checksum length implies the algorithm used,
but with the `--tag` format, I see no benefit, it seems safe to mix different algorithms together in the same file.

So one last question is : 
is it also acceptable to mix gnu/bsd output conventions together ?
Are there cases where the more terse gnu convention introduce risks of confusion ?

Well, actually, that's already the case: a 32-bit hash could be either `XXH32` or `XXH32_LE`.

Looking at further potential situations, a 128-bit hash implies `XXH128`, not `MD5`.
And a 64-bit hash implies `XXH64`, not `XXH3` (if it becomes available one day).
I guess the solution is that, in the context of `xxhsum -c`, the meaning of a `gnu`-style checksum should be documented in a very strict way.

WayneD:
Indeed. One of the big differences between xxhsum & the sha/md5 tools is that xxh has a single script while the sha tools separates sha1sum from sha256sum (etc).  So, sha256sum -c will reject checking any sha1sums as invalid lines.  For xxhsum, since the one script accepts all the checksum files with -c, I think that it's nice that it can accept any of them together in a single file.

Perhaps in the future the tool could be extended to allow the user to specify a stricter checking experience, such as using `xxsum -H1 -c` to make the tool reject any other checksum types.  I'm not sure that's all that beneficial, but it could be done if it was desired.

What the sha/md5 tools have in common with xxhsum is that they all accept any mix of --tag and non-tag input lines when doing a --check.

As for big/little endian outputs, perhaps the man page should be tweaked to recommend that the user combine --tag with --little-endian because --check works fine with the data without the user needing to remember to re-specify --little-endian, unlike the non-tag output.

PR #426, opened 2020-07-11
fix P9 build

jglaser:
This fixes an error message I encountered while trying to build a downstream package (apache arrow):

```In file included from /ccs/proj/stf006/glaser/rapids/arrow/cpp/src/arrow/array/dict_internal.h:35:0,
                 from /ccs/proj/stf006/glaser/rapids/arrow/cpp/src/arrow/array/builder_dict.cc:22:
/ccs/proj/stf006/glaser/rapids/arrow/cpp/src/arrow/util/hashing.h: In static member function 'static uint32_t arrow::internal::SmallScalarTraits<__vector(4) __bool int>::AsIndex(__vector(4) __bool int)':
/ccs/proj/stf006/glaser/rapids/arrow/cpp/src/arrow/util/hashing.h:495:60: error: cannot convert '__vector(4) int' to 'uint32_t {aka unsigned int}' in return
   static uint32_t AsIndex(bool value) { return value ? 1 : 0; }```

See  https://github.com/apache/arrow/pull/7711

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -362,6 +362,7 @@
 #  endif
 
 #  undef vector /* Undo the pollution */
+#  undef bool
 
 typedef __vector unsigned long long xxh_u64x2;
 typedef __vector unsigned char xxh_u8x16;

### CODE REVIEW

**Comment by Cyan4973** [comment]
Neither `xxh3.h` nor `xxhash.h` define nor use `bool` anywhere.
Consequently, this is not the right place to disable it.

Besides, looking at the code, it's questionable if disabling `bool` is the right thing to do.
To be fair, I'm even surprised the arrow code builds after disabling `bool`, unless there is some kind of "default" type that I'm not aware of, or the `#undef` merely undo a wrong `#define bool` that was shadowing a real `bool` type.

Anyway, this concern seems completely beyond `xxhash` scope.
I recommend looking at this issue from an `arrow` project perspective.

**Comment by easyaspi314** [comment]
Might as well add `#undef pixel` as well to undo all the pollution.

**Comment by wesm** [comment]
`bool` is being defined by altivec.h

https://github.com/gcc-mirror/gcc/blob/master/gcc/config/rs6000/altivec.h#L48

```c
#if !defined(__APPLE_ALTIVEC__) \
    && (!defined(__STRICT_ANSI__) || !defined(__cplusplus))
#define vector __vector
#define pixel __pixel
#define bool __bool
#endif
```

It seems best not to leak these defines

Cyan4973:
OK ! Thanks for the diagnosis !

That's surprising (to me), and annoying. We basically inherit a few leaks from `altivec.h`. I wonder how this code works in situations where the user code already defines or use these names. Especially with the definition of `bool`, which is a [reserved `C99` alias](https://en.wikipedia.org/wiki/C_data_types#stdbool.h), _and_ a c++ type, I would expect this header to cause a lot of troubles for many altivec ports.

If the issue is specific to `altivec.h`, as opposed to `s390intrin.h`, we should probably put the `#undef` right next to the `#include`.
Additionally, if the issue is limited to `gcc`, this could also be guarded by an additional `__GNUC__`test.
Reason is, `bool` could be already defined for perfectly valid reasons (like `stdbool.h`), and this `#undef` is going to alter the behavior of such user code when xxhash is inlined. So we want to limit the impact to situations where it's absolutely required, and does not introduce additional issues.

Also, the code comment could be improved, to "undo the pollution _from altivec.h_", since I initially believed the pollution issue was coming directly from our code, instead of being inherited from a system header.

Finally, we could indeed also `#undef pixel` while at it. It would cover all 3 annoying `#define` within `altivec.h`.

Cyan4973 (changes requested):
move, document and complete the #undef list
inside a conditional code section 
 limited to `gcc` only (not `clang`)
right after `#include altivec.h`.

Cyan4973:
Quick follow up : I was trying to reproduce the issue, so that I can develop a test case that will be able to catch this issue continuously.

Unfortunately, so far, I could not : whatever the value of `bool` or `pixel` I could `#define`,
it's still set exactly the same after including `altivec.h`.

I'm using `gcc`, more specifically `powerpc64-linux-gnu-gcc` `v9.3.0`.
I went to check `altivec.h`'s content, and could find the file in `usr/lib/gcc-cross/powerpc64-linux-gnu/9/include/altivec.h`.

It does contain the definition paragraph : 
```C
#if !defined(__APPLE_ALTIVEC__)
/* You are allowed to undef these for C++ compatibility.  */
#define vector __vector
#define pixel __pixel
#define bool __bool
#endif
```

These `#define` should trigger some redefinition warning. But nothing happens.

Notice that the `#define` paragraph is disabled if `__APPLE_ALTIVEC__` is defined. Since I'm compiling on `linux`, it makes sense that this macro is not expected to be defined.
But, after checking it, it is actually defined.

That's a surprise. Why would `__APPLE_ALTIVEC__` be defined on `linux` ? 
Maybe `__APPLE_ALTIVEC__` doesn't signal an Apple system, but a compiler capability to understand `vector`, `pixel` and `bool` keywords in the context of `altivec` without additional preprocessor logic? And maybe `gcc` has this capability? I don't know.
Anyway, that explains why my test doesn't work. 

Since this test is achieved on `linux` + `gcc`, and doesn't trigger the problem, it begs the question : what is the test configuration (cpu, compiler, flags) that produces this bug ? 
This would be useful in order to employ the correct test case.

jglaser:
Hi Yann, the problem occurred on
```
Linux login4 4.14.0-115.21.2.el7a.ppc64le #1 SMP Thu May 7 22:22:31 UTC 2020 ppc64le ppc64le ppc64le GNU/Linux
```

with `gcc 7.4.0`. The error occurred while building the `arrow`library  (before it was patched in https://github.com/apache/arrow/commit/b02095fa2dcd889fa10e519bb30a8bc58a8d4c31)

```
git clone https://github.com/apache/arrow
# choose an earlier commit....

cd arrow
cd cpp
export ARROW_SRC=$PWD
mkdir -p /tmp/$USER/arrow_build
cd /tmp/$USER/arrow_build
module load boost/1.66.0
cmake -DARROW_PARQUET=ON -DARROW_ORC=ON -D ARROW_PYTHON=ON -DARROW_DATASET=ON -DARROW_CUDA=ON -D ARROW_IPC=ON -D CMAKE_INSTALL_PREFIX=$VIRTUAL_ENV $ARROW_SRC
make -j16 install
cd $ARROW_SRC/python
export PYARROW_WITH_PARQUET=1
export PYARROW_WITH_ORC=1
export PYARROW_WITH_CUDA=1
python setup.py install
```

Cyan4973:
Thanks for details @jglaser .
My local test system doesn't have `gcc-7` for `powerpc`, but does provide `gcc-8` and `gcc-9`.
Both set `__APPLE_ALTIVEC__` during compilation (on `linux`!), circumventing the redefinition problem.

Could it be related to compilation flags ?
I'm using the following flags to trigger the `VSX` code path : 
`-maltivec -mvsx -mpower8-vector`.
Only the last one seems to matter, but it both triggers `VSX` code path _and_ defines `__APPLE_ALTIVEC__`.

jglaser:
These are the flags I extracted from the `src/arrow/CMakeFiles/arrow_objlib.dir/flags.make` (I hope these are the relevant ones).

```
C_FLAGS =  -O3 -DNDEBUG   -Wall -maltivec -O3 -DNDEBUG -fPIC   -std=gnu11

C_DEFINES = -DARROW_EXPORTING -DARROW_HDFS -DARROW_JEMALLOC -DARROW_JEMALLOC_INCLUDE_DIR="" -DARROW_WITH_BACKTRACE -DARROW_WITH_LZ4 -DARROW_WITH_SNAPPY -DARROW_WITH_TIMING_TESTS -DARROW_WITH_ZLIB -DARROW_WITH_ZSTD -DURI_STATIC_BUILD

C_INCLUDES = -I/tmp/glaser/arrow_build/src -I/ccs/proj/stf006/glaser/rapids/arrow/cpp/src -I/ccs/proj/stf006/glaser/rapids/arrow/cpp/src/generated -isystem /ccs/proj/stf006/glaser/rapids/arrow/cpp/thirdparty/flatbuffers/include -isystem /autofs/nccs-svm1_proj/stf006/glaser/envs/rapids/include -isystem /tmp/glaser/arrow_build/jemalloc_ep-prefix/src -isystem /tmp/glaser/arrow_build/src/rapidjson_ep/include -isystem /ccs/proj/stf006/glaser/rapids/arrow/cpp/thirdparty/hadoop/include -isystem /tmp/glaser/arrow_build/orc_ep-install/include 

CXX_FLAGS =  -Wno-noexcept-type  -fdiagnostics-color=always -O3 -DNDEBUG  -Wall -maltivec  -O3 -DNDEBUG -fPIC   -std=gnu++11
```

jglaser:
If these flags don't reproduce the behavior, and if you have a small test case, I'll be happy to run it here (on Summit).

Cyan4973:
> I hope these are the relevant ones

Yes, it seems to be relevant.
I see they only use `-maltivec`.
Interestingly, on my test system, `-maltivec` is _not enough_ in order to enable the `VSX` code path support ,
and when I try to force my way into this code path by setting `XXH_VECTOR=XXH_VSX`, it doesn't end well : 
`../github/xxh3.h:1407:9: note: builtin ‘__builtin_vec_rl’ requires builtin ‘__builtin_altivec_vrld’`

Cyan4973:
> If these flags don't reproduce the behavior, and if you have a small test case, I'll be happy to run it here (on Summit).

Sure, here is a possible test : 

```C
#define XSTR(x) STR(x)
#define STR(x) #x

#define BOOL_VALUE 32123456
#define bool BOOL_VALUE

#define VECTOR_VALUE 374464784
#define vector VECTOR_VALUE

#if defined(__APPLE_ALTIVEC__)
#  warning "__APPLE_ALTIVEC__ is defined"
#endif

#define XXH_INLINE_ALL  // required, to pollute the namespace
#include <xxhash.h>   // or any other place where `xxhash.h` and `xxh3.h` are

#if (bool != BOOL_VALUE)
#  error "bool macro was redefined !"
#else
#  pragma message "warning: bool = " XSTR(bool)
#endif

int g_nonEmptyUnit = 0;
```

So far, I can make it fail if I manually `#undef __APPLE_ALTIVEC__` before the `#include`, 
but one has to wonder which real-world use-case am I testing if I have to mess with macro definitions.

jglaser:
This is what I get:
```
(rapids) [glaser@login3.summit glaser]$ gcc -o test test.c -I$RAPIDS_SRC/arrow/cpp/src  -I/gpfs/alpine/scratch/glaser/stf006/rapids/arrow/cpp/src/arrow/vendored/xxhash/ -maltivec
test.c:11:4: warning: #warning "__APPLE_ALTIVEC__ is defined" [-Wcpp]
 #  warning "__APPLE_ALTIVEC__ is defined"
    ^~~~~~~
test.c:18:4: error: #error "bool macro was redefined !"
 #  error "bool macro was redefined !"
    ^~~~~
```

Looks like this is either gcc version or system dependent?

jglaser:
and just a guess, this might be vestige from the times when Macs had Power processors??

Cyan4973:
> This is what I get

The test detects the macro redefinition, which is a good thing.
However, it also shows that `__APPLE_ALTIVEC__` is defined on your system too, like mine,
yet it _doesn't prevent_ the redefinition issue.

Consequently, I suspect that the source code of `altivec.h` is probably different between our 2 systems.
It would be interesting to compare.
The content of mine is [here for gcc-9](https://gist.github.com/Cyan4973/ccaa71def08f819c5ee67228fcf50355) and [here for gcc-8](https://gist.github.com/Cyan4973/1f87f02b2c96a37470205baaf12e1407).

> just a guess, this might be vestige from the times when Macs had Power processors??

From the code comment, my understanding is that Mac compiler was apparently interpreting `vector`, `bool` and `pixel` as keywords in the context of altivec on Mac. Presumably, `gcc`'s `altivec.h` tries to emulate the same behavior with macros.
However, in the process, it becomes incompatible with any source code which requires one of these keyword for any purpose,
such as `C++`'s `bool`.

> Looks like this is either gcc version or system dependent?

Note : I still haven't got a `gcc-7` version to test. But I have a `gcc-5` version on an older VM.
On this version, `altivec.h` already features a `#if !defined(__APPLE_ALTIVEC__)` condition.

Cyan4973:
@jglaser, I'm not sure if the version of `altivec.h` on your system is the same as the one linked by @wesm ,
but if it is, I note that the condition to disable the `#define` block is different : 

```
#if !defined(__APPLE_ALTIVEC__) \
    && (!defined(__STRICT_ANSI__) || !defined(__cplusplus))
```

If I do understand it correctly, a defined `__APPLE_ALTIVEC__` alone should nonetheless be enough to disable the `#define` block.

The test on your system states that `__APPLE_ALTIVEC__` is defined,
yet it seems that `bool` definition is nonetheless overwritten.
So maybe `altivec.h` is different on your system.

I also note that @easyaspi314 added a `#undef vector` as a partial mitigation, which presumes that he encountered namespace pollution issues during development.

Cyan4973:
@jglaser, the reason I'm asking for the content of `altivec.h` on your system, presuming it comes from `gcc` and is therefore open source, is that it may change the nature of the solution.

The current solution, using `#undef`, is incorrect imho.
It disables preprocessor entry points for `bool` or `vector`, irrespective of the fact that maybe the user program has set them for a good reason. For example, in `C99`, `bool` is a preprocessor keyword (`_Bool` is the actual type). So `#undef bool` would disable it.

The better solution is to avoid redefining `bool`. For that, it seems enough, so far, to `#define __APPLE_ALTIVEC__`, as it disables the code segment which adds those unwanted `#define`.

At least, that's the case on all versions of `gcc`'s `altivec.h` I could find (while `clang` versions don't have `#define` issues to begin with).

The problem is that, according to your reported results of the test file, `#  warning "__APPLE_ALTIVEC__ is defined"`, but nonetheless `error "bool macro was redefined !"`.

So, unless one of these files was modified in some ways (did you modified `xxh3.h` or `xxhash.h` for example ?), it points to the fact the `#define __APPLE_ALTIVEC__` was not enough to prevent the `#define bool` from happening. This radically changes the solution required.

jglaser:
Here's the relevant part of the header

```#ifndef _ALTIVEC_H
#define _ALTIVEC_H 1

#if !defined(__VEC__) || !defined(__ALTIVEC__)
#error Use the "-maltivec" flag to enable PowerPC AltiVec support
#endif

/* If __APPLE_ALTIVEC__ is defined, the compiler supports 'vector',
   'pixel' and 'bool' as context-sensitive AltiVec keywords (in 
   non-AltiVec contexts, they revert to their original meanings,
   if any), so we do not need to define them as macros.  */

#if !defined(__APPLE_ALTIVEC__)
/* You are allowed to undef these for C++ compatibility.  */
#define vector __vector
#define pixel __pixel
#define bool __bool
#endif
```

Cyan4973:
Thank @jglaser ,
so it seems that it's the same code segment,
therefore a `#define __APPLE_ALTIVEC__` should be enough to ensure that
`bool`, `pixel` and `vector` are not #defined within `altivec.h`.

Yet, the test result states that `__APPLE_ALTIVEC__` was defined, and nonetheless `bool macro was redefined`.

The only explanation I can think of is that you may have been testing with a modified version of `xxhash`,
such as the one in this PR, which adds an `#undef bool` statement, which would make the test fail when testing `bool`.

jglaser:
Ahh.... that might be... yikes

easyaspi314:
Side note: Was testing on actual hardware, and the VSX path has -O3 memes on old GCC/IBM XL versions. Investigating.

Edit: It is strict aliasing memes with our xacc cast. Using vec_xl and vec_xst seems to fix it

Cyan4973:
> Ahh.... that might be... yikes

And unfortunately, it's not the end of the topic.

Now, let's presume that `xxh3.h` code is not modified, and as a consequence, the test passes.

Which means, the test detects that `__APPLE_ALTIVEC__` is automatically #defined by the compiler,
hence the `#define` paragraph is disabled,
hence `bool` is _not_ redefined,
hence there is no problem with `bool`,

the question is : how come you even met this problem to begin with ?
The test just established that it should _not_ happen, yet this PR starts with the premise that it _did_ happen.

That's why it is important properly reproduce and then observe the issue.
We need to establish why it happens,
so that we can fix it properly, without breaking other scenarios
(which the `#undef` solution will).

_Speculation:_
One of the possible issues I can think of is that
while the test uses a C source file, hence likely triggers a C compiler,
the apache arrow project may be using C++ instead,
which could result in some difference regarding the `__APPLE_ALTIVEC__` macro.

_Note_ : I just tried it, using `powerpc64-linux-gnu-g++` cross-compiler, but it did not change the outcome : `__APPLE_ALTIVEC__` macro was still properly defined by the compiler, so no `bool` redefinition issue detected.

easyaspi314:
As for an alternative solution for `bool`:
```c
/* in case it was defined by stdbool.h */
#undef bool
#include <altivec.h>
/* Clean up the excess macros that it may have defined */
#undef vector
#undef pixel
#undef bool 
/* Redefine bool if stdbool.h was included  */
#ifdef __bool_true_false_are_defined
#  define bool _Bool
#endif
```

Cyan4973:
> As for an alternative solution for `bool`:

This works specifically for `C99`'s `stdbool.h`. Unfortunately, this is a bit too narrow : we can't rule out that one of these keywords is used for a purpose we can't guess in advance. A foolproof solution can be done, but is quite a bit heavier than that.

A better solution would be to ensure that these redefinitions do _not_ happen.
For this, it seems enough to disable the `#define` code block, which can be done by defining `__APPLE_ALTIVEC__`.
But tests so far have established that `__APPLE_ALTIVEC__` is already defined,
so we need to reproduce a scenario where `bool` is redefined and becomes a problem,
like the one mentioned above,
so that we can observe if `__APPLE_ALTIVEC__` was defined or not in this case.
It influences the selected solution.

Cyan4973:
Hi @jglaser 

In #436, I've tried to put together an update of `xxhash` which is expected to solve your compilation issue with apache/arrow on P9.
Since I can't reproduce it directly, there is always a risk that it may not work.
Would you be so kind as testing it on your target system ?

PR #430, opened 2020-07-14
Support '--' to signify end of options

### LINKED ISSUE #428: Parse args like shasum
Since xxhash is cross-platform compatible, I assume it tries to stay away from using `getopt`, which `shasum` uses (see coreutils `src/md5sum.c` main func).  The current argument parsing of xxhsum has some bugs and some enhancements it could use:

1. BUG: Don't double-interpret arguments as option and filename
2. ENHANCEMENT: Interpret `--` as end of option parsing

For 1., try

```sh
touch file file2
./xxhsum file >file.xxhsum
./xxhsum file.xxhsum --check
```

and get

```
file: OK
Error: Could not open '--check': No such file or directory
```

Now, try

```sh
./xxhsum file2 >--check
./xxhsum file.xxhsum --check
```

and get

```
file: OK
file2: OK
```

I'm pretty sure we don't want to be double-interpreting `--check` as an option and filename.

For 2., try

```sh
touch -- -_-
./xxhsum -_-
```

and get

```
Wrong parameters
...
```

Try

```sh
./xxhsum -- -_-
```

and get

```
Wrong parameters
...
```

So there appears to be no way to run `xxshum` on a file that begins with `-`.  Sure, we can pass it through stdin (e.g. `cat -_- | xxhsum`), but that's not ideal (and you also need to attach the filename to the output).

I tried working on these myself, but it's nontrivial because the current code mostly uses an `int` variable `filenamesStart` to determine where to start processing filenames.  I don't really have experience with `getopt`, but I assume it would be the best solution (but then, we need to find an alternative for Windows systems, use `#ifdef _WIN32`, etc.).

Here are other examples of what `shasum` does along with their not-working `xxhsum` equivalents in comments:

```sh
touch a b
shasum a -- b # xxhsum a -- b
```

```
da39a3ee5e6b4b0d3255bfef95601890afd80709  a
da39a3ee5e6b4b0d3255bfef95601890afd80709  b
```

```sh
touch -- --
shasum -- -- # xxhsum -- --
```

```
da39a3ee5e6b4b0d3255bfef95601890afd80709  --
```

```sh
touch a b
shasum a --algorithm 224 b # xxhsum a -H32 b
```

```
d14a028c2a3a2bc9476102bb288234c415a2b01f828ea62ac5b3e42f  a
d14a028c2a3a2bc9476102bb288234c415a2b01f828ea62ac5b3e42f  b
```

jaki:
When parsing arguments, the first non-option argument is treated as the
start of a list of file names.  However, arguments continue to be
parsed, so any later arguments that look like options will have a dual
meaning as an option and file name.  Fix this by treating later
option-like arguments strictly as file names.

Also interpret a double dash (i.e. `--`) as the point where future
arguments are treated as file names.  This is useful for files that
begin with a dash (e.g. `./xxhsum -- -foo`).  Note that `--` itself can
be interpreted as a file (e.g. `./xxhsum foo --`).

Close: #428

### DIFF
--- a/xxhsum.c
+++ b/xxhsum.c
@@ -2322,11 +2322,11 @@ static int XXH_main(int argc, const char* const* argv)
 
         if (!strcmp(argument, "--")) {
             if (filenamesStart==0 && i!=argc-1) filenamesStart=i+1; /* only supports a continuous list of filenames */
-            break;
+            break;  /* treat rest of arguments as strictly file names */
         }
         if (*argument!='-') {
             if (filenamesStart==0) filenamesStart=i;   /* only supports a continuous list of filenames */
-            break;
+            break;  /* treat rest of arguments as strictly file names */
         }
 
         /* command selection */
--- a/xxhsum.c
+++ b/xxhsum.c
@@ -2322,7 +2322,7 @@ static int XXH_main(int argc, const char* const* argv)
 
         if (*argument!='-') {
             if (filenamesStart==0) filenamesStart=i;   /* only supports a continuous list of filenames */
-            continue;
+            break;
         }
 
         /* command selection */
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):

Cyan4973:
Thanks @jaki !

PR #432, opened 2020-07-19
[WIP] Split up xxhsum

easyaspi314:
I am currently working on splitting xxhsum into multiple files.

xxhsum.c slowly but steadily grew from a few hundred lines to a few thousand over the years, changing from a simple benchmark/selftest to an entire command line utility. It is incredibly disorganized and hard to follow.

 Unlike xxHash's implementation which needs to be in one file, xxhsum does not, so I am splitting it into multiple files to extract the inline feature tests, Windows wrappers, benchmark code, etc.

Nothing builds yet, just laying things out right now.


Cyan4973:
I'm sympathetic to the objective. It's true that `xxhsum.c` has become large overtime, and that splitting it into more focused units is a good move to tackle code complexity.

In term of strategy to get there, I can see 2 broad families : 
- Incremental evolution : this involves splitting some scope out of `xxhsum.c`, one at a time, and making sure that the resulting repository continues to run fine (compilation and tests). This reduces risks, and make each review more focused, but can require more work and progress slower due to the need to ensure that each step produces a valid repository.
- Big bang : the whole file is split into multiple units in a single operation. This skips all intermediate steps, for the benefit of execution speed. However, it also means that the merge step has a huge scope, and there are a lot of questions to analyze and code fixes at review time. It's also likely that a few details will be missed due to the sheer amount of code to review and consequences to understand. Also, long reviews may prevent the non-controversial part of the work to be merged for some time, since it's an all or nothing decision.

While I'm fine with both approaches, it's good to be aware of the pros and cons of each one. I would instinctively tend to lean on the first strategy, merely because I'm risk-averse and worried of protracted debate at review time, but I don't make it a condition.

Cyan4973:
Minor comments : 
- my understanding is that this PR is just "to see", it's not meant to be merged. This explains why it feels incomplete, with several sections of `xxhsum` still missing.
- I noticed some "generic" names, such as `util`. While it can be convenient to just have a "bag of everything", be aware that it's very unspecific for any other code contributor. Whenever possible, I prefer more specific unit names, presuming a corresponding scope can be defined, like `displayControl` or `windowsUnicode` (for example).

easyaspi314:
Ok so I will switch to a "split it up gradually" thing. I tend to get ahead of myself sometimes. 

I'll start again and do it step by step.

easyaspi314:
Ok I am planning this out.

Here is my idea. It splits things into little digestible chunks, and makes it so all symbols start with `XSUM_` (this allows it to be easily compiled into BusyBox or something).

 - `config.h`: arch/compiler/feature detection macros
 - `stdio-helpers.c/h`:
   - Variadic macro → `XSUM_vfprintf` implementation
   - `DISPLAY` → `XSUM_log`
   - `DISPLAYLEVEL` → `XSUM_logVerbose`
   - `DISPLAYRESULT` → `XSUM_display`
   - `g_displayLevel` → `XSUM_verbosityLevel`
   - `IS_CONSOLE` →  `XSUM_isConsole`
   - `XSUM_isDirectory`
 - `utf16-helpers.c/h`: 
   - `utf8_to_utf16_len` → `XSUM_widenString`
   - `utf16_to_utf8_len` → `XSUM_narrowString`
 - `main.c`: 
   - `XXH_main` → `XSUM_main`
   - `main`/`wmain`
   - `XXH_wmain` → `XSUM_wmain`
   - Command line parsing helpers
 - `bench.c`:
   - `BMK_benchInternal` → `XSUM_benchInternal`
   - `BMK_benchFiles` → `XSUM_benchFiles`
 - `selftest.c/h`
   - `BMK_sanityCheck` → `XSUM_selfTest`
 - `hasher.c/h`:
   - `XSUM_hashFiles`
 - `checker.c`:
   - `checkFiles` → `XSUM_checkFiles`
   - `getLine` → `XSUM_getLine`

Is this a good plan? I still have to map them out more of course.

And as for the process:
 1. Rename symbols
 2. Split up one file at a time
 3. Leave `xxhsum.c` as a stub which `#includes` all the `.c` files.

Cyan4973:
Yes, this goes into the right direction. 
The proposed new naming scheme seems fairly good. And I like the idea to preserve code amalgamation capability through `xxhsum.c`.

A few other thoughts on this topic : 
- Group all files related to the CLI into their own folder. This is meant to make the separation between the CLI program and `libxxhash` library much clearer.
- Avoid "too generic" file names, like `config.h` or `main.c`. This is bound to develop issues in the future for some build scenarios. Make these names a little bit more specific. If uninspired, even just a prefix would be enough (like `xsum_config.h` for example).
- Try to isolate character set issues from the command line parser proper. Things like Windows's Unicode support `XXH_wmain()` could be isolated in their own units, so that systems for which this support is irrelevant could entirely ignore said unit.

So mostly good, I'm sure we'll find more stuff along the way, and we'll adjust the details on the go.

easyaspi314:
Yeah for the folders we could probably move the CLI tools into a folder called "tools" since that makes more sense than "tests"

```
tools/
   |-- benchmark
   |-- collisions
   '-- xxhsum
       '--- xxhsum-tests (from the Makefile)
```

Cyan4973:
I would rather keep `tests` and `xxhsum` separated.

`xxhsum` is a core product delivered by this project.

easyaspi314:
Well I would name it `xxhsum/` but that would screw over any non-Windows user who runs `git pull` before `make clean`.

I used `src` but having source files outside `src` is weird.

Ideas:
 - `xxhsum-src`
 - `cli`
 - `driver`
 - `frontend`

Do you have a good name for it?

Cyan4973:
A "generic" name I use for these cases is `programs`.
Is also like your suggestions `cli` and `frontend`.

easyaspi314:
Started renaming the symbols in a force push.

Cyan4973:
@easyaspi314 , please ping me when you believe the PR is ready for a review and a merge
(it doesn't have to reach the final objective, it can also be a complete-enough intermediate stage worth integrating by itself).

easyaspi314:
I think this is a good start. Merging (or splitting?) this will give more incentive to split the rest. 

Waiting on CI for Windows checks, don't have my Windows PC available right now. 

I also made some simplifications — the `fprintf` wrapper now extracts the effective `vasprintf` call, and I fixed a few small bugs like #451 and the usage of an ANSI api on Windows (`_stat64`).

easyaspi314:
Additionally, added a few config options:
 - `XSUM_WIN32_USE_WCHAR`: Whether to use the `wchar_t` APIs on Windows instead of the 8-bit "ANSI" versions. Defaults to `1` on Windows, `0` on non-Windows. Requires `_WIN32` to be defined.
 - `XSUM_WIN32_USE_WMAIN`: Whether to use `wmain` or `main+__wgetmainargs`. Defaults to `1` on MSVC and MinGW-w64 with the `-municode` option, `0` otherwise. Requires `XSUM_WIN32_USE_WCHAR=1` or it will be disabled.
 - `XSUM_NO_MAIN`: Doesn't define `main`, the entry point is instead `XSUM_main`, allowing it to easily be integrated into a multi-call binary.

easyaspi314:
I think I fixed the errors now. 
Makefile dependencies are obnoxious.

Additionally, compiling xxhsum.c with XXH_INLINE_ALL now includes the split .c files.

easyaspi314:
`__attribute__((__format__))` is a pain.

easyaspi314:
I think I fixed all the typos. 

BTW, Appveyor for mingw64 doesn't seem to be doing anything. :thinking:

easyaspi314:
Once tests pass and this is merged, I am going to do individual component PRs from now on instead of getting carried away again.

easyaspi314:
@Cyan4973 How is this so far? I think this is a good stopping point for now.

.gitignore
```diff
@@ -13,6 +13,7 @@ xxh32sum
 xxh64sum
 xxh128sum
 xxhsum
+!programs/xxhsum
```
Cyan4973: Do you plan on having multiple program targets in `programs/`  ?
If not, I do not find it beneficial to define a second directory level below `programs/`.

Makefile
```diff
@@ -71,6 +71,13 @@ endif
 
 LIBXXH = libxxhash.$(SHARED_EXT_VER)
 
+XXHSUM_SPLIT_SRCS = programs/xxhsum/xsum_os_specific.c \
```
Cyan4973: try to define something like : 
`XXHSUM_DIR = programs/xxhsum`
so that it's easier to update the directory

.gitignore
easyaspi314: Hmm
I don't plan on it.

What about another path idea, `bin`?

Cyan4973: `bin/` is usually employed to host the generated artifacts, not the sources.

Coming back to `programs/xxhsum/`,
I believe we already had this discussion, and there were many candidates to choose from.
I'm fine with most of them, be it `xxhsum-src/`, `cli/`, `frontend/`, or even `programs/` (directly).
`xxhsum/` would have been great, though it's apparently a problem for `Windows`.

easyaspi314: Well I am not sure now. Tell me what you want, it is your decision.

Also, `xxhsum/` would be a problem for non-Windows, as if one ran `git pull` before `make clean`, `git` will try to overwrite the *executable* `xxhsum` with the *folder* `xxhsum/`.

```
$ make
...
cc xxhsum.o xxhash.o -o xxhsum
$ file xxhsum
xxhsum: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, not stripped
$ git pull
...
error: The following untracked working tree files would be overwritten by merge:
        xxhsum
        xxhsum
        xxhsum
        xxhsum
        xxhsum
        xxhsum
Please move or remove them before you merge.
Aborting

```

Cyan4973: OK, let's go for `cli` then

easyaspi314:
Looks good?

Makefile
```diff
@@ -71,6 +71,14 @@ endif
 
 LIBXXH = libxxhash.$(SHARED_EXT_VER)
 
+XXHSUM_SRC_DIR = cli
```
Cyan4973, line 74: Yes, much better.
These variables really help maintenance and readability of the `Makefile` script

Cyan4973:
Yep, been through it, it's a pretty good rewrite, good to go

Cyan4973:
It seems this PR has accidentally broken `xxhsum` compile-time cpu detection.

Before this PR : 
```
compiled as 64-bit x86_64 + SSE2 little endian with Apple Clang 11.0.3 (clang-1103.0.32.62)
```
After this PR : 
```
compiled as 64-bit unknown little endian with Apple Clang 11.0.3 (clang-1103.0.32.62)
```

Cyan4973:
Issue now fixed in `dev` branch

mukikak61 (commented):
easyaspi314

mukikak61 (commented):
#475

PR #433, opened 2020-07-22
Vsx fixes

easyaspi314:
Fixes issues on older compilers, as well as being a solution to what also happens in #426 

Hasn't been fully tested, but it fixes -O3 on all PPC64LE compilers I had available which previously had issues.

### DIFF
--- a/xxh3.h
+++ b/xxh3.h
@@ -1426,7 +1426,7 @@ XXH3_accumulate_512_vsx(  void* XXH_RESTRICT acc,
         /* product = ((xxh_u64x2)data_key & 0xFFFFFFFF) * ((xxh_u64x2)shuffled & 0xFFFFFFFF); */
         xxh_u64x2 const product  = XXH_vec_mulo((xxh_u32x4)data_key, (xxh_u32x4)shuffled);
         /* acc_vec = xacc[i]; */
-        xxh_u64x2 acc_vec        = vec_xl(16 * i, xacc);
+        xxh_u64x2 acc_vec        = vec_xl(0, xacc + 2 * i);
         acc_vec += product;
 
         /* swap high and low halves */
@@ -1436,7 +1436,7 @@ XXH3_accumulate_512_vsx(  void* XXH_RESTRICT acc,
         acc_vec += vec_xxpermdi(data_vec, data_vec, 2);
 #endif
         /* xacc[i] = acc_vec; */
-        vec_xst(acc_vec, 16 * i, xacc);
+        vec_xst(acc_vec, 0, xacc + 2 * i);
     }
 }
 
@@ -1454,7 +1454,7 @@ XXH3_scrambleAcc_vsx(void* XXH_RESTRICT acc, void const* XXH_RESTRICT secret)
         size_t i;
         for (i = 0; i < XXH_STRIPE_LEN / sizeof(xxh_u64x2); i++) {
             /* xacc[i] ^= (xacc[i] >> 47); */
-            xxh_u64x2 const acc_vec  = vec_xl(16 * i, xacc);
+            xxh_u64x2 const acc_vec  = vec_xl(0, xacc + 2 * i);
             xxh_u64x2 const data_vec = acc_vec ^ (acc_vec >> v47);
 
             /* xacc[i] ^= xsecret[i]; */
@@ -1466,8 +1466,8 @@ XXH3_scrambleAcc_vsx(void* XXH_RESTRICT acc, void const* XXH_RESTRICT secret)
             xxh_u64x2 const prod_lo = XXH_vec_mulo((xxh_u32x4)data_key, prime);
             /* prod_hi = ((xxh_u64x2)data_key >> 32) * ((xxh_u64x2)prime >> 32);  */
             xxh_u64x2 const prod_hi = XXH_vec_mule((xxh_u32x4)data_key, prime);
-            xxh_u64x2 const product = prod_odd + (prod_even << v32);
-            vec_xst(product, 16 * i, xacc);
+            xxh_u64x2 const product = prod_lo + (prod_hi << v32);
+            vec_xst(product, 0, xacc + 2 * i);
     }   }
 }
 
--- a/xxh3.h
+++ b/xxh3.h
@@ -355,13 +355,35 @@
  * inconsistent intrinsics, spotty coverage, and multiple endiannesses.
  */
 #if XXH_VECTOR == XXH_VSX
+/*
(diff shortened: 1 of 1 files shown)


easyaspi314:
…nobody saw that typo

Cyan4973:
A few comments : 
- The PR seems to bundle fixes for 2 issues : 
   + the namespace pollution (#426) 
   + compatibility with some PPC64LE compilers 

In both cases, we are lacking a reproduction case, a test which would prove that there was a problem before, and it is fixed now.

I also believe both topics could be decoupled. They don't depend on each other, and it would be a pity to _not_ merge one fix because there is an ongoing discussion for the other one.

I can take care of #426, it's relatively "easy" to understand the issue, but there are many possible fixes, with subtle side-effect differences, I guess it's not worth having many back-and-forth for tiny appreciation details.

You are in much stronger position to fix the issue regarding PPC64LE compilers. If that's possible, try to update our test suite to show that there is a problem with current code, and that the patch fixes it (note: for the time being, the PR seems to break existing PPC64 compilation tests on TravisCI).

Finally, there is an important PR ongoing, [`merge_xxh3`](https://github.com/Cyan4973/xxHash/pull/434), which completely change the location of `xxh3` source code, resulting in conflicts with any PR targeting `xxh3` source code. It might be better to wait for it to be merged in `dev` and pickup from there.

_edit_: `xxh3.h` content now merged into `xxhash.h`.

easyaspi314:
Ok, can redo this.

For the reference, this is the issue I encountered.

This is using xlc 13.1.6 and GCC 4.8.5 on ppc64le
```
[easyaspi314@gcc2-power8 xxhash-xxh3]$ xlc --version
IBM XL C/C++ for Linux, V13.1.6 (Community Edition)
Version: 13.01.0006.0001
/opt/ibm/xlC/13.1.6/bin/.orig/xlc: note: XL C/C++ Community Edition is a no-charge product and does not include official IBM support. You can provide feedback at the XL on POWER C/C++ Community Edition forum (http://ibm.biz/xlcpp-linux-ce). For information about a
 fully supported XL C/C++ compiler, visit XL C/C++ for Linux (http://ibm.biz/xlcpp-linux).
[easyaspi314@gcc2-power8 xxhash-xxh3]$ make clean; make CC="xlc -qarch=pwr8"
cleaning completed                                                                      
xlc -qarch=pwr8 -O3     -c -o xxhash.o xxhash.c
./xxhash.h:3428:36: error: 1540-2993 The builtin "__builtin_altivec_vmulouw" is not
      supported.
        xxh_u64x2 const product  = XXH_vec_mulo((xxh_u32x4)data_key, shuffled);
                                   ^
./xxhash.h:2442:24: note: expanded from macro 'XXH_vec_mulo'
#  define XXH_vec_mulo __builtin_altivec_vmulouw
                       ^
1 error generated.
Error while processing xxhash.c.
make: *** [xxhash.o] Error 1
[easyaspi314@gcc2-power8 xxhash-xxh3]$ nano xxhash.h
// Comment out the built-in preprocessor def
[easyaspi314@gcc2-power8 xxhash-xxh3]$ make clean; make CC="xlc -qarch=pwr8"
cleaning completed
xlc -qarch=pwr8 -O3     -c -o xxhash.o xxhash.c
ar rcs libxxhash.a xxhash.o
xlc -qarch=pwr8 -O3   -fPIC  xxhash.c -shared -Wl,-soname=libxxhash.so.0 -o libxxhash.so
.0.8.0
ln -sf libxxhash.so.0.8.0 libxxhash.so.0
ln -sf libxxhash.so.0.8.0 libxxhash.so
xlc -qarch=pwr8 -O3     -c -o xxhsum.o xxhsum.c
xlc -qarch=pwr8 -O3    xxhash.o xxhsum.o  -o xxhsum
ln -sf xxhsum xxh32sum
ln -sf xxhsum xxh64sum
ln -sf xxhsum xxh128sum
[easyaspi314@gcc2-power8 xxhash-xxh3]$ ./xxhsum -b
xxhsum 0.8.0 by Yann Collet
compiled as 64-bit ppc64 + POWER8 vector little endian with Clang 4.0.1 (tags/RELEASE_40
1/final)
Error: 64-bit hash test 100: Internal sanity check failed!
Got 0x1B375308E3E9E32CULL, expected 0xCDEB804D65C6DEA4ULL.
or temporarily comment out the tests in BMK_sanityCheck.
[easyaspi314@gcc2-power8 xxhash-xxh3]$ make clean; make CC="gcc-4.8 -mvsx"
// Snip
[easyaspi314@gcc2-power8 xxhash-xxh3]$ ./xxhsum -b
xxhsum 0.8.0 by Yann Collet
compiled as 64-bit ppc64 + POWER8 vector little endian with GCC 4.8.5 20150623 (Red Hat 4.8.5-39)
Error: 64-bit hash test 105: Internal sanity check failed!
Got 0x686CF2FFE656A0B9ULL, expected 0x6259F6ECFD6443FDULL.
or temporarily comment out the tests in BMK_sanityCheck.
```

Cyan4973:
Interesting, I did not pay attention to `xlc` before .

I wonder if this is something we should consider testing on TravisCI's native `ppc64le` platform ...

easyaspi314:
Update: Compiled GCC 4.8.5 as a cross compiler and ran on qemu. It gets that same `0x686CF2FFE656A0B9ULL` output.

The compiler error that is causing the builds to fail is due to a bug in older GCC versions which don't have the right overloads.

The only solution I found is to use `memcpy`.


But this comes to discussion:

What if we kill the void pointers in the main loop and maintain the types?
 - Fixes most, if not all of the remaining strict aliasing violations.
 - This should work on all code paths safely:
   - Scalar is scalar and uses the types directly
   - SSE2 and AVX2 explicitly allows aliasing to `__m128i *` and `__m256i *` for the load and store intrinsics:
```c
    __m128i acc_val = _mm_load_si128((const __m128i *)(xacc + 2 * i));
    _mm_store_si128((__m128i *)(xacc + 2 * i), acc_val);
```
   - AVX512 finally uses `void *` for its load and store intrinsics
   - NEON has `vld1q_u64` and `vst1q_u64`
   - VSX/S390X is going to be using `memcpy` with my new proposal
 - This makes it unambiguous which pointers are raw and which pointers are native and aligned
 - This removes the need to recast the pointers over and over again making the loop seem more complicated than it is.

This also makes it clear which types are aligned/native endian and which are unaligned/little endian.

Cyan4973:
OK, let's try out

easyaspi314:
Ok...we might have to cast on NEON...LP64 ABI strikes again.

PR #436, opened 2020-07-23
fix altivec.h redefinition issue

Cyan4973:
fix #426


Cyan4973:
Hi @jglaser 

This PR is expected to solve the compilation issue described in #426.
However, while it comes bundled with its own compilation test, it's not directly equivalent to your apache/arrow compilation case.
Would you be so kind as testing it on your target system, to verify that it's indeed fixed ?

PR #437, opened 2020-07-24
Support INCLUDEDIR and LIBDIR in libxxhash.pc

ellert:
The pkgconfig file hardcodes the default locations rather than adapting to the directories used during installation.
This means that:

$ pkg-config --libs libxxhash
-L/usr/lib -lxxhash 

even if the library is installed in /usr/lib64 by setting LIBDIR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,6 +138,8 @@ lib: libxxhash.a libxxhash
 
 pkgconfig:
 	@$(SED) -e 's|@PREFIX@|$(PREFIX)|' \
+		-e 's|@INCLUDEDIR@|$(INCLUDEDIR)|' \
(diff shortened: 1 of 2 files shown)


Cyan4973:
Thanks for pointing that out.
This is a difficult topic, as there are different ways to use `pkgconfig`, and some will insist on have control during installation time, while others will want an ability to redirect some paths later at usage time.
I will change that to use a completely different recipe, based on prior experience from `zstd`, which seems to be compatible with both usage patterns.

Cyan4973:
#438 attempts to fix the issue presented in this PR.

It gives control of installation dir at installation time, through `LIBDIR` and `INCLUDEDIR`,
and preserve ability to change meaning of `prefix` at invocation time.

PR #439, opened 2020-07-24
update generateSecret for customSeedSize==8

Cyan4973:
Change the generated secret when `customSeed` has a size of 8 bytes.
In which case, uses the same initialization function as the `_withSeed()` variant.

Worth noting : for reproducibility across platforms,
the buffer must be interpreted with a specified endianness.
The format selected is the Canonical format, aka big-endian.

This is a preparation step towards a potential solution for #416 .

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -4183,6 +4183,12 @@ XXH3_generateSecret(void* secretBuffer, const void* customSeed, size_t customSee
         return;
     }
     XXH_ASSERT(customSeed != NULL);
+    if (customSeedSize == 8) {
+        const XXH64_canonical_t* const seedCanon = (const XXH64_canonical_t*)customSeed;
+        XXH64_hash_t const seed64 = XXH64_hashFromCanonical(seedCanon);
+        XXH3_initCustomSecret(secretBuffer, seed64);
+        return;
+    }
 
     {   size_t const segmentSize = sizeof(XXH128_hash_t);
         size_t const nbSegments = XXH_SECRET_DEFAULT_SIZE / segmentSize;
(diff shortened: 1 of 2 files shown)


Cyan4973:
OK, so the issue is that
`XXH3_initCustomSecret()` _requires_ destination buffer `secretBuffer`
to be fully aligned on vector width when using recently introduced vector variants `SSE2`, `AVX2` or `AVX512`.

When `XXH3_initCustomSecret()` is only invoked internally, as part of `_withSeed()` initialization, the destination buffer is well controlled, and therefore correctly aligned.

But when it's invoked through `XXH3_generateSecret()`, the destination is whatever the user provide, hence not necessarily aligned.

Multiple possibilities here : 
- Generate `secret` into an internal buffer, correctly aligned, then copy into user-defined destination buffer
   + presumes minor speed and memory hits are not too important for `XXH3_generateSecret()`
- Modify `_avx2` and `_avx512` variants of `XXH3_initCustomSecret()` to not require alignment of destination buffer.
   + check if it translates into any noticeable performance impact when used "normally", inside `_withSeed()`.
- Only use `_scalar*` variant of `XXH3_initCustomSecret()` within `XXH3_generateSecret()` (no alignment requirement)
   + presumes that minor speed wins are less important for `XXH3_generateSecret()`
- Abandon the special mode for `size==8`
   + doesn't help #416

Cyan4973:
I will change the logic to introduce a dedicated secret generator focused on generating the same secret as the produced when using a `seed`. This will avoid strange side effects / complexity when interpreting the `seed` as buffer (differences in endianess typically).

PR #454, opened 2020-09-23
[WIP] Split xxhsum benchmark code, use better timers

easyaspi314:
- [x] Separate code into its own file
- [x] Add a `XSUM_NO_BENCH` option
- [ ] Cleanup
- [ ] Change benchmark to use the more accurate `timefn.c`

That last one is an issue — where should `timefn.c` go? 

I was thinking `/utils/` or `/common`. We could also put some of the functions we reuse there.


Cyan4973:
Good, this rewrite makes the code much clearer imho.

> where should `timefn.c` go?

Good question.

As a possible reference, in other repositories like `lz4` and `zstd`, 
I do have a strict include policy that enforces : `lib <- programs <- tests`,
aka `lib` must not depend on anything outside of `lib`,
`programs` can depend on `lib`, but nothing else.
A user should be able to extract _just_ `lib` and `programs`, and get it to compile `zstd` or `lz4` just fine.
Consequently, if there is something useful for `tests` hosted in `programs` (such as `timefn`),
`tests` has to include it from there.

Now, while this is strictly enforced in `lz4` and `zstd`, 
nothing like that exists yet in `xxhash` (since everything was flat so far).
To be fair, one can already note a few differences : 
we don't have a `lib` directory (`xxhash` lib is directly at root),
`programs` is replaced by `cli`,
and `tests` hosts each project separately into its own subfolder,
while `zstd` and `lz4` store many of them flat (mainly because there are multiple simple one-file test programs).

While we could go into other directions,
I presume it would be simpler for me if all repositories follow the same convention.
I have nothing against `common/` per se, it's a sensible choice,
just it would make `xxhash` repository a bit different in layout from `lz4` or `zstd`
(and lose the property that `cli/` depends only on root).
Not a big deal, it could go either way,
though that's exactly what conventions are for : arbitrate a choice, between otherwise equivalent options, to remain consistent and reduce surprises with other projects following the same conventions.

easyaspi314:
So you are saying put `timefn.c` in `cli`?

Cyan4973:
Yeah, let's start this way.

easyaspi314:
Added `clock_gettime` to `timefn.c`: it is really easy to slip in and covers < C11. This means that output is deterministic on POSIX systems regardless of C11 support.

tests/bench/benchfn.c
```diff
@@ -18,7 +18,7 @@
 #undef NDEBUG            /* assert must not be disabled */
 #include <assert.h>      /* assert */
 
-#include "timefn.h"        /* UTIL_time_t, UTIL_getTime */
+#include <timefn.h>        /* UTIL_time_t, UTIL_getTime */
```
Cyan4973: `timefn.h` is a project's header, not a system one.
It should use `""` double quotes

easyaspi314: Good point. I will fix that. I should just use relative paths, correct?

Cyan4973: correct

Cyan4973:
If you are still interested in this PR objective, I believe it would be a good next step.

Cyan4973:
Well, a lot has changed since this PR was proposed,
in particular, all files related to the CLI, including `xxhsum.c`, are not hosted into the `cli` directory.
But I still believe that this PR has a good objective, so could be updated.

Cyan4973:
Superseded by #614

PR #459, opened 2020-10-08
 Fix xxh_x86dispatch build with MSVC + two cosmetic changes

apankrat:
Hello Yann,

Basically as per title - a patch to fix dispatch build on Windows with MS compiler, coalesce related stuff together under the same #if / #elif and remove duplicate include of <intrin.h>.

Previously, if it wasn't gcc, it would just bail out right away.

### DIFF
--- a/xxh_x86dispatch.c
+++ b/xxh_x86dispatch.c
@@ -78,14 +78,6 @@ extern "C" {
 #include "xxhash.h"
 
 /*
- * Modified version of Intel's guide
- * https://software.intel.com/en-us/articles/how-to-detect-new-instruction-support-in-the-4th-generation-intel-core-processor-family
- */
-#if defined(_MSC_VER)
-# include <intrin.h>
-#endif
-
-/*
  * Support both AT&T and Intel dialects
  *
  * GCC doesn't convert AT&T syntax to Intel syntax, and will error out if
@@ -102,7 +94,6 @@ extern "C" {
 #else
 #  define I_ATT(intel, att) "{" att "|" intel "}\n\t"
 #endif
-
 
 static void XXH_cpuid(xxh_u32 eax, xxh_u32 ecx, xxh_u32* abcd)
 {
@@ -133,6 +124,11 @@ static void XXH_cpuid(xxh_u32 eax, xxh_u32 ecx, xxh_u32* abcd)
     abcd[3] = edx;
 #endif
 }
+
+/*
+ * Modified version of Intel's guide
+ * https://software.intel.com/en-us/articles/how-to-detect-new-instruction-support-in-the-4th-generation-intel-core-processor-family
+ */
 
 #if defined(XXH_DISPATCH_AVX2) || defined(XXH_DISPATCH_AVX512)
 /*
--- a/xxh_x86dispatch.c
+++ b/xxh_x86dispatch.c
@@ -44,8 +44,19 @@ extern "C" {
 #  error "Dispatching is currently only supported on x86 and x86_64."
 #endif
 
-#ifndef __GNUC__
-#  error "Dispatching requires __attribute__((__target__)) capability"
+#if defined(__GNUC__)
+#  include <immintrin.h> /* sse2 */
+#  include <emmintrin.h> /* avx2 */
+#  define XXH_TARGET_AVX512 __attribute__((__target__("avx512f")))
+#  define XXH_TARGET_AVX2 __attribute__((__target__("avx2")))
+#  define XXH_TARGET_SSE2 __attribute__((__target__("sse2")))
+#elif defined(_MSC_VER)
+#  include <intrin.h>
+#  define XXH_TARGET_AVX512
+#  define XXH_TARGET_AVX2
+#  define XXH_TARGET_SSE2
+#else
+#  error "Dispatching is currently not supported for your compiler."
 #endif
 
 #define XXH_DISPATCH_AVX2    /* enable dispatch towards AVX2 */
@@ -62,18 +73,8 @@ extern "C" {
 #endif
 #include <assert.h>
 
-#if defined(__GNUC__)
-#  include <immintrin.h> /* sse2 */
-#  include <emmintrin.h> /* avx2 */
-#elif defined(_MSC_VER)
-#  include <intrin.h>
-#endif
-
 #define XXH_INLINE_ALL
 #define XXH_X86DISPATCH
-#define XXH_TARGET_AVX512 __attribute__((__target__("avx512f")))
-#define XXH_TARGET_AVX2 __attribute__((__target__("avx2")))
-#define XXH_TARGET_SSE2 __attribute__((__target__("sse2")))
 #include "xxhash.h"
 
 /*


easyaspi314:
So MSVC will only use AVX2 instructions in blocks where it has AVX2 intrinsics even without `/arch:AVX2`, or do we have a chance of pollution in the non-AVX2 functions?

Cyan4973:
Thanks for the PR @apankrat !

While I don't doubt that this PR makes the source code compile fine on MSVC,
one of the remaining opened questions is
to ensure that each variant uses only the intended instruction set.

Since the code is employing explicit Intel intrinsics,
one direction is fine : 
I would expect the `avx2` path to employ `avx2` instructions.

But the problem is a bit more subtle.
What about auto-vectorization ?
Does the `sse2` path employ _only_ `sse2` instructions,
or is the compiler clever enough to refactor them into a more efficient `avx2` variant if it detects that these instructions are available ?
Same thing for the `scalar` path : is it completely free of vectorial instructions ?

This is, in essence, what the `__attribute__((__target__` statement is in charge of : 
ensure that each variant remains in its vector domain, no more, no less.

Now, it would seem that, if `MSVC` auto-vectorization abilities are rather weak,
it would kind of solve this problem.
But how to check that ?
And how to ensure that's still the case in the future ?

easyaspi314:
Well I compiled that code on my non-AVX laptop with MSVC x64 2019 19.27.29112.00 (1st gen Core i5-M430, Arrandale, supports up to SSE4.2) and the assembly clearly showed AVX2 and AVX512 instructions in the specialized blocks, but the program ran without any `SIGILL` issues, and the benchmark speeds and debug messages clearly took the SSE2 path. 

Also, side note: The benchmark calls the architecture "unknown" on MSVC x64...

Wait my bad, 
```c
#if defined(ARCH_X86)
```
I forgot to fix this line.

easyaspi314:
Additionally, running Intel's SDE in various modes shows it running AVX2 and AVX512 instructions as expected. 

I think it is safe to say that it is working. :+1:

apankrat:
@Cyan4973 - You are overthinking it... or not having enough trust in the sanity of MSVC devs :)

It's a common sense that a compiler should not use any instruction set extensions unless _explicitly_ allowed to do so. In MSVC case it needs [`/arch:xxx`](https://docs.microsoft.com/en-us/cpp/build/reference/arch-x64?view=vs-2019) to *auto* emit AVX instructions. The use of intrinsics doesn't imply this flag is set. Because if it were the case, I'm pretty sure it would've been documented, in very large friendly letters.

Also, for what it's worth, I've been using this sort of cpu-based code branching with MSVC since the early '00s, across multiple projects. No issues whatsoever, it just works as you'd expect.

Cyan4973 (approved):

Cyan4973:
As one can guess, without using MSVC regularly, there are multiple details that can only be anticipated with caution.

> In MSVC case it needs /arch:xxx to auto emit AVX instructions. The use of intrinsics doesn't imply this flag is set.

This last part of the sentence is key, and surprising to me.
I was expecting MSVC to _require_ `/arch:avx2` in order _accept_ introducing `avx2` instructions in the binary.
This is because in `gcc` and `clang`, trying to compile the `avx2` code path _without_ setting `-mavx2` results in code generation error. It appears the situation is different for MSVC. That's very interesting.

> > Does the sse2 path employ only sse2 instructions,
> > or is the compiler clever enough to refactor them into a more efficient avx2 variant if it detects that these instructions are available ?
>
> You are overthinking it... 

Well, and just FYI, the scenario described above is _exactly_ what happens with `clang` typically, which is very aggressive with auto-vectorization. So scalar code, and even `sse2` code, can be "upgraded" into `avx2` code if the compiler feels like it's more efficient. All it takes is to specify `-mavx2`, which is more or less the first idea that comes to mind to enable `avx2` instructions (and that's the problem solved by introducing `__attribute__((__target__`).

Anyway, given these elements, the present solution proposed for MSVC looks good and clean to me, and simpler than anticipated.
It's great to be able to extend the dispatcher's availability into MSVC territory!

easyaspi314:
Agreed. The main reason we are wary about it is that both Yann and I regularly use the GCC family and are unfamiliar with MSVC.  We are basically operating on [mutual intelligibility](https://en.m.wikipedia.org/wiki/Mutual_intelligibility).

This works fine for ANSI C code, but once you start drifting away from that (Unicode support, SIMD, command line options, etc) it gets very confusing.

apankrat:
> Well, and just FYI, the scenario described above is exactly what happens with clang typically, which is very aggressive with auto-vectorization. So scalar code, and even sse2 code, can be "upgraded" into avx2 code if the compiler feels like it's more efficient. All it takes is to specify -mavx2.

No, hold on.

`-mavx2` ultimately tells the compiler that we are OK with having any avx2 instructions in the code, so it's not a big surprise that it ends up emitting them whenever it deems appropriate. The whole situation is that we are basically enlarging the default instruction set for the whole code and then restricting it for a specific function with `__target__`.

This is the opposite of targeting a smaller instruction set by default, but including extended instructions in select functions via intrinsics. If we were to build the code with `/arch:avx2` with msvc, then we would've absolutely needed an equivalent of `__target__` to protect SSE2 versions from avx2 instructions.

This also means that the `__target__` trick is not needed even for the non-msvc builds. They can't be building with `-mavx2` and targeting non-avx2 CPUs, because that doesn't make much sense. So even if SSE2 code will be auto-vectorized with AVX2 instructions, no harm done - it will work on every CPU that we've targeted!

The one and only case when we would want to have pure SSE2 code in AVX2-targeted builds is when benchmarking the SSE2 implementation. To me, as a user of a library, this is an edge case.

Even more generally speaking the need for `__target__` can be entirely done away by extracting SIMD bits into separate .c files, one for each instruction set extension. Then build each file with highest applicable `-m...` level and that'll do it.

In any case and irrespective of any of this minutae, thanks a lot for your work on the xxHash! I've started using it as a part of the delta copying code in a file backup software a couple of years ago ([here](https://bvckup2.com/wip/25042018)) and it made a lot of difference. Fantastic code.

easyaspi314:
> The whole situation is that we are basically enlarging the default instruction set for the whole code and then restricting it for a specific function with `__target__`.

I believe you will find that we use `__target__` to enlarge a specific function, not the other way around. Restricting each function would be very painful, as we would have to mark *every single function* to be safe from these intrinsics, which also causes issues with inlining, as `__target__` implies `__noinline__` if the target is not already enabled in that function. (This is how the error is generated — the intrinsics are basically marked `__attribute__((__always_inline__, __target__("avx2")))` and the contradiction between the implicit `__noinline__` and `__always_inline__` causes the warning)


`__target__` literally says "Compile this function with this `-m` flag".


> Even more generally speaking the need for `__target__` can be entirely done away by extracting SIMD bits into separate .c files, one for each instruction set extension. Then build each file with highest applicable `-m`... level and that'll do it.

Yeah that was my original plan but Yann insisted that `xxhash.h` was contained in one file, with the only exception being the dispatch code. 

We also tried recompiling with a wrapper but that was considered too complicated.

apankrat:
Do I read you correctly that `gcc` will throw an error if I do *not* use `__target__ (x)` for the function that's has some `x` intrinsics?

apankrat:
Ah, yeah, I see what you mean. Tried it just now. It fails due to inlining restrictions, just as you explained. This is basically a gcc-ism  - functions that use extended instructions (via intrinsics) should be declared in a special way. I can see how it can be useful. Not a fan of how it's implemented though :)

easyaspi314:
> Ah, yeah, I see what you mean. Tried it just now. It fails due to inlining restrictions, just as you explained. This is basically a gcc-ism - functions that use extended instructions (via intrinsics) should be declared in a special way. I can see how it can be useful. Not a fan of how it's implemented though :)

Oh really?

Au contraire, I think MSVC's implementation is terrible.

It won't stop you from accidentally using unsupported intrinsics, such as found [here](https://github.com/RPCS3/rpcs3/issues/6479) where the SSE4.1 intrinsic `_mm_extract_epi32` was added and unnoticed because it is built with MSVC on Windows. 

GCC requires you to be explicit in what you allow and will stop you in a heartbeat if you screw up. 🙂

Cyan4973:
It's the old "more permissive" vs "more restrictive" debate all over again.

more permissive => easier to use, more unwanted behavior can slip in
more restrictive => more annoying to use, more control over what is allowed

Retrospectively, it's not too surprising to see these 2 compilers leaning in different directions given their respective audiences.

apankrat:
> Oh really?
> Au contraire, I think MSVC's implementation is terrible.

I was referring to the compiler errors this produces - https://gist.github.com/apankrat/a45b44dde2e3a792909f4c6ced34dfe4

> It won't stop you from accidentally using unsupported intrinsics, such as found [here](https://github.com/RPCS3/rpcs3/issues/6479) where the SSE4.1 intrinsic `_mm_extract_epi32` was added and unnoticed because it is built with MSVC on Windows.

This is not a very common mistake and that's exactly what the code reviews are for.

> GCC requires you to be explicit in what you allow and will stop you in a heartbeat if you screw up. slightly_smiling_face

Wrapping an intrinsic in a target-tagged function doesn't do much. The very same people who'd use `_mm_extract_epi32` on a general exec path would just call the wrapper function instead and be none the wiser.

This is inherently the same argument as in favor of using `if (NULL == x) ...` notation. And the solution should be similar - the compiler should emit a warning when a CPU instruction outside of the base `-m` set is being used in a function. That's it. If I want this to be an error, I'd use `-Werror`. If I want to suppress it, because I know the code is correct, I'd use a `#pragma`.

So neither gcc nor MSVC do it right. Yann's correct though, this is a theological debate for sure.

PR #460, opened 2020-10-12
[WIP] XXH3 multithreading

easyaspi314:
With this code, XXH3 will spawn a second thread when the length >= `XXH_THREADS_THRESHOLD` and `XXH_THREADS` is defined.

See details in the code comment.

TODO:
 - [X] Write basic implementation
 - [X] Test pthread implementation
 - [ ] Test Win32 implementation
 - [ ] Verify if all possible errors are handled
 - [ ] Find good default for `XXH_THREADS_THRESHOLD`. The current default is 8MB. I'd rather have an overestimate than an underestimate.
 - [ ] Add Makefile and CMake support
 - [ ] Add CI tests
 - [ ] Add implementation for WebAssembly?
 - [ ] Perhaps simplify `XXH3_threadData_t`
 - [ ] Reduce duplicate code

Please post your benchmarks if possible so we can find a decent default for `XXH_THREADS_THRESHOLD`.

Build instructions:
 - pthreads: `make CFLAGS="-O2 -g -DXXH_THREADS -pthread"`
 - Win32: `cmake . -DCMAKE_C_FLAGS="-O2 -DXXH_THREADS"`

### DIFF
--- a/cli/xsum_arch.h
+++ b/cli/xsum_arch.h
@@ -149,5 +149,10 @@
 #  define XSUM_ARCH "unknown"
 #endif
 
+#if defined(XXH_THREADS) && (defined(_WIN32) || (defined(_REENTRANT) && !defined(__EMSCRIPTEN__) && !defined(__wasm__)))
+#  define XSUM_THREADED " (multithreaded)"
+#else
+#  define XSUM_THREADED ""
+#endif
 
 #endif /* XSUM_ARCH_H */
--- a/xxhash.h
+++ b/xxhash.h
@@ -3766,15 +3766,17 @@ XXH3_accumulate(     xxh_u64* XXH_RESTRICT acc,
  *     3. When we are done, we check if the thread spawned successfully.
  *       - If it succeeded, we merge the accs and mark the second half as done.
  *       - If it failed, we just let the main thread process the second half.
- *         This followas the same path that it would if the length did not meet
+ *         This follows the same path that it would if the length did not meet
  *         the threshold.
  *    However, if we add more threads, we have to track *which* section we
  *    processed which is not as simple.
  *  - The "performance formula" is basically this:
- *        benefit = ((R * L) / T) - (C * (T - 1)); T >= 1.
- *    where T is the number of threads, C is the constant cost of setting up a
- *    thread, R is the constant time it takes to process a length of input, and
- *    L is the length of the input.
+ *        benefit = ((R * L) / T) - (C * (T - 1))
+ *    where:
+ *      - R is the constant time it takes to process a length of input
+ *      - L is the length of the input.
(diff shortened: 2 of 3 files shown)


easyaspi314:
LOL my `#warning` completely borks the syntax highlighting...I might change that.

easyaspi314:
@Cyan4973 how does this look?

easyaspi314:
Win32 is currently broken. Investigating now. Pthreads should still work, though.

Cyan4973:
> A few months ago, you mentioned how XXH3 is threadable

Yes, it is threadable.

While the algorithm's design makes it possible, one important remaining question is : when is that useful ?

To be fair, for a majority of "normal" `XXH3` usage scenarios, I don't see a point in making `XXH3` even faster . Reason is, in most cases, given a "powerful" enough core, such as any `aarch64` or `x64` cpu, `XXH3` is already too fast to begin with. No I/O can keep up with that. Even RAM can have difficulties keeping up.

That being said, in other cases, this could indeed prove useful. For example, less powerful cpus, but still multi-core ones, featuring multiple weak compute units. This is however not the "core" audience of `XXH3`, at least not yet.

Given that it's not for the "core" audience of `XXH3`, and that it introduces a _giant_ dependency (`pthread` or equivalent) complicating its integration, I'm not tempted into embedding this new MT capability into the library proper, `xxhash.h`.

That being said, it is decent ground for a dedicated project.
Either its own repository, or a project fully hosted into `xxhash` repository, as you prefer,
for example in `contrib/xxhash-mt`.
It could also become an opt-in variant of `cli`, though it's unclear if it would prove as useful : 
an MT library might be tasked to hash an object in raw memory, where there might be some speed left to extract, but the `cli` is most likely tasked to hash an object in I/O (storage or network), which is much slower, so unlikely to reach single-thread `XXH3` speed.

In term of implementation, this project could remain fairly close to what you just did : merely `inline` `xxhash.h` into the new unit, giving access to all its content, and now you can add your own functions, `#include`, warnings, and redirectors. Then, if you want the user-experience to be transparent, calling the same `XXH3_64bits()` or `XXH128()` symbols as the regular `xxhash` library, overload the symbol name with a `#define`, like `x86-dispatch.h` is doing.

For discussion.

easyaspi314:
Hmm...yes you do raise some good points.

I am mostly putting this here as a proof-of-concept.

From my tests, it appears that this benefits aarch64 mobile devices much more than x86, as aarch64's design is the slow clock speed and multicore design as you mentioned. However, XXH3's neon path is still faster than M.2 PCIe's 4000 MB/s. And most aarch64 devices that are slower than 4000 MB/s use eMMC. 

But yes, we could do a hook macro and have the MT code be separate.

PR #465, opened 2020-10-18
Various dispatcher improvements

### LINKED ISSUE #464: Can't compile on CentOS 7
I'm trying to do best-efforts compilation of 0.8.0 including the X86 dispatch code for use in my database synchronization software.

CentOS 7 has GCC 4.8.5 which doesn't support AVX-512, that is, it doesn't support the `-mavx512f` and `-mavx512vl` flags and gives errors like `error: attribute(target("avx512f")) is unknown`, `error: '__m512i' was not declared in this scope` etc. without them.

As a result, I can't compile the dispatch code, because it will turn on compilation of all vectorized variants:

`#if (XXH_VECTOR == XXH_AVX512) || defined(XXH_X86DISPATCH)`

We could mix in more of the `__AVX512F__` define tests, but I suspect that's going to require compiler-specific stuff to be spread throughout the file. Instead I propose that we refactor this to a set of vectorization support flags (eg. `XXH_VECTOR_SUPPORT_AVX512`), which can be left unset when the compiler flag tests fail.

(I don't use CentOS and I don't really get why people do. But they do, and they sent me bug reports and patches for my open-source software to support it.)

easyaspi314:
- Test the compiler for AVX2/AVX512 support instead of unconditionally defining `XXH_DISPATCH_*` (fixes #464)
   - Can also be enabled/disabled on the command line
 - Use a macro template to reduce code repetition
 - Don't dispatch the scalar path when we don't need it. It can be rather wasteful, especially on 32-bit.
   - Specifically, don't dispatch when SSE2 is globally enabled on the compiler or when it is guaranteed on the platform.
 - Add some Doxygen documentation for xxh_x86dispatch.c.

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -20,7 +20,7 @@ QUIET                  = YES
 WARN_IF_UNDOCUMENTED   = NO
 
 # TODO: Add the other files. It is just xxhash.h for now.
-FILE_PATTERNS          = xxhash.h
+FILE_PATTERNS          = xxhash.h xxh_x86dispatch.c
 # Note: xxHash's source files are technically ASCII only.
 INPUT_ENCODING         = UTF-8
 TAB_SIZE               = 4
--- a/xxh_x86dispatch.c
+++ b/xxh_x86dispatch.c
@@ -220,9 +220,9 @@ extern "C" {
  * Note: Comments are written in the inline assembly itself.
  */
 #ifdef __clang__
-#  define I_ATT(intel, att) att "\n\t"
+#  define XXH_I_ATT(intel, att) att "\n\t"
 #else
-#  define I_ATT(intel, att) "{" att "|" intel "}\n\t"
+#  define XXH_I_ATT(intel, att) "{" att "|" intel "}\n\t"
 #endif
 
 /*!
@@ -244,14 +244,14 @@ static void XXH_cpuid(xxh_u32 eax, xxh_u32 ecx, xxh_u32* abcd)
         "#\n\t"
         "# On 32-bit x86 with PIC enabled, we are not allowed to overwrite\n\t"
         "# EBX, so we use EDI instead.\n\t"
-        I_ATT("mov     edi, ebx",   "movl    %%ebx, %%edi")
-        I_ATT("cpuid",              "cpuid"               )
-        I_ATT("xchg    edi, ebx",   "xchgl   %%ebx, %%edi")
+        XXH_I_ATT("mov     edi, ebx",   "movl    %%ebx, %%edi")
+        XXH_I_ATT("cpuid",              "cpuid"               )
+        XXH_I_ATT("xchg    edi, ebx",   "xchgl   %%ebx, %%edi")
         : "=D" (ebx),
 # else
     __asm__(
         "# Call CPUID\n\t"
-        I_ATT("cpuid",              "cpuid")
+        XXH_I_ATT("cpuid",              "cpuid")
         : "=b" (ebx),
 # endif
               "+a" (eax), "+c" (ecx), "=d" (edx));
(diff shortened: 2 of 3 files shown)


Doxyfile
```diff
@@ -20,7 +20,7 @@ QUIET                  = YES
 WARN_IF_UNDOCUMENTED   = NO
 
 # TODO: Add the other files. It is just xxhash.h for now.
-FILE_PATTERNS          = xxhash.h
+FILE_PATTERNS          = xxhash.h xxh_x86dispatch.c
```
Cyan4973, line 23: minor : more a question :
why is it `xxh_x86dispatch.c`, rather than `xxh_x86dispatch.h` ?

xxh_x86dispatch.c
```diff
@@ -572,25 +560,40 @@ typedef struct {
...
+    /* AVX512 */ { XXHL64_default_avx512, XXHL64_seed_avx512, XXHL64_secret_avx512, XXH3_update_avx512 }
+#else
+    /* AVX512 */ { NULL, NULL, NULL, NULL }
 #endif
 };
+/*!
+ * @internal
+ * @brief The selected dispatch table for @ref XXH3_64bits().
+ */
+static XXH_dispatchFunctions_s XXH_g_dispatch = { NULL, NULL, NULL, NULL };
```
Cyan4973, line 595: minor : more a comment : 
having a `XXH_*` prefix in front of every symbol and name
is important to mitigate namespace pollution,
particularly for source code which will be included into some 3rd-party source code, like a `*.h`.

But here, `xxh_x86dispatch.c` is not designed to be included,
so this additional namespace precaution feels superfluous.

not a big issue either : it's not detrimental.

xxh_x86dispatch.c
```diff
@@ -293,269 +443,107 @@ static int XXH_featureTest(void)
...
+/*!
+ * @internal
+ * @brief Defines the various dispatch functions.
+ *
+ * TODO: Consolidate?
+ *
+ * @param suffix The suffix for the functions, e.g. sse2 or scalar
+ * @param target XXH_TARGET_* or empty.
+ */
+#define XXH_DEFINE_DISPATCH_FUNCS(suffix, target)                             \
```
Cyan4973, line 455: I'm not a big fan of long macros like that,
but I can't deny it reduces code repetition.
Anyway, I don't have any better proposal at hand.

Cyan4973 (approved):

Cyan4973:
It's a pretty good improvement.
It gives better, more accurate, control over variant generation,
and even improves documentation.

Cyan4973:
Interesting warning when compiling `xxhsum DISPATCH=1` after applying this PR : 

```
xxh_x86dispatch.c: In function ‘XXH_featureTest’:
xxh_x86dispatch.c:396:33: warning: ‘max_leaves’ is used uninitialized in this function [-Wuninitialized]
  396 |     if (XXH_unlikely(max_leaves < 7))
      |                                 ^
xxh_x86dispatch.c: In function ‘XXH_setDispatch’:
xxh_x86dispatch.c:400:18: warning: ‘abcd[2]’ may be used uninitialized in this function [-Wmaybe-uninitialized]
  400 |     if ((abcd[2] & XXH_OSXSAVE_CPUID_MASK) != XXH_OSXSAVE_CPUID_MASK)
      |         ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
xxh_x86dispatch.c:324:13: note: ‘abcd[2]’ was declared here
  324 |     xxh_u32 abcd[4];
```

It also seems to break `AVX2` runtime detection,
as the speed of `XXH3` after this patch is more in line with `SSE2` than `AVX2`.

xxh_x86dispatch.c
```diff
@@ -163,22 +303,31 @@ static xxh_u64 XXH_xgetbv(void)
...
 static int XXH_featureTest(void)
 {
     xxh_u32 abcd[4];
     xxh_u32 max_leaves;
     int best = XXH_SCALAR;
-#if defined(XXH_DISPATCH_AVX2) || defined(XXH_DISPATCH_AVX512)
+#if XXH_DISPATCH_AVX2 || XXH_DISPATCH_AVX512
     xxh_u64 xgetbv_val;
 #endif
+#if XXH_DISPATCH_SCALAR
```
Cyan4973, line 330: Why is this code inside an `#if XXH_DISPATCH_SCALAR` conditional ?
`max_leaves` initialization happens within this segment.

PR #466, opened 2020-10-19
Expose XXH_alignedMalloc, XXH3_state_t new/delete

easyaspi314:
`XXH_alignedMalloc` and `XXH_alignedFree` are now exposed because they may be helpful. They autoclamp their arguments instead of asserting.

Additionally, when compiled in C++, `XXH3_state_t` now has overloads for `operator new` and `operator delete` so it can safely be used in containers without any alignment issues. They simply call `XXH_alignedMalloc` and `XXH_alignedFree`.

These are static member functions, so they don't change the layout of `XXH3_state_t`, they are just inlines.

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -838,6 +838,59 @@ typedef struct { unsigned char digest[sizeof(XXH128_hash_t)]; } XXH128_canonical
 XXH_PUBLIC_API void XXH128_canonicalFromHash(XXH128_canonical_t* dst, XXH128_hash_t hash);
 XXH_PUBLIC_API XXH128_hash_t XXH128_hashFromCanonical(const XXH128_canonical_t* src);
 
+/*!
+ * @ingroup public
+ * @brief Mallocs a pointer that is aligned to a certain power of 2.
+ *
+ * Acts like malloc() but takes an @p align parameter.
+ *
+ * malloc() typically guarantees 16 byte alignment on 64-bit systems and 8 byte
+ * alignment on 32-bit. This isn't enough for the 32 byte aligned loads in AVX2
+ * or on 32-bit, the 16 byte aligned loads in SSE2 and NEON.
+ *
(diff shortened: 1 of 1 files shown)


Cyan4973:
Is the need to have `XXH_alignedMalloc()` and `XXH_alignedFree()` public, as opposed to private,
related to the introduction of C++ overload operators ?

easyaspi314:
Well the main problem is that by the standard, `operator new` is a raw malloc equivalent. 

Therefore, even if you only use it to allocate one class, it must be able to take any allocation amount. 

If that were not the case, I'd just do
```c++
    static void* operator new(size_t /* unused */) {
        return XXH3_createState();
    }
```

I could technically do
```c++
struct XXH3_state_s {
    // ...
    static void* operator new(size_t);
};

// implementation
void* XXH3_state_s::operator new(size_t size) {
    return XXH_alignedMalloc(size, 64);
}
```
But that would cause a linker error if xxhash.c is (logically) compiled as a C file, as that is obviously going to be a C++ symbol.

I could copypaste it but that is a blatant DRY violation.
I could put alignedMalloc *above* the struct but remain static, but that would move multiple definitions into the header section of the file. 

What do you think?

Cyan4973:
Sorry, I haven't looked at this topic in a long time.

If I do remember correctly, the ultimate goal of this PR is to provide a `new` operator which aligns properly for `XXH3_state_t`.

I'm puzzled because the goal of `xxhash.h` is to be a `C90` library. This functionality introduces `C++` specificities right into it.

I would prefer a design in which the underlying `C90` code is wrapped into a `C++` overlay.
It could be called `xxh++.h` for example. It would allows concentrating C++-specific logic into this unit.
What matters is that it should be introduced into a unit separated from the `C90` one.

Cyan4973:
closing, 
the PR has remained stale for a while,
and I believe we should first discuss the objectives, and agree about a way to deliver them.

PR #468, opened 2020-10-26
Fix issue with XXHASH_BUILD_XXHSUM cmake setting

devpow112:
Using `set(... CACHE ...)` causes `XXHASH_BUILD_XXHSUM` to be stuck `ON` even when doing a `set(XXHASH_BUILD_XXHSUM OFF)` for some cmake version/generators. Switching to `option(...)` fixes this issue and allows you to properly turn off the building of xxhsum in all situations.

Currently using `set(XXHASH_BUILD_XXHSUM OFF CACHE BOOL "" FORCE)` before including xxhash into another cmake project gets around the issue but is non-obvious.


Cyan4973 (approved):

devpow112:
Thanks @Cyan4973 for the quick review 👍

PR #477, opened 2020-11-24
Support no-op use of xxh_x86dispatch files on non-X86 platforms

willbryant:
Currently the core xxhash.{h,c} files are globally usable and take care of all platform-specific logic.

xxh_x86dispatch.c on the other hand can currently only be compiled on x86 so the user's build system needs to detect the platform and include or exclude it from compilation, and the code that will call the dispatch functions needs to also detect the platform and determine whether to enable/use the dispatch code or not.

So right now the x86 logic needs to be in 3 places - xxh_x86dispatch.c, the build system, and the code including xxh_x86dispatch.h. I don't think this is beneficial, even if currently dispatching is only actually useful on X86.

In this PR I have changed the logic so that compilation of xxh_x86dispatch.c on non-X86 is legal but a no-op, so the build system doesn't need to know, and changed xxh_x86dispatch.h to automatically set XXH_DISPATCH_DISABLE_REPLACE so the including code doesn't need to know.

This isn't the only way to do it; the other obvious option would be to not mess around with XXH_DISPATCH_DISABLE_REPLACE and instead define passthrough *_dispatch functions on non-X86 platforms. That approach would have the advantage of supporting users who prefer to specifically invoke the functions with *_dispatch names, but would have the cost of one unnecessary indirection (function call) for each call on non-X86 platforms. I can't think of the use case for preferring calling _dispatch names, so I have instead gone the route I describe above.

After this commit, IMHO it would also make sense to rename these two files to something more generic, but that's a separate decision.

### DIFF
--- a/xxh_x86dispatch.c
+++ b/xxh_x86dispatch.c
@@ -52,9 +52,7 @@
 extern "C" {
 #endif
 
-#if !(defined(__x86_64__) || defined(__i386__) || defined(_M_IX86) || defined(_M_X64))
-#  error "Dispatching is currently only supported on x86 and x86_64."
-#endif
+#if (defined(__x86_64__) || defined(__i386__) || defined(_M_IX86) || defined(_M_X64))
 
 /*!
  * @def XXH_X86DISPATCH_ALLOW_AVX
@@ -763,6 +761,8 @@ XXH3_128bits_update_dispatch(XXH3_state_t* state, const void* input, size_t len)
     if (XXH_g_dispatch128.update == NULL) XXH_setDispatch();
     return XXH_g_dispatch128.update(state, (const xxh_u8*)input, len);
 }
+
+#endif
 
 #if defined (__cplusplus)
 }
--- a/xxh_x86dispatch.c
+++ b/xxh_x86dispatch.c
@@ -52,7 +52,12 @@
 extern "C" {
 #endif
 
-#if (defined(__x86_64__) || defined(__i386__) || defined(_M_IX86) || defined(_M_X64))
+#if !(defined(__x86_64__) || defined(__i386__) || defined(_M_IX86) || defined(_M_X64))
+
+/* some build systems complain about empty units, so define this to save trouble */
+extern int __xxh_x86dispatch_non_empty_unit;
+
+#else
 
 /*!
  * @def XXH_X86DISPATCH_ALLOW_AVX
--- a/xxh_x86dispatch.h
+++ b/xxh_x86dispatch.h
@@ -41,15 +41,21 @@
 extern "C" {
 #endif
 
-XXH_PUBLIC_API XXH64_hash_t  XXH3_64bits_dispatch(const void* input, size_t len);
-XXH_PUBLIC_API XXH64_hash_t  XXH3_64bits_withSeed_dispatch(const void* input, size_t len, XXH64_hash_t seed);
-XXH_PUBLIC_API XXH64_hash_t  XXH3_64bits_withSecret_dispatch(const void* input, size_t len, const void* secret, size_t secretLen);
-XXH_PUBLIC_API XXH_errorcode XXH3_64bits_update_dispatch(XXH3_state_t* state, const void* input, size_t len);
+#if (defined(__x86_64__) || defined(__i386__) || defined(_M_IX86) || defined(_M_X64))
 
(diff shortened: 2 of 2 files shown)


Cyan4973:
OK, sounds like a good idea.

My only comment here is that units that are completely empty tend to generate warnings in some build systems,
and restrictive environments which do not tolerate any warning may not like that.

I typically get around such issue by defining a completely useless variable in the unit, with a name complex enough that it should not reasonably produce any symbol collision at link time.

willbryant:
Maybe I should always define the _dispatch functions *and* set XXH_DISPATCH_DISABLE_REPLACE on non-X86 platforms? Then we get the efficiency of no extra function indirection, the availability of those names, and non-empty units :).

Thoughts?

Cyan4973:
I'm not sure to fully understand this last proposal, especially its visible side effects. It feels more complex than merely ensuring that non-x86 can compile `xxh_x86dispatch.c` without an "empty-unit" warning.

alli03106 (approved):

willbryant:
Sorry, I didn't explain that very well.

I've pushed the other approach in #480 so you can see what I mean.

And I've added a commit to this PR for the dummy variable, is that what you meant?

alli03106 (approved):

Cyan4973:
Sorry, I completely forgot about this topic.

I think it's a rather good idea to simplify build logic for all platforms.
Let me have a look at the specifics, and notably get back at your 2 proposals.

xxh_x86dispatch.c
```diff
@@ -53,8 +53,11 @@ extern "C" {
 #endif
 
 #if !(defined(__x86_64__) || defined(__i386__) || defined(_M_IX86) || defined(_M_X64))
-#  error "Dispatching is currently only supported on x86 and x86_64."
-#endif
+
+/* some build systems complain about empty units, so define this to save trouble */
+extern int __xxh_x86dispatch_non_empty_unit;
```
Cyan4973, line 58: 👍

t-mat: It'd be nice to avoid `__` (double-underscore) prefix to follow the C standard clause ["Reserved identifiers"](https://stackoverflow.com/a/228797).

Cyan4973: triple-underscore ?

t-mat: Sorry for confusion.

1. Since xxHash always follow C standard, it'd be nice to avoid double-underscore prefix (e.g. `__xxh...`) to follow the C standard clause ["Reserved identifiers"](https://stackoverflow.com/a/228797).
   We also should avoid triple or further underscore prefix because it also contains double underscore prefix.  ( `strncmp("__", "___", 2) == 0` )

2. xxHash also cares basic C++ standard (not explicitly, but LZ4 has `ctocppcheck`).  xxHash has been avoiding underscore + upper case prefix (e.g. `_Xxh...`).

3. It's a not strict rule.  But conventionally and historically, all external xxHash macro, variable, function and enum names are started with `XXH`.  Some type name starts with `xxh`.

I think intention of this symbol is able to achieve without double underscore prefix.

willbryant: Yup, it's totally arbitrary, I can change it.

I'm just a bit confused - I thought the comment on #480 meant that approach was going to be pursued and this PR could be closed?

t-mat: @Cyan4973, could you suggest which PR (#477 or #480) should be pursued, abandoned or... mixed?

Cyan4973: Sorry, I will need a bit more time for this topic.

The short answer is : none will be pursued as is. 
I will probably borrow a bit of both, but mostly end up with a fairly different PR to support the requested feature.

willbryant: Keen to progress this. Anything I can do to help?

Cyan4973: Sorry, this is a difficult topic, that requires some dedicated focus.
And I'm sorry to report that I'm completely overwhelmed these days, with too many hot topics stealing my attention.

If you want to make progresses on this topic, you are absolutely welcome to do so.
It's one of the primary benefits of open source, that you can fork the project and make progresses on your own without further dependency from upstream. And even advocate for it as a better alternative for some use cases.
Hopefully, there will be a period in the future where it will be possible to focus and join efforts.

PR #480, opened 2020-11-29
Alternate approach to support no-op use of xxh_x86dispatch files on non-X86 platforms

willbryant:
Per #477, here is the other reasonable way to do it, for comparison.

Please consider merging either that one or this one :).

### DIFF
--- a/xxh_x86dispatch.c
+++ b/xxh_x86dispatch.c
@@ -53,8 +53,21 @@ extern "C" {
 #endif
 
 #if !(defined(__x86_64__) || defined(__i386__) || defined(_M_IX86) || defined(_M_X64))
-#  error "Dispatching is currently only supported on x86 and x86_64."
-#endif
+
+/* passthrough functions, not normally called because XXH_DISPATCH_DISABLE_REPLACE is set, but available
+   for the benefit of codebases that explicitly invoke the _dispatch functions by those names */
+
+XXH64_hash_t  XXH3_64bits_dispatch(const void* input, size_t len)                                                  { return XXH3_64bits(input, len); }
+XXH64_hash_t  XXH3_64bits_withSeed_dispatch(const void* input, size_t len, XXH64_hash_t seed)                      { return XXH3_64bits_withSeed(input, len, seed); }
+XXH64_hash_t  XXH3_64bits_withSecret_dispatch(const void* input, size_t len, const void* secret, size_t secretLen) { return XXH3_64bits_withSecret(input, len, secret, secretLen); }
+XXH_errorcode XXH3_64bits_update_dispatch(XXH3_state_t* state, const void* input, size_t len)                      { return XXH3_64bits_update(state, input, len); }
+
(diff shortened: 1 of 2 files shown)


Cyan4973:
OK, it took a while, but I think I'm getting a sense of the respective trade-offs at stake here.

Here is a list of properties that I believe are desirable:

- It should be possible to compile `xxhsum` CLI application on any platform, including non `x64` ones, and transparently get the benefits of automatic vector dispatch on `x64` platforms without any specific action from the user nor the package maintainer.
	- Currently, the variant of `xxhsum` with vector dispatch is not enabled by default. It's a specific target that must be selected by the package maintainer, either `make dispatch`, or `make xxhsum DISPATCH=1`. 
		- This variant only works on `x64` target systems, hence why it's not enabled by default.
- Likewise, it should be possible to compile `libxxhash` on any platform, and transparently include dispatching capabilities in the library when compiled for `x64` platforms.
	- However, unlike the CLI application, the _default symbols_, like `XXH3_64bits()`, ___must___ continue targeting the native compile-time implementation.
		- The presence of a cpu detection instruction in the middle of hash calculation is mostly unwelcome.
		- The trade-off, requiring to save the outcome of this instruction in a global variable, that is then retrieved and checked on each successive hash computation, still is a form of load.
			- It setups a global variable, which is not always acceptable from a library
			- It may also result in a limited form of race condition (probably not dangerous but still detectable) if multiple hashes are started before a single one has time to write the result into the global variable.
			- The good news is that this load has been confined to "large" inputs only (>240 bytes), so small inputs are unaffected (they don't need vector instructions anyway).
		- This is especially important in combination with inlining.
			- Note that dispatch functions are designed to _not_ be inlined.
	- Therefore, invoking a dispatch function should be an _explicit_ operation.
		- This also means that the automatic translation of _normal_ symbols (like `XXH3_64bits()`) to dispatch ones (like` XXH3_64bits_dispatch()`) shouldn't be the default. This is at odd with current implementation.
	- Likewise, for the library to present a uniformed interface whatever the target instruction set, `x64`, `arm` or else, the symbols related to the `dispatch` variant should remain available
		- Even if they technically don't dispatch anything, and just forward to the only variant compiled for the target instruction set.
		- This makes it possible for an application to always invoke the `dispatch` variant if it wishes, without having to worry if it's present or not in the linked library, since it would always be present (on condition of minimum library version number).

Therefore, it favors this second proposed design.

I will then update `xxh_x86dispatch.h` to disable symbol replacement by default (it will now be an explicit request, controlled by an include macro).

The next step is to make `xxh_x86dispatch` part of `libxxhash` by default, since it's essentially harmless (just a redirection) on non-x64 platforms. I might as well drop the `x86` moniker from the name, since the header would become compatible with any platform. It might actually be possible to extend dynamic detection and usage of local vector capabilities to other systems, if it makes sense.

There might be additional complex `#include` scenarios to test with it, such as what happens in situations involving multiple `#include`, which could inadvertently happen transitively.

willbryant:
OK, so we're going ahead with this PR (#480) instead of #477?

I don't remember why I used wrapper functions for the non-X86 case, instead of #defines. That seems unnecessary overhead for every other platform. Thoughts?

PR #498, opened 2021-02-21
solaris restrict keyword fix

traversc:
Does not compile on Solaris due to the restrict keyword. 

I believe it is the same issue as: https://bugzilla.mozilla.org/show_bug.cgi?id=1158444

A simple fix is to disable restrict on Solaris.

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -2679,8 +2679,9 @@ XXH_PUBLIC_API XXH64_hash_t XXH64_hashFromCanonical(const XXH64_canonical_t* src
 
 /* ===   Compiler specifics   === */
 
-/* ===   Compiler specifics   === */
-#if defined(sun) || defined(__sun)
+#if ((defined(sun) || defined(__sun)) && __cplusplus) /* Solaris includes __STDC_VERSION__ with C++. Tested with GCC 5.5 */
+#  define XXH_RESTRICT /* disable */
+#elif defined (__STDC_VERSION__) && __STDC_VERSION__ >= 199901L   /* >= C99 */
 #  define XXH_RESTRICT /* disable */
 #elif defined (__STDC_VERSION__) && __STDC_VERSION__ >= 199901L   /* >= C99 */
 #  define XXH_RESTRICT   restrict
--- a/xxhash.h
+++ b/xxhash.h
@@ -2679,7 +2679,10 @@ XXH_PUBLIC_API XXH64_hash_t XXH64_hashFromCanonical(const XXH64_canonical_t* src
 
 /* ===   Compiler specifics   === */
 
-#if defined (__STDC_VERSION__) && __STDC_VERSION__ >= 199901L   /* >= C99 */
+/* ===   Compiler specifics   === */
+#if defined(sun) || defined(__sun)
+#  define XXH_RESTRICT /* disable */
+#elif defined (__STDC_VERSION__) && __STDC_VERSION__ >= 199901L   /* >= C99 */
 #  define XXH_RESTRICT   restrict
 #else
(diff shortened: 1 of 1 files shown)


Cyan4973:
In the bug report in reference,
the issue was that `restrict` would cause trouble to Solaris' ***C++*** compiler.

Is that your case ? Are you using a `C` or a `C++` compiler ?

This matters to design a proper fix to the situation.

traversc:
Yes, it was C++.

Cyan4973:
Then the issue becomes that

`define XXH_RESTRICT   restrict`
is protected by this test
`defined (__STDC_VERSION__) && __STDC_VERSION__ >= 199901L`

which basically means that the compiler announces `C99` compliance,
meaning it's supposed to support the whole compulsory part of the spec
which `restrict` belongs to.

Why would a `C++` compiler claim being _also_ compliant with `C99`
which is certainly not correct by default since `restrict` is not part of the `C++` standard ?

Likely an oversight.

Anyway, the interesting question becomes : 
is this compiler _also_ able to compile in `C99` mode (hence _not_ `C++`) ?
(_note_ : `C99` != `C90`)
If yes, does it compile `xxhash` correctly in this mode (with `restrict`) or does it fail too ?

traversc:
I don't have full access to a Solaris machine, but it's something that's part of an automated check for R packages. 

So you can reproduce the issue using: https://github.com/traversc/xxhash_test

And in R:

```
library(rhub)
setwd("xxhash_test")
check_on_solaris()
```

Switching compilation back to C in that repo compiles fine. It's only with C++.

Cyan4973:
Good enough,
so I guess the correct test should check both Solaris compiler (`defined(sun) || defined(__sun)`),
*and* `__cplusplus`,
otherwise `restrict` will also be removed when compiling in `C` mode
which would be a pity, as it's a fairly important keyword to enable optimizations
(assuming the compiler is able to take advantage of it).

Cyan4973 (approved):

Cyan4973:
Thanks @traversc !

xxhash.h
```diff
@@ -2679,7 +2679,11 @@ XXH_PUBLIC_API XXH64_hash_t XXH64_hashFromCanonical(const XXH64_canonical_t* src
 
 /* ===   Compiler specifics   === */
 
-#if defined (__STDC_VERSION__) && __STDC_VERSION__ >= 199901L   /* >= C99 */
+#if ((defined(sun) || defined(__sun)) && __cplusplus) /* Solaris includes __STDC_VERSION__ with C++. Tested with GCC 5.5 */
+#  define XXH_RESTRICT /* disable */
+#elif defined (__STDC_VERSION__) && __STDC_VERSION__ >= 199901L   /* >= C99 */
+#  define XXH_RESTRICT /* disable */
+#elif defined (__STDC_VERSION__) && __STDC_VERSION__ >= 199901L   /* >= C99 */
```
nealrichardson, line 2686: Is this right? L2684 and L2686 look like they're testing the same condition.

Cyan4973: Good point @nealrichardson !
This is actually incorrect.
The first test shadows the second one
resulting in `XXH_RESTRICT` being always disabled.

nealrichardson: IIUC you don't need to revert the whole thing, you just want to prune L2685-L2686 right?

Cyan4973: Yes. The same PR, but with these 2 lines removed.

nealrichardson: 👍 I can do that; in a strange coincidence I happened upon this exact issue this weekend myself.

traversc: My bad! Sorry that's embarassing!

PR #502, opened 2021-02-22
Fix the solaris restrict patch

nealrichardson:
Followup to https://github.com/Cyan4973/xxHash/pull/498 cc @traversc


Cyan4973 (approved):

Cyan4973:
Thanks for detecting the issue and the fix @nealrichardson !

pitrou:
For the record, here's the upstream (gcc) issue: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=57025

PR #506, opened 2021-03-08
cmake script builds and installs a different version than the makefile

gabm:
From a packaging point of view, it would be great to achieve parity between the binaries and install created by the Makefile and cmake. 3 things stand in the way

1. the SONAME is different (I consider this a copy-paste bug in the current script)
2. some additional CFLAGS get added to the compiler flags. The Makefile does not mess with that. Let's do the same! *on a minor note*: the makefile uses both: `CFLAGS` and `CPPFLAGS` which is dubious, given that we only build a C lib
3. the pkg-config file does not get created and installed
4. symlinks for `{xxh128, xxh64, xxh32} -> xxhsum` are being created by the Makefile, cmake only installs `xxhsum`

This PR solves the first 3 problems. Not sure about 4?

### DIFF
--- a/cmake_unofficial/CMakeLists.txt
+++ b/cmake_unofficial/CMakeLists.txt
@@ -78,7 +78,7 @@ if (BUILD_SHARED_LIBS)
   target_compile_definitions(xxhash PUBLIC XXH_EXPORT)
 endif ()
 set_target_properties(xxhash PROPERTIES
-  SOVERSION "${XXHASH_VERSION_STRING}"
+  SOVERSION "${XXHASH_LIB_SOVERSION}"
   VERSION "${XXHASH_VERSION_STRING}")
 
 if(XXHASH_BUILD_XXHSUM)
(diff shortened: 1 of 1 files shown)


Cyan4973:
> the makefile uses both: CFLAGS and CPPFLAGS which is dubious, given that we only build a C lib

`CPPFLAGS` stands for "C PreProcessor Flags".
So yes, it's needed.

> This PR solves the first 3 problems. Not sure about 4?

Adding symlinks is a secondary topic, 
it is not required for this PR.

Cyan4973:
minor :
it seems a formatting test detected a few trailing white spaces in updated `CMakeLists.txt`

gabm:
> CPPFLAGS stands for "C PreProcessor Flags".

this is embarrassing. so many years of programming and I assumed it meant "C++"...

gabm:
hopefully I catched all whitespaces

gabm:
looks good @Cyan4973 👍

Cyan4973 (approved):

gabm:
oops sorry, that was unintended

gabm:
@Cyan4973 thanks for approving the changes. Would you also merge? or is there another policy?

Cyan4973:
no, just waiting a bit, in case external comments show up.
Should be fine now.

PR #510, opened 2021-03-12
Update xxhash.h

goodengineer:


xxhash.h
```diff
@@ -4382,7 +4383,7 @@ static XXH64_hash_t
 XXH3_mergeAccs(const xxh_u64* XXH_RESTRICT acc, const xxh_u8* XXH_RESTRICT secret, xxh_u64 start)
 {
     xxh_u64 result64 = start;
-    size_t i = 0;
+    uint8_t i = 0;
```
Cyan4973, line 4386: `uint8_t`, aka `unsigned char`, is pretty much the worst type for performance.
Plus it requires to drag in `stdint.h` dependency.
Prefer `int`.

xxhash.h
```diff
@@ -284,6 +284,7 @@ XXH_PUBLIC_API unsigned XXH_versionNumber (void);
 *  Definitions
 ******************************/
 #include <stddef.h>   /* size_t */
+#include <stdint.h>
```
Cyan4973, line 287: No dependency should be added lightly.
This one just broke `C90` compliance.

Cyan4973 (changes requested):
remove `uint8_t` and its associated dependency.

PR #513, opened 2021-03-12
Update timefn.c

goodengineer:


tests/bench/timefn.c
```diff
@@ -28,52 +27,51 @@ UTIL_time_t UTIL_getTime(void) { UTIL_time_t x; QueryPerformanceCounter(&x); ret
 PTime UTIL_getSpanTimeMicro(UTIL_time_t clockStart, UTIL_time_t clockEnd)
 {
     static LARGE_INTEGER ticksPerSecond;
-    static int init = 0;
+    static bool init = false;
```
Cyan4973, line 30: `bool` requires a preamble to work.
`#include <stdbool.h>` is the "standard" way, but only works on `C99` compilers.
Some C++ compiler like `clang` might make it look like it could work without it, but that's not "standard compliant", therefore it will fail on other compilers.

Considering the very low value this type brings, I don't think it's worth the associated hassle.

Issue #54, opened 2016-02-15
Generating collisions


leventov:
For e. g. testing purposes I need to generate different sequences, whose XXH64 hash codes collide.

I've found [this thread](http://crypto.stackexchange.com/q/6408/8646) and applied (rather blindly) the same algorithm to XXH64, so I got

```
0xBA79078168D4BAF * 14029467366897019727 = 1 (mod 2^64)
-7046029288634856825 * 2^31 = 0xC2F5E54380000000 (mod 2^64)
0x9C90005B80000000 * -4417276706812531889 = -0xC2F5E54380000000 (mod 2^64)
```

 1. Add `0xBA79078168D4BAF` to the 64-bit value at position `P` (in the little-endian order)
 2. Add `0x9C90005B80000000` to the value at `P + 32`.

But it doesn't work (the hash codes are different).

Cyan4973:
Hi Roman

The thread you found explain how to create a collision with xxh32.
For xxh64, similar reasoning may apply, but it nonetheless needs to be studied thoroughly. There could be a single minor difference, but it will make the solution different.

leventov:
Yes I see that it is for xxh32, so I applied the same reasoning with xxh64's constants (P1, P2) and multiplied offsets by 2 (16 -> 32).

Cyan4973:
I see you adapted the constants to xxh64.
But since it doesn't work, it could be either : 
- a calculation error 
- another difference, which does not allow to blindly apply the same method.

I don't see where the difference could be, since both xxh32 & xxh64 internal loops are very similar,

so maybe a calculation error ?

Cyan4973:
Note that I'm currently finding the same constants as you, so constants error looks less likely.

Next in the list are the test conditions.
I guess you already checked them, but just in case : 
- Constant 1 at position P, constant 2 at position P+32
  + Ensure that P is a multiple of 8
  + Each constant must be added using little convention
- Segments must be full : 
  + Ensure that there are _at least_ 32 more bytes of input after position P+32

leventov:
No, I did not obey additional conditions you listed, but now with them, it works. Thank you!

BTW it would be useful to document these collision generation procedures in the repo.

chmike:
I also need short byte sequences that collide with XXH64 for testing purpose. Could we have two or three samples ? My use case of to use xxHash for a map (index). Unfortunately, @leventov did not provide his results. The algorithm he used is unclear.

leventov:
[...]

PR #518, opened 2021-03-25
WIP: add a 128bit function XXH64X2 based on XXH64

gzm55:
`XXH64X2` is modified directly from `XXH64`, and the low 64bits are the same as `XXH64`. The purpose is that when porting to an environment without SIMD vectoring, `XXH64` is is still faster than `XXH3` and `XXH128`. The new function can cover the 128b hash usage on these environments.

The new hash has passed all the smhasher tests and the collision ratio tests as 128b hash. The performance is about 85% of the `XXH64`. While porting to the java, the performance is 95% of the `XXH64`.

### DIFF
--- a/xxh64x2.h
+++ b/xxh64x2.h
@@ -25,137 +25,32 @@ XXH_FORCE_INLINE XXH128_hash_t XXH64X2_avalanche(xxh_u64 const h64, xxh_u64 cons
 static XXH128_hash_t
 XXH64X2_finalize(xxh_u64 h64, xxh_u64 hi, const xxh_u8* ptr, size_t len, XXH_alignment align)
 {
-#define XXH_PROCESS1_64X2 do {                                   \
-    xxh_u64 const d = (xxh_u64)(*ptr++) * XXH_PRIME64_5;                                     \
-    h64 ^= d;                                    \
-    h64 = XXH_rotl64(h64, 11) * XXH_PRIME64_1;                 \
-    hi  ^= d + h64;                                    \
-} while (0)
-
-#define XXH_PROCESS4_64X2 do {                                   \
-    xxh_u64 const d = (xxh_u64)(XXH_get32bits(ptr)) * XXH_PRIME64_1; \
-    ptr += 4;                                              \
-    h64 ^= d;      \
-    h64 = XXH_rotl64(h64, 23) * XXH_PRIME64_2 + XXH_PRIME64_3;     \
-    hi  ^= d + h64;      \
-} while (0)
-
-#define XXH_PROCESS8_64X2 do {                                   \
-    xxh_u64 const k1 = XXH64_round(0, XXH_get64bits(ptr)); \
-    ptr += 8;                                              \
-    h64 ^= k1;                                             \
-    h64  = XXH_rotl64(h64,27) * XXH_PRIME64_1 + XXH_PRIME64_4;     \
-    hi  ^= k1 + h64;                                    \
-} while (0)
-
-    /* Rerolled version for 32-bit targets is faster and much smaller. */
-    if (XXH_REROLL || XXH_REROLL_XXH64) {
-        len &= 31;
-        while (len >= 8) {
-            XXH_PROCESS8_64X2;
-            len -= 8;
-        }
-        if (len >= 4) {
-            XXH_PROCESS4_64X2;
-            len -= 4;
-        }
-        while (len > 0) {
-            XXH_PROCESS1_64X2;
-            --len;
-        }
-         return  XXH64X2_avalanche(h64, hi);
-    } else {
-        switch(len & 31) {
-           case 24: XXH_PROCESS8_64X2;
-                         /* fallthrough */
-           case 16: XXH_PROCESS8_64X2;
-                         /* fallthrough */
-           case  8: XXH_PROCESS8_64X2;
-                    return XXH64X2_avalanche(h64, hi);
-
-           case 28: XXH_PROCESS8_64X2;
-                         /* fallthrough */
-           case 20: XXH_PROCESS8_64X2;
-                         /* fallthrough */
-           case 12: XXH_PROCESS8_64X2;
-                         /* fallthrough */
-           case  4: XXH_PROCESS4_64X2;
-                    return XXH64X2_avalanche(h64, hi);
-
-           case 25: XXH_PROCESS8_64X2;
-                         /* fallthrough */
-           case 17: XXH_PROCESS8_64X2;
-                         /* fallthrough */
-           case  9: XXH_PROCESS8_64X2;
-                    XXH_PROCESS1_64X2;
-                    return XXH64X2_avalanche(h64, hi);
-
-           case 29: XXH_PROCESS8_64X2;
-                         /* fallthrough */
-           case 21: XXH_PROCESS8_64X2;
-                         /* fallthrough */
-           case 13: XXH_PROCESS8_64X2;
-                         /* fallthrough */
-           case  5: XXH_PROCESS4_64X2;
-                    XXH_PROCESS1_64X2;
-                    return XXH64X2_avalanche(h64, hi);
-
-           case 26: XXH_PROCESS8_64X2;
-                         /* fallthrough */
-           case 18: XXH_PROCESS8_64X2;
-                         /* fallthrough */
-           case 10: XXH_PROCESS8_64X2;
-                    XXH_PROCESS1_64X2;
-                    XXH_PROCESS1_64X2;
-                    return XXH64X2_avalanche(h64, hi);
-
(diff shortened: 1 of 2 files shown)


Cyan4973:
Thanks for the suggestion @gzm55 .

A question on the design goal : 
is it supposed to be a 128-bit hash,
or a 64-bit hash which outputs a 128-bit return value ?

gzm55:
it is a 128-bit hash.

gzm55:
When cannot take the benefit of the SIMD instructions, a 128-bit hash is very rare with both high quality and high speed. The `XXH64` use four 64-bit integers internal state for the main 32-byte strip loop, so it could be a good candidate to output 128-bit hash. Comparing to hash twice with two different XXH64 seeds (`XXH64 + XXH64`),  the new algorithm reuse the most avalanche operations and got a better speed. The name `XXH64X2` expresses two meaning: a) it is based on `XXH64`, 2) it generate 128 bit output (64 x 2)

easyaspi314:
I think this is rather redundant. We already have XXH3 which is much faster.

Cyan4973:
Well, the suggested `XXH64x2` uses a design that has been envisioned as a possible backup strategy for `XXH128`. If `XXH3` had not worked as intended, I probably would have considered something similar, aka an `XXH64` variant.
There are a few things that made me consider it a "second" choice rather than a first, and so it's interesting to ask a few questions about that.

- Performance : Not all cpus have strong 64-bit multipliers. Modern Intel (and AMD) cpus do, and even some high end ARM ones. But most others have much weaker performance, because it costs a lot of gates. Using 64-bit multiplications on such cpus leads to sub-par performance (see for a recent example : https://github.com/Cyan4973/xxHash/pull/519#issuecomment-807375099). Yet, strong cpus also tend to feature a vector unit. In other words, cpus without a vector unit tend to have a poor 64-bit multiplier, making `XXH64` also a poor fit. 
So, the question is : 
    + in which environment are you unable to take advantage of a vector unit, yet can benefit from a good 64-bit multiplier ?
- The core loop seems relatively unmodified (I guess it must be, in order to ensure that the low 64-bit remain identical to regular `XXH64`). That means that, although internal states is 256-bit, it's still composed of independent 64-bit lanes. Therefore, the collision ratio might be closer to 64-bit rather than 128-bit.
    + have you measured it ?
    + have you taken some precautions to ensure 128-bit of entropy ?  
- `smhasher` was really meant to measure entropy of 32-bit hashes. It's just unable to provide accurate measurements for 64-bit hashes, let alone 128-bit hashes.
    + An incomplete yet useful work-around is to cut the output into 32-bit slices, and test `smhasher` on each one of them. 
    + It seems this could be useful for the high part of `xxh64x2`, since the low one is equivalent to regular `xxh64` which was already validated using this method.

easyaspi314:
> In other words, cpus without a vector unit tend to have a poor 64-bit multiplier, making XXH64 also a poor fit.

Indeed. That is my main concern. ARM64 and x86_64 machine both require NEON and SSE2, respectively, where the XXH3 implementation is _significantly_ faster than XXH64.

Additionally, XXH64 has some known flaws in its core loop. 

It feels like developing a new sword when lightsabers already exist.

Cyan4973:
There _might_ be a valid reason.
Let just @gzm55 answer.

gzm55:
> in which environment are you unable to take advantage of a vector unit, yet can benefit from a good 64-bit multiplier ?

The JVM on x64 is this kind of environment. After trying to port the xxh3/128 to native java code, the performance is 30%~40% of xxhash64. Even jni or new vector api of jdk16 still take much overhead and cannot keep up the the xxhash64.

> have you measured it ?

Yes, all the collisions tests on the xxh128 from the wiki are passed (zero collisions). And on the high 64 bit part, 294 collisions are found within 100G inputs with length of  256byte.

> have you taken some precautions to ensure 128-bit of entropy ?

`xxh64x2` borrow some of methods from `xxh128` for generating 128 bit results from internal state. And two independent hashes with 64-bit of entropy can be combined as a 128-bit hash with full width entropy.

> It seems this could be useful for the high part of xxh64x2

For high 64 bits part, the `smhasher` tests and collision tests are both passed.
For two 32bits parts (65-96 and 97-128) pass all the smhashers tests except LongNeighbors giving some warnings:
- Among 1455 tested bases, we expected 323.171 bads. Actually there were 307. The most striking collision has surprise score 110.765 and length 72 bytes
- Among 1455 tested bases, we expected 323.171 bads. Actually there were 335. The most striking collision has surprise score 100.045 and lengths 25 and 17 bytes

easyaspi314:
> The JVM on x64 is this kind of environment.

Good point, I wasn't even considering virtual machines.

> After trying to port the xxh3/128 to native java code, the performance is 30%~40% of xxhash64.

It is true that *scalar* XXH3 is slower than XXH64. However, that seems to be a bit much.

> Even jni or new vector api of jdk16 still take much overhead and cannot keep up the the xxhash64.

What about a hybrid JNI/Java version? The overhead of JNI is pretty nasty and the short keys are all available in Java intrinsics, so it is only worth it on long keys.

Looking at the new vector API, I doubt XXH3 will work well with it, as it seems rather half-baked. It lacks shifting [wtf?!] and long multiplication.

I would like to see those implementations, though. Perhaps they can be improved.

gzm55:
> What about a hybrid JNI/Java version? The overhead of JNI is pretty nasty and the short keys are all available in Java intrinsics, so it is only worth it on long keys.

yes, this is a major candidate method, and the split length becomes another tuning point.

> Looking at the new vector API, I doubt XXH3 will work well with it, as it seems rather half-baked. It lacks shifting [wtf?!] and long multiplication.

It is still under incubator, and cannot 1-to-1 map the api to instructions of X64 arch. This set of api would not graduates in LTS JDK17, so even the api can boost the jdk to control all instructions, the industry (big data, spring, etc.) may need at least a decade to accept and deploy the new LTS JDK with vector api.

NB: OpenHFT/Zero-Allocation-Hashing#58 is the draft of pure java porting of xxh3.

Anyway, imo, `XXH64X2`, as a fast scalar version of 128-bit hash, could have its position, although it depends on 64bit multiplications.

Cyan4973:
I spent some time testing the proposed variant `xxh64x2`.
In a test generating 100 Billion hashes, I got 1 collision.
This may not sound like much, but such an occurrence is supposed to have an unbelievably small probably, at a 1 / 59029581036420320 rate, so it doesn't feel like "just bad luck".

While it's fairly difficult to pinpoint exactly the limit,
it seems to me that this 128-bit hash doesn't provide 128-bit of entropy.
It's still larger than 64-bit, but rather around ~80-bit,
Unfortunately, I can't access a platform powerful enough to more accurately estimate the entropy.

Depending on the target application, it may still be good enough.
But better be aware of it.

gzm55:
> I spent some time testing the proposed variant `xxh64x2`.
> In a test generating 100 Billion hashes, I got 1 collision.
> This may not sound like much, but such an occurrence is supposed to have an unbelievably small probably, at a 1 / 59029581036420320 rate, so it doesn't sound like "bad luck".
> 
> While it's fairly difficult to pinpoint exactly the limit,
> it seems to me that this 128-bit hash doesn't provide 128-bit of entropy.
> It's still larger than 64-bit, but rather around ~80-bit,
> Unfortunately, I can't access a platform powerful enough to more accurately estimate the entropy.
> 
> Depending on the target application, it may still be good enough.
> But better be aware of it.

@Cyan4973 does the `xxh128` pass with similar test and got 0 collision? and I also wonder for larger hash algos (256+ bits), how to prove the entropy bit.

Cyan4973:
> does the xxh128 pass with similar test and got 0 collision? 

Yes it does.

Arguably, it's not a proof that `xxh128` has achieved real 128-bit entropy. It's more a "minimum" to achieve.

In the future, systems with more RAM would make it possible to increase the measurement accuracy. I doubt we'll ever have enough to reach 128-bit, but given that the hash doesn't target cryptographic quality, that's not necessarily a "big problem". Even ~100-bit entropy is plenty-enough for most applications out there.

gzm55:
@Cyan4973 [the collision test](https://github.com/Cyan4973/xxHash/tree/dev/tests/collisions) hash samples with a fixed length. It could be better to sample strings over a range of length, which is more like the real use cases and avoid the overfit of segment hash algorithm.

easyaspi314:
I am guessing that this collision is related to #54.

Cyan4973:
Closing. I applaud the effort, but it's difficult to introduce this variant within `xxHash` library without also introducing confusion for library users. From an ecosystem perspective, I would prefer to spend the effort ensuring that `XXH128` is fast enough for your needs, though I understand that specific setups, such as JVM on x64, may play against its natural strengths.

PR #519, opened 2021-03-25
AArch64 tuning, put asm guard in macro

easyaspi314:
- AArch64 does not benefit enough from unrolling XXH64_finalize to
   justify the code size increase.
 - Include aarch64 in the XXH32 asm guard. Clang autovectorizes this incorrectly. 2.5->4 GB/s on Clang 11 + Snapdragon 730G.
 - Replace the asm guards with a macro.

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -1394,6 +1394,27 @@ static void* XXH_memcpy(void* dest, const void* src, size_t size)
 /* note: use after variable declarations */
 #define XXH_STATIC_ASSERT(c)  do { enum { XXH_sa = 1/(int)(!!(c)) }; } while (0)
 
+/*!
+ * @internal
+ * @def XXH_COMPILER_GUARD(var)
+ * @brief Used to prevent unwanted optimizations for @p var.
+ *
+ * It uses an empty GCC inline assembly statement with a register constraint
+ * which forces @p var into a general purpose register (eg eax, ebx, ecx
+ * on x86) and marks it as modified.
+ *
+ * This is used in a few places to avoid unwanted autovectorization (e.g.
+ * XXH32_round()). All vectorization we want is explicit via intrinsics,
+ * and _usually_ isn't wanted elsewhere.
+ *
+ * We also use it to prevent unwanted constant folding for AArch64 in
+ * XXH3_initCustomSecret_scalar().
+ */
+#ifdef __GNUC__
+#  define XXH_COMPILER_GUARD(var) __asm__ __volatile__("" : "+r" (var))
+#else
+#  define XXH_COMPILER_GUARD(var) ((void)0)
+#endif
 
 /* *************************************
 *  Basic Types
@@ -1734,13 +1755,12 @@ static xxh_u32 XXH32_round(xxh_u32 acc, xxh_u32 input)
     acc += input * XXH_PRIME32_2;
     acc  = XXH_rotl32(acc, 13);
     acc *= XXH_PRIME32_1;
-#if defined(__GNUC__) && defined(__SSE4_1__) && !defined(XXH_ENABLE_AUTOVECTORIZE)
+#if (defined(__SSE4_1__) || defined(__aarch64__)) && !defined(XXH_ENABLE_AUTOVECTORIZE)
     /*
      * UGLY HACK:
-     * This inline assembly hack forces acc into a normal register. This is the
-     * only thing that prevents GCC and Clang from autovectorizing the XXH32
-     * loop (pragmas and attributes don't work for some reason) without globally
-     * disabling SSE4.1.
+     * A compiler fence is the only thing that prevents GCC and Clang from
+     * autovectorizing the XXH32 loop (pragmas and attributes don't work for some
+     * reason) without globally disabling SSE4.1.
      *
      * The reason we want to avoid vectorization is because despite working on
      * 4 integers at a time, there are multiple factors slowing XXH32 down on
@@ -1765,22 +1785,11 @@ static xxh_u32 XXH32_round(xxh_u32 acc, xxh_u32 input)
      *   can load data, while v3 can multiply. SSE forces them to operate
      *   together.
      *
-     * How this hack works:
-     * __asm__(""       // Declare an assembly block but don't declare any instructions
-     *          :       // However, as an Input/Output Operand,
-     *          "+r"    // constrain a read/write operand (+) as a general purpose register (r).
-     *          (acc)   // and set acc as the operand
-     * );
-     *
-     * Because of the 'r', the compiler has promised that seed will be in a
-     * general purpose register and the '+' says that it will be 'read/write',
-     * so it has to assume it has changed. It is like volatile without all the
-     * loads and stores.
-     *
-     * Since the argument has to be in a normal register (not an SSE register),
-     * each time XXH32_round is called, it is impossible to vectorize.
+     * This is also enabled on AArch64, as Clang autovectorizes it incorrectly
+     * and it is pointless writing a NEON implementation that is basically the
+     * same speed as scalar for XXH32.
      */
-    __asm__("" : "+r" (acc));
+    XXH_COMPILER_GUARD(acc);
 #endif
     return acc;
 }
@@ -2149,12 +2158,14 @@ typedef XXH64_hash_t xxh_u64;
  * also slightly faster because it fits into cache better and is more likely
  * to be inlined by the compiler.
  *
+ * Unrolling XXH64 is also disabled on AArch64. While it is a 64-bit platform,
+ * there isn't enough benefit to justify the larger code size.
+ *
  * If XXH_REROLL is defined, this is ignored and the loop is always rerolled.
  */
 #ifndef XXH_REROLL_XXH64
 #  if (defined(__ILP32__) || defined(_ILP32)) /* ILP32 is often defined on 32-bit GCC family */ \
    || !(defined(__x86_64__) || defined(_M_X64) || defined(_M_AMD64) /* x86-64 */ \
-     || defined(_M_ARM64) || defined(__aarch64__) || defined(__arm64__) /* aarch64 */ \
      || defined(__PPC64__) || defined(__PPC64LE__) || defined(__ppc64__) || defined(__powerpc64__) /* ppc64 */ \
      || defined(__mips64__) || defined(__mips64)) /* mips64 */ \
    || (!defined(SIZE_MAX) || SIZE_MAX < ULLONG_MAX) /* check limits */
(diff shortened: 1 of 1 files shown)


easyaspi314:
BTW, I got a new phone, a Pixel 4a, as my Pixel 2 XL broke.

It has a Snapdragon 730g which has 2x 2.2 GHz Cortex-A76 and 6x 1.8 GHz Cortex-A55.

It seems that NEON and 32-bit arithmetic are better, while 64-bit arithmetic is slightly worse (it still only appears to have a 32->64-bit hardware multiplier).

<details><summary><code>xxhsum -b99</code></summary>

Clang 11.1.0:
```
xxhsum 0.8.0 by Yann Collet
compiled as 64-bit aarch64 + NEON little endian with Clang 11.1.0 (https://github.com/termux/termux-packages 07320e6ed6e28aafc95d8af42d0da9a33bc3b79a)
Sample of 100 KB...
 1#XXH32                         :     102400 ->    41987 it/s ( 4100.3 MB/s)
 2#XXH32 unaligned               :     102400 ->    41765 it/s ( 4078.7 MB/s)
 3#XXH64                         :     102400 ->    28714 it/s ( 2804.1 MB/s)
 4#XXH64 unaligned               :     102400 ->    28713 it/s ( 2804.0 MB/s)
 5#XXH3_64b                      :     102400 ->    87600 it/s ( 8554.7 MB/s)
 6#XXH3_64b unaligned            :     102400 ->    84826 it/s ( 8283.8 MB/s)
 7#XXH3_64b w/seed               :     102400 ->    90471 it/s ( 8835.0 MB/s)
 8#XXH3_64b w/seed unaligned     :     102400 ->    87562 it/s ( 8551.0 MB/s)
 9#XXH3_64b w/secret             :     102400 ->    83114 it/s ( 8116.6 MB/s)
10#XXH3_64b w/secret unaligned   :     102400 ->    79992 it/s ( 7811.7 MB/s)
11#XXH128                        :     102400 ->    87572 it/s ( 8552.0 MB/s)
12#XXH128 unaligned              :     102400 ->    84957 it/s ( 8296.6 MB/s)
13#XXH128 w/seed                 :     102400 ->    90164 it/s ( 8805.0 MB/s)
14#XXH128 w/seed unaligned       :     102400 ->    87904 it/s ( 8584.4 MB/s)
15#XXH128 w/secret               :     102400 ->    82214 it/s ( 8028.7 MB/s)
16#XXH128 w/secret unaligned     :     102400 ->    79358 it/s ( 7749.8 MB/s)
17#XXH3_stream                   :     102400 ->    81928 it/s ( 8000.8 MB/s)
18#XXH3_stream unaligned         :     102400 ->    77774 it/s ( 7595.1 MB/s)
19#XXH3_stream w/seed            :     102400 ->    81832 it/s ( 7991.4 MB/s)
20#XXH3_stream w/seed unaligned  :     102400 ->    77802 it/s ( 7597.8 MB/s)
21#XXH128_stream                 :     102400 ->    81742 it/s ( 7982.6 MB/s)
22#XXH128_stream unaligned       :     102400 ->    77701 it/s ( 7587.9 MB/s)
23#XXH128_stream w/seed          :     102400 ->    81720 it/s ( 7980.4 MB/s)
24#XXH128_stream w/seed unaligne :     102400 ->    77665 it/s ( 7584.5 MB/s)
```
GCC 10.2.0:
```
xxhsum 0.8.0 by Yann Collet
compiled as 64-bit aarch64 + NEON little endian with GCC 10.2.0
Sample of 100 KB...
 1#XXH32                         :     102400 ->    41926 it/s ( 4094.3 MB/s)
 2#XXH32 unaligned               :     102400 ->    41452 it/s ( 4048.1 MB/s)
 3#XXH64                         :     102400 ->    28713 it/s ( 2804.0 MB/s)
 4#XXH64 unaligned               :     102400 ->    28714 it/s ( 2804.1 MB/s}
 5#XXH3_64b                      :     102400 ->    90464 it/s ( 8834.3 MB/s)
 6#XXH3_64b unaligned            :     102400 ->    79840 it/s ( 7796.9 MB/s)
 7#XXH3_64b w/seed               :     102400 ->    90278 it/s ( 8816.2 MB/s)
 8#XXH3_64b w/seed unaligned     :     102400 ->    79609 it/s ( 7774.3 MB/s)
 9#XXH3_64b w/secret             :     102400 ->    83909 it/s ( 8194.3 MB/s)
10#XXH3_64b w/secret unaligned   :     102400 ->    74449 it/s ( 7270.4 MB/s)
11#XXH128                        :     102400 ->    90810 it/s ( 8868.1 MB/s)
12#XXH128 unaligned              :     102400 ->    79283 it/s ( 7742.4 MB/s)
13#XXH128 w/seed                 :     102400 ->    89881 it/s ( 8777.4 MB/s)
14#XXH128 w/seed unaligned       :     102400 ->    79511 it/s ( 7764.8 MB/s)
15#XXH128 w/secret               :     102400 ->    83923 it/s ( 8195.6 MB/s)
16#XXH128 w/secret unaligned     :     102400 ->    74333 it/s ( 7259.1 MB/s)
17#XXH3_stream                   :     102400 ->    69356 it/s ( 6773.1 MB/s)
18#XXH3_stream unaligned         :     102400 ->    59634 it/s ( 5823.6 MB/s)
19#XXH3_stream w/seed            :     102400 ->    69271 it/s ( 6764.8 MB/s)
20#XXH3_stream w/seed unaligned  :     102400 ->    59652 it/s ( 5825.4 MB/s)
21#XXH128_stream                 :     102400 ->    69317 it/s ( 6769.3 MB/s)
22#XXH128_stream unaligned       :     102400 ->    59561 it/s ( 5816.5 MB/s)
23#XXH128_stream w/seed          :     102400 ->    69238 it/s ( 6761.5 MB/s)
24#XXH128_stream w/seed unaligne :     102400 ->    59527 it/s ( 5813.2 MB/s)
```
</details>

Cyan4973:
Thanks @easyaspi314 !

For comparison, my current laptop offers a speed of 5 GB/s for `XXH32()`,
so your new smartphone seems to get pretty close.

Cyan4973 (approved):

xxhash.h
```diff
@@ -2149,12 +2158,14 @@ typedef XXH64_hash_t xxh_u64;
  * also slightly faster because it fits into cache better and is more likely
  * to be inlined by the compiler.
  *
+ * Unrolling XXH64 is also disabled on AArch64. While it is a 64-bit platform,
```
Cyan4973, line 2161: Q: Is there a scenario (architecture) in which unrolling XXH64 finalization is a net gain ?

easyaspi314: I'm starting to think that this isn't the case.

Was the unrolling added before XXH64? **Edit:** definitely not.

Cyan4973: The unrolling was added after `XXH64` creation, but before `XXH3` existed.

I think the contributor was mostly interested in `XXH32`, and stated that it was beneficial for speed on his target.
The `XXH64` extension was a stretch, along the line "if it works for `XXH32`, it should work for `XXH64`".

Thing is, back in these days, my ability to just measure such an impact was quite limited : xxHash was still very much intended as a checksum algorithm for large input. So I could not really tell if it was better or worse on small data, it just seemed "about the same" (I also did not pay attention to the binary size increase).

Anyway, today I've got a benchmark program specifically created for performance comparison on small data, so let's use that and compare.

Cyan4973:
I made a few measurements on my desktop workstation (core i7-9700k),
comparing the "loop" and "switch" initializers of `XXH64` using `benchHash`, and here are the learnings : 

- **Throughput on fixed size inputs** : this one is a small win for "switch", although it's not large (up to +10%) and doesn't last long : the advance gets thinner as input size grows, and by the time it reaches 32 bytes, both variants are essentially equivalent. 
    This scenario is the easiest to design and measure, and likely the one that was tested at the time the "switch" finalizer was introduced.
- **Throughput on variable size inputs** : this one is a _very clear victory_ for "loop", with gains reaching up to +100% speed, converging towards a healthy ~+30% as input size grows. This is a big surprise, as the reasoning for the introduction of the "switch" finalizer was that, sure, the jump is unpredictable, but there is only one of them, whereas there are multiple unpredictable branches in the "loop" initializer. It appears that the cost of unpredictability for the jump is much higher.
- **Latency on fixed size inputs** : small surprise, though retrospectively logical : both variants perform essentially the same, at all lengths. It seems that the branch predictor makes a good job, so the job is essentially reduced to its core arithmetic operations, which are identical in both versions. 
One could wonder why the outcome is slightly different from the "throughput" scenario. It might be because it's cheaper to pipeline a single predictable jump than multiple consecutive branches, leaving a bit more room for execution units to do a bit more work. This "extra availability" is useless in the latency scenario, where each hash must be completed before starting a new one.
- **Latency on variable size inputs** : that's the most important scenarios for hashing of small inputs, and also the most difficult to create. Predictably, considering previous scenarios, the "loop" finalizer wins, decisively at all lengths, though with a reduced margin compared to the throughput scenario (up to +15%, though generally converging towards +10% as input size grows).

**Summary** : it looks to me that this points towards a victory of the "loop" finalizer (which was the original finalizer).
The only place where it loses is for throughput on inputs of fixed size, and even then only by a little. Furthermore, this is more a "benchmark" scenario, easy to measure, but I expect it to be much less frequent in real-life applications, than the more important "latency on variable size inputs" scenario.
On top of that, we already know that the "switch" finalizer is a pretty bad contender on any weaker ISA (`x86`) or architecture (`arm`), and that it weights down binary size.

So it looks pretty clear to me that we would be better off without it.

Cyan4973:
After checking the situation for the `switch` finalizer of `XXH64`,
I felt it would be a good idea to do the same evaluation for the `switch` finalizer of `XXH32`.

So I ran the same set of tests, comparing the `switch` and `loop` finalizers,
and found ... basically the same thing.

It actually seems even more in favor of the `loop` finalizer : 
- throughput, fixed size : `switch` wins, but by an even smaller margin (~5%) and lower duration (up to `len < 16`)
- throughput, random size : _big_ win for `loop`, by up to +150%, converging towards +30% as `len` grows
- latency, fixed size : identical results for both finalizers
- latency, random size : solid win for `loop`, at all lengths, converging towards +15% as `len` grows

So it seems this is once again a clear victory for the `loop` finalizer.

I'm surprised though, as the contributor was really certain that it was helping his (underpowered) platform. Though this might be another case of measuring what's easy to measure, hence the first synthetic scenario that is slightly favorable to `switch` finalizer.

I will nonetheless try to compare on other platforms, just to be sure that the results are reproduced.
(also, the binary size win is much smaller in this case, as the `switch` finalizer is nowhere near as big as the `XXH64` one).

_edit_ : found back the commit : https://github.com/Cyan4973/xxHash/commit/3bb5bb64baff1035d562b601eec3e50f8313e1e1
the commit message seems pretty certain that it improved performance, _especially for random input sizes_, which is the reverse of this evaluation's finding. Strange ...

_edit_ : I also found back the correspondence leading to this change. The developper, Jens Bauer, sent his message directly by email, presumably because he doesn't have a github account. His target was ARM cortex & v7, with `gcc-7` mentioned as a compiler. He mentions speed benefits for the `switch` finalizer at several moments. Test results also seem to confirm nice benefits for inputs of random size, so it's all the more surprising that the situation seems to be completely reversed in the more recent benchmark measurement. Something is off....

Issue #470, opened 2020-10-28
xxhsum is unable to verify checksum of file from STDIN


lacek:
In common hash utilities (such as `md5sum` and `shasum`), the `-c` option allows input file from STDIN when the file name is `-`, e.g. for `md5sum`:

```
$ md5sum < test.txt > test.txt.md5
$ md5sum -c test.txt.md5 < test.txt
-: OK
```

When it comes to `xxhsum` (v0.8.0), error is reported:

```
$ xxhsum < test.txt > test.txt.xxh
$ xxhsum -c test.txt.xxh < test.txt
test.txt.xxh:1: Could not open or read 'stdin': No such file or directory.
1 listed file could not be read
```

Cyan4973:
Implemented by @t-mat in latest `dev` branch update

PR #528, opened 2021-05-24
Fix issue #470

t-mat:
This PR attempts to fix issue #470.  The following test works as expected.

```
$ ./xxhsum < README.md > README.md.xxh
$ cat README.md.xxh
2dfe8e9158834eb3  stdin
$ ./xxhsum -c README.md.xxh < README.md
stdin: OK
```

## Potential issue

To avoid breaking change, `xxhsum` still output `"stdin"` as a special filename for stdin.  I'm not sure we should change it to `"-"` or not.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -223,6 +223,9 @@ test-xxhsum-c: xxhsum
 	# xxhsum to/from pipe
 	./xxhsum xxh* | ./xxhsum -c -
 	./xxhsum -H0 xxh* | ./xxhsum -c -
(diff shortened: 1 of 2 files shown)


Cyan4973 (approved):

Cyan4973:
> To avoid breaking change, xxhsum still output "stdin" as a special filename for stdin. I'm not sure we should change it to "-" or not.

I'm not sure either. Let's keep it this way for the time being.

Thanks for implementing this new capability !

PR #532, opened 2021-05-26
Enable AVX512 tests on GH-Actions

t-mat:
This PR eases AVX512 issue of GitHub Actions.

## The problem

- AVX512 is relatively minor so far, but important.  This code path must be checked by CI.
- Currently, GitHub Actions' test runner system doesn't guarantee to support AVX512.
- Actually, some GH-Actions test runner has AVX512.  But AFAIK, there's no method which requests spec of the test runner.

## Possible solutions

(a) Provide private test runner which supports AVX512.  This test runner is placed outside of GitHub.  GitHub Actions recommends this solution.

(b) Use other CI which fully supports AVX512 (travis-ci, etc).  And invoke them from GitHub Actions.

(c) Invoke runtime AVX512 test only if test runner supports AVX512, in luck.

This PR uses method (c).

## This PR

(1) Always builds AVX512 code.
- CI script always indicates compile time errors and warnings such as invalid pointer aliases, etc.

(2) Runs AVX512 runtime test if test runner supports AVX512.
- Yes, it's unreliable.  But hopefully, better than nothing.

AVX512 detection logic is implemented by `grep` for `/proc/cpuinfo`.  For example:

```sh
grep -q "^flags.*\bfpu\b" /proc/cpuinfo && echo fpu-enabled || echo fpu-disabled
grep -q "^flags.*\bavx512\b" /proc/cpuinfo && echo avx512-enabled || echo avx512-disabled
grep -q "^flags.*\bavx1024\b" /proc/cpuinfo && echo avx1024-enabled || echo avx1024-disabled
```

- `-q` suppresses all normal output.
- `\b` matches space or end of line.

If test runner doesn't support AVX512, CI script shows [the following log](https://github.com/t-mat/xxHash/runs/2674150206?check_suite_focus=true#step:7:5) at "Linux x64 check results consistency > AVX512 code path".

```
This test runner does not support AVX512.
```

When something wrong, you can investigate [actual `/proc/cpuinfo`](https://github.com/t-mat/xxHash/runs/2674150206?check_suite_focus=true#step:3:39) at "Environment info" log.

## Possible issues

- Suppose we'll find actual AVX512 runtime error which is detected by GH-Actions CI.  In this case, follow up commits/PRs aren't able to rely on GH-Actions.  Because it reports AVX512 runtime error stochastically.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,9 +1,15 @@
 # Known critical issues:
-# - AVX512 related tests are disabled.  Because default environment of
+# - AVX512 related tests are incomplete.  Because default environment of
 #   GitHub Actions doesn't guarantee to support AVX512.
 #   As of May 2021, they're using Xeon E5-2673 (which doesn't support
 #   AVX512) and Xeon Platinum 8171M (which supports AVX512).
 #   See also https://github.com/actions/runner/issues/1069
+#
+#   In this CI script, it always run `make default` which compiles xxHash
+#   with AVX512 intrinsics.  But if test runner doesn't support AVX512,
+#   it doesn't run `make check` which tests runtime error/consistency.
+#   It means that this test stochastically detects a failure in AVX512
+#   code path.
 #
 # Known issues:
 # - This test script ignores exit code of cppcheck which can see under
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):

Cyan4973:
Thanks @t-mat 

It looks like the best that can be achieved given the environment limitation.

PR #538, opened 2021-06-28
Correct assertion

mattmook:


Cyan4973 (approved):

Cyan4973:
Yes, indeed ! Thanks for the fix !

PR #541, opened 2021-07-07
Create SECURITY.md

lubas569:


Cyan4973:
Well, this _might_ be interesting if it was filled, instead of a placeholder.
Anyway, `xxHash` tends to have a very small vulnerability area, just by virtue of being merely a hash function. So there's much less to expect on this front (than a regular application).

PR #542, opened 2021-07-13
regroup all xxhsum files into cli/

Cyan4973:


t-mat:
I'd like to point out [the reason of CI failure of this PR](https://github.com/Cyan4973/xxHash/pull/542/checks?check_run_id=3053919782#step:11:1395).

We can reproduce this error with the following commands

```sh
# FAIL
cd
git clone -b xxhsum_cli https://github.com/Cyan4973/xxHash.git xxHash-pull-542
cd ~/xxHash-pull-542
make clean usan test-unicode
```

Direct reason of the error seems `make test-unicode`.  But it works fine.

```sh
# OK
cd ~/xxHash-pull-542
make clean test-unicode
```

Actually, it fails when we run `test-unicode` after `usan`.

```sh
# FAIL
cd ~/xxHash-pull-542
make clean usan test-unicode
```

More specifically, it fails with the following sequence.

```sh
# FAIL (case-A)
make clean
CC=clang CXX=clang++ make V=1 test-xxhsum-c MOREFLAGS="-g -fsanitize=undefined -fno-sanitize-recover=all"
CC=clang CXX=clang++ make V=1 c90test MOREFLAGS="-g -fsanitize=undefined -fno-sanitize-recover=all"
make V=1 test-unicode
```

But the following sequence works without error.

```sh
# OK (case-B)
make clean
CC=clang CXX=clang++ make V=1 test-xxhsum-c MOREFLAGS="-g -fsanitize=undefined -fno-sanitize-recover=all"
make V=1 test-unicode
```

The error message is fairly cryptic.

```
/usr/bin/ld: cli/xxhsum.o: relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a PIE object; recompile with -fPIE
/usr/bin/ld: cli/xsum_os_specific.o: relocation R_X86_64_32 against `.data' can not be used when making a PIE object; recompile with -fPIE
/usr/bin/ld: cli/xsum_sanity_check.o: relocation R_X86_64_32 against `.data' can not be used when making a PIE object; recompile with -fPIE
```

But in this case, it means we're trying to link normal object (`cli/xxhsum.o`) with sanitized objects.

As for `case-A`,

(1) `make test-xxhsum-c` compiles all source files as a sanitized object.
(2) `make c90test` compiles and removes `xxhsum.o`
(3-1) `test-unicode` invokes `make -C tests test_unicode`
(3-2) `-C tests test_unicode` invokes `make -C .. xxhsum`
(3-3) `-C .. xxhsum` compiles `xxhsum.o` as a *normal* object.
(3-4) Linker error occures, because (1) and (3-3) are different type of object.

As for `case-B`, linker error doesn't happen.  Because `xxhsum.o` is not removed.

-----

Since we must to avoid to mix sanitized and normal object, I think we should add `$(MAKE) clean` as a last command of `usan`.

Cyan4973:
Thanks @t-mat for the excellent investigation, as always.

I got rid of the problem by re-ordering the tests grouped under `test-all`.
Your suggestion (adding a `make clean` order) would likely work too.

The fix means that there is a dependency order between the tests, which is not ideal, and forbid running them in parallel. That being said, if that was the objective, there are a ton of other problems that prevent these tests to be run in parallel anyway. Even `make clean` orders become a problem for a parallel pipeline.

Thankfully, we don't need to make this action a priority, as tests do not last that long to begin with, so running them serially is not an issue.

Issue #543, opened 2021-07-20
Easier implementation for software running on "normal" architectures


fcorbelli:
I wish there was a simplified implementation of xxhash, in a single cpp file, without #ifdef, a ~6KB file instead of ~145KB

Something like this
https://create.stephan-brumme.com/xxhash/
but for XXH3 (the 128 bit version), just to get (snippet)
```
    XXH3_state_t state128;
    (void)XXH3_128bits_reset(&state128);
	
	size_t readSize;
	while ((readSize = fread(buffer, 1, blockSize, inFile)) > 0) 
	{
		(void)XXH3_128bits_update(&state128, buffer, readSize);
	}
```

In fact, I find it particularly difficult to debug other programs in which XXH3 is "integrated", it is really difficult to understand the interaction of all the #ifdefs.

Often you have "normal" environments (x86 / x64 machines) in which to use xxhash (actually a great program, when ... it works!)

I'm having trouble with a software that compiles and works fine with gcc on Windows and FreeBSD but not Linux, due to XXH3_state_t

Thank you

Cyan4973:
> I wish there was a simplified implementation of xxhash, in a single cpp file, without #ifdef, a ~6KB file

I wish it too.
I like Stefan's variants.

Problem is, it would violate one of the rules of this repository, which is "one `xxhash.h` file for all", on the ground that it's easier to integrate.

The objective is not all lost though. We could consider smaller specialized files, which are then aggregated into a single `xxhash.h` file. This way, both approaches (small specialized files + single file for all) would be possible.

However, this is a complex goal, and I don't see that happening on short term.
For your immediate needs, better consider a fix.

> I'm having trouble with a software that compiles and works fine with gcc on Windows and FreeBSD but not Linux, due to XXH3_state_t

Could you please describe your issue ?

`XXH3` is supposed to work fine on Linux.
It's actually continuously tested on Linux.

fcorbelli:
It's a 64-bit alignment issue with the gcc 10.3 compiler (Ubuntu's 21 default) and the -O3 option.

You can find the ongoing thread here
https://encode.su/threads/3664-C-expert-(on-Linux)-needed-with-xxhash

Short version: if into a struct that do a lot of other things you insert a XXH3_state_t (that should be 64 byte aligned) very nasty things happens turning on optimizations.
```
struct DT 
{
	int64_t date;          	// decimal YYYYMMDDHHMMSS (UT) or 0 if deleted
[...]

PR #545, opened 2021-07-22
Use `alignas()` in C++11

felixhandte:
The previous macro test only detected C11 and failed in modern C++, which
actually goes one step further and makes `alignas` a keyword. It's not clear
that this actually improves the situation with respect to #543, but it should
be slightly more correct in some sense.

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -915,8 +915,11 @@ struct XXH64_state_s {
    XXH64_hash_t reserved64;   /*!< Reserved field. Do not read or write to it, it may be removed. */
 };   /* typedef'd to XXH64_state_t */
 
-#if defined (__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L)   /* C11+ */
+#if defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L) /* >= C11 */
 #  include <stdalign.h>
+#  define XXH_ALIGN(n)      alignas(n)
+#elif defined(__cplusplus) && (__cplusplus >= 201103L) /* >= C++11 */
+/* In C++ alignas() is a keyword */
 #  define XXH_ALIGN(n)      alignas(n)
 #elif defined(__GNUC__)
 #  define XXH_ALIGN(n)      __attribute__ ((aligned(n)))


Cyan4973 (approved):

Cyan4973:
Thanks @felixhandte .
Your PR looks good to me.

Before merging it, I was hoping I could find some time to create a test scenario which would fail, as in #543.
This would help determine if this PR is good enough to fix it, or if there's more to it.

t-mat:
First of all, I think this PR is very good for compatibility.


## How we can test this PR?

If we want to see this PR works or not, we need a C++ compiler which satisfy the following conditions

~(1) It doesn't support C11.  (to fail `(__STDC_VERSION__ >= 201112L)`)~
(2) It supports C++11.
(3) It doesn't define `__GNUC__` and `_MSC_VER`.

~I have no idea how we can find C++ compiler which satisfies these requirements.~

~MSVCs (with default compiler switch) may satisfy (1) and (2).  But they don't pass (3).~

edit#1 : Remove condition (1).
edit#2 : Perhaps Intel compiler may pass these requirements.


## Test for alignment of XXH3_state_t and its members

As for both of C and C++, I think if the following tests are `true`, it guarantees proper alignment of `XXH3_state_t` and its members.

- `alignof(XXH3_state_t) % 64 == 0`
- `offsetof(XXH3_state_t, acc) % 64 == 0`
- `offsetof(XXH3_state_t, customSecret) % 64 == 0`
- `offsetof(XXH3_state_t, buffer) % 64 == 0`

For example,

```C++
/* We should compile this file as C and C++ code.
   Perhaps C89, C11, C17, C++98, C++11, C++17.
   For now, C89 and C++98 may fail this test. */
#include <stdalign.h>               /* C11: alignas() macro */
#include <stdio.h>

#define XXH_STATIC_LINKING_ONLY     /* access advanced declarations */
#define XXH_IMPLEMENTATION          /* access definitions */
#include "../xxhash.h"

int main(int argc, char** argv) {
    XXH_STATIC_ASSERT((alignof(XXH3_state_t)                % 64) == 0);
    XXH_STATIC_ASSERT((offsetof(XXH3_state_t, acc)          % 64) == 0);
    XXH_STATIC_ASSERT((offsetof(XXH3_state_t, customSecret) % 64) == 0);
    XXH_STATIC_ASSERT((offsetof(XXH3_state_t, buffer)       % 64) == 0);

    printf("alignof(XXH3_state_t)                : %zd\n", alignof(XXH3_state_t));
    printf("offsetof(XXH3_state_t, acc)          : %zd\n", offsetof(XXH3_state_t, acc));
    printf("offsetof(XXH3_state_t, customSecret) : %zd\n", offsetof(XXH3_state_t, customSecret));
    printf("offsetof(XXH3_state_t, buffer)       : %zd\n", offsetof(XXH3_state_t, buffer));

    return EXIT_SUCCESS;
}
```


## Issue #543

I think it is user side issue.  That misalignment is caused by user's code.

- Before C++17, user must allocate aligned memory by themselves if user needs special alignment.
- Since C++17, `operator new()` respects `alignof()` the object.

The root cause of the issue is misaligned memory of their own struct.

- `offsetof(THEIR_STRUCT, file_xxh3) % 64 == 0` is always true.  It means `XXH3_state_t` is declared and aligned properly.
- If they use C++11 or earlier compiler, they must guarantee `((uintptr_t) pointer_of_their_struct) % alignof(THEIR_STRUCT) == 0` by hand.  But it seems they failed to maintan this requirement.

Cyan4973:
note : the author of #543 said that he produced the bug with gcc 10.3 .

Cyan4973:
I've been trying to reproduce the issue described in #543, and have been unable to, even using `g++-10`, even using `-O3`, `-std=c++11`, `placeholder[1]`, etc.

The only case I could produce was a manual violation of the alignment contract, by positioning the structure into a non-aligned buffer, resulting obviously into an alignment issue. This is the scenario described by @t-mat .

So indeed, from a code standard perspective, such an issue is generated user-side.

Now, I can easily see how this could happen, with the user merely using a `malloc()` or `new()` invocation without providing any alignment information, and then assuming all is good, which it is not when alignment on 64-bytes boundaries is required.

Maybe there is a need to improve documentation to underline this point.

_PS_ : I've been looking at the documentation in the code,
and honestly, the alignment requirement looks to me correctly documented by @easyaspi314 .
I don't see how it could be made even more clear.

t-mat:
> PS : I've been looking at the documentation in the code,
> and honestly, the alignment requirement looks to me correctly documented by @easyaspi314 .
> I don't see how it could be made even more clear.

I agree with you.  It seems difficult to improve the document since it has precise notes about the alignment.

Perhaps, we can add explicit `assert()` for alignment in `XXH3_reset_internal()`.

```C
	static void XXH3_reset_internal(XXH3_state_t* statePtr, ...) {
	    XXH_ASSERT(offsetof(XXH3_state_t, nbStripesPerBlock) > initStart);
+	    XXH_ASSERT( ((size_t) statePtr) % 64 == 0); /* Please check alignment and offset of XXH3_state_t.
+	        If XXH3_state_t is the member variable of your struct/class (YOUR_STRUCT), please check the following
+	        (1) Pointer of your struct is aligned properly.
+	            ((uintptr_t)pointer_of_your_struct) % alignof(YOUR_STRUCT) == 0
+	        (2) Offset of XXH3_state_t is aligned properly.
+	            offsetof(YOUR_STRUCT, YOUR_xxh3_state_t) % alignof(XXH3_state_t) == 0 
+	        See also: XXH3_state_s.
+	        */
	    ...
	}
```

I think confusion is the main problem of #543.  And it may be caused by the cryptic runtime error message (e.g. `SEGFAULT`, `Address error`, etc).
Some `assert()` may ease to find right path to detect their own problem.

PR #549, opened 2021-08-06
Fix technical UB + compiler warnings

pdillinger:
Summary:
* XXH*_update functions technically have UB on pointer arithmetic
beyond the bounds of the input object, which can trigger a compiler
warning in some cases (e.g. statically sized input buffer). This can be
suppressed with cast to uintptr_t where available. (In theory if your
buffer is close to an extreme of your address space, it could still
over/underflow, but eh.)

* typedef XXH_endianess in implementation is unused, so removed.
(When it was in xxhash.c it would cause compiler warning.)

Test Plan: existing tests

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -1429,6 +1429,14 @@ static void* XXH_memcpy(void* dest, const void* src, size_t size)
 #endif
 typedef XXH32_hash_t xxh_u32;
 
+#ifdef UINTPTR_MAX
+  /* Pointer arithmetic slightly beyond bounds can save some cycles but is
+   * technically UB on pointer type. Use uintptr_t when available. */
+  typedef uintptr_t xxh_u8_ptr_arith_t;
+#else
+  typedef const xxh_u8* xxh_u8_ptr_arith_t;
+#endif
+
 #ifdef XXH_OLD_NAMES
 #  define BYTE xxh_u8
 #  define U8   xxh_u8
@@ -1534,8 +1542,6 @@ static xxh_u32 XXH_read32(const void* memPtr)
 
 
 /* ***   Endianness   *** */
-typedef enum { XXH_bigEndian=0, XXH_littleEndian=1 } XXH_endianess;
-
 /*!
  * @ingroup tuning
  * @def XXH_CPU_LITTLE_ENDIAN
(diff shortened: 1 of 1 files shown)


Cyan4973:
Unfortunately `uintptr_t` is not `C90` compatible, which remains a goal of this code base.

Also FYI, note that, even in `C99` environments, `uintptr_t` is an optional type, not guaranteed to be present, thus with portability limitations.

A potential workaround could be to work with `bEnd-p` instead, though it might come with (minor) performance impact, and ensuring `p <= bEnd` now becomes critical (I believe it's guaranteed by preceding code, in which case an `assert()` would be appropriate).

pdillinger:
I don't think anything becomes more "critical," because pointer difference is signed. `bEnd - p` is probably what you (we) need to be fully safe.

For speed, can't we whenever `state->memsize + len < 32` fails,
* Observe: one of the next to if's must be true, therefore
* Go ahead and unconditionally copy v1->v4 to locals and unconditionally save before last if.
* Jump straight into loop if `state->memsize` fails, skipping the `bEnd - p < something` in some cases

?

Cyan4973:
You are right, I forgot that ptr substraction is a signed operation.

Maybe the speed improvement could be the topic a separate PR ?

pdillinger:
So my update seeks not to affect performance, but get us into defined behavior whenever uintptr_t is available. Well-defined over/underflow is still possible in exotic cases.

Cyan4973:
I'm a bit puzzled regarding this PR.
On the one hand, it works fine, that part is in no doubt. I'm also pretty convinced it won't negatively impact performance.

On the other hand, I can easily imagine any future maintainers coming to this part of the code and being puzzled by the complex casting operation. 
It "feels" complex with regards to the original intention (avoid UB pointer comparison).
It also seems to "hide" the issue whenever possible, by moving the problem into unsigned arithmetic area when it's available (`UINTPTR_MAX`), while still relying on traditional ptr arithmetic and its potential UB when it's not. So it kind of half-fix the original problem.

Cyan4973:
I'm wondering if there is a way to reproduce this issue.
I tried using a statically defined buffer (on stack), but it failed to generate any warning.
@t-mat found one, using `gcc-8` (specifically), but it doesn't  trigger all the positions modified by this PR, just one of them.
Having a reproducible case would help, both to control proposed fixes and to improve the test suite.

pdillinger:
I tried reproducing the warning in various ways but was not successful in latest xxhash, so I'm dropping this PR.  https://github.com/facebook/rocksdb/pull/8691

Issue #552, opened 2021-08-15
Annotate switch case fall throughs


TheNicker:
CPP17 allows to annotate switch case fall through using the `[[fallthrough]]` attribute instead of a raw comment.
On newer compilers this cause redundant warnings when compiling with the `-Wimplicit-fallthrough` diagnostic flag.

Example for unannotated switch case fall through at [xxhash.h#L1869](https://github.com/Cyan4973/xxHash/blob/be534ed8c9fa677f18f230735c114373b1ed7d34/xxhash.h#L1869 )

TheNicker:
Resolved in PR #553

Cyan4973:
> On newer compilers this cause redundant warnings.

I've been unable to reproduce this issue.
Could you describe some reproduction procedure ?

TheNicker:
Add `-Wimplicit-fallthrough` to the compilation command. 
I use CLang, but I believe GCC will produce the same results.

Cyan4973:
I could reproduce the issue using `-Wimplicit-fallthrough` flag with `clang-10` and `clang-11` on Ubuntu (though not macos).

Interestingly, the warning issue is still present after merging your PR #553.

```
In file included from xxhash.c:43:
./xxhash.h:1896:12: warning: unannotated fall-through between switch labels [-Wimplicit-fallthrough]
           case 8:       XXH_PROCESS4;
           ^
./xxhash.h:1896:12: note: insert '__attribute__((fallthrough));' to silence this warning                                                                                                                                                                                                                                                                      case 8:       XXH_PROCESS4;
           ^                                                                                                                                                                                                                                                                                                                                                  __attribute__((fallthrough));
./xxhash.h:1896:12: note: insert 'break;' to avoid fall-through                                                                                                                                                                                                                                                                                               case 8:       XXH_PROCESS4;
           ^                                                                                                                                                                                                                                                                                                                                                  break;
./xxhash.h:1898:12: warning: unannotated fall-through between switch labels [-Wimplicit-fallthrough]                                                                                                                                                                                                                                                          case 4:       XXH_PROCESS4;
           ^
./xxhash.h:1898:12: note: insert '__attribute__((fallthrough));' to silence this warning
           case 4:       XXH_PROCESS4;
           ^
           __attribute__((fallthrough));
./xxhash.h:1898:12: note: insert 'break;' to avoid fall-through
           case 4:       XXH_PROCESS4;
           ^
           break;
[...]

PR #553, opened 2021-08-15
added support for C and CPP [[fallthrough]] attribute

TheNicker:


TheNicker:
Fixes issue #552

t-mat:
Hi @TheNicker, the intention of this PR is very good since it helps static analysis.  But it seems there're some issues.

I'm supposing you're using `gcc-10` or greater, but this PR fails to pass our tests with `gcc-9` or earlier unfortunately.

You can reproduce these failures with the following commands.

```sh
cd
git clone https://github.com/Cyan4973/xxHash.git xxHash-pull-553
cd xxHash-pull-553
git fetch origin pull/553/head:pull_553
git checkout pull_553
git log -1
CC='gcc-9'  make clean xxhsum  # FAIL
CC='gcc-10' make clean xxhsum  # PASS
```

The reason is, when compiling *C code* with `gcc-9`

- It doesn't define `__has_c_attribute()`.  Because it `gcc` as a C compiler doesn't support it.
- But it *defines* `__has_cpp_attribute()` for *C code*.  And `__has_cpp_attribute(fallthrough)` returns `!0`.  But since the code itself is compiled as a *C code*, actual C code compilation procedure fails to understand `[[fallthrough]]`.

Honestly, I don't understand why `gcc` defines `__has_cpp_attribute()` and returns `!0` for C code.  All `__has_cpp_attribute()` must be failed for C code.

But we must deal with it.  For example,

```cpp
-	#elif defined(__has_cpp_attribute)
+	#elif defined(__has_cpp_attribute) && defined(__cplusplus)
	#   if __has_cpp_attribute(fallthrough)
	#       define XXH_FALLTHROUGH [[fallthrough]]
	#   endif
```

----

I also noticed that definition of `XXH_FALLTHROUGH` is within `#ifndef XXH_NO_LONG_LONG` block.  This PR fails the following test

```sh
CC='gcc-10' make clean c90test  # FAIL
```

Please replace the definition at the outside of the `XXH_NO_LONG_LONG` block.

----

To complete the test, please run the following commands

```sh
CC='gcc-9'  make clean test
CC='gcc-10' make clean test
```

t-mat:
For older version of `gcc`, `/* fallthrough */` is a special comment for [-Wimplicit-fallthrough](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wimplicit-fallthrough).

Please leave them as a comment.

```c
           case 12:      XXH_PROCESS4;
                         XXH_FALLTHROUGH; /* fallthrough */
```

TheNicker:
> For older version of gcc, /* fallthrough */ is a special comment for -Wimplicit-fallthrough.
Please leave them as a comment.

I've [added special case, line 568](https://github.com/Cyan4973/xxHash/pull/553/files#diff-343c0b024b7734da2547ae51a138702bd0bb2f2adc2bc7fb7eb97b6d543e9785R568) for old compilers, no warnings now.

I can see that precious time is being spent on supporting old compilers.

Does someone even use it ? C90  with xxhash ?

I've had to pull out from the garret my old 286 XT to fix these issues.  :wink:

Cyan4973 (approved):

PR #564, opened 2021-08-17
fix #563

### LINKED ISSUE #563: `tests/collisions/Makefile` needs explicit `-std=c99`
Since gcc-4.8's default compiler version is pre-C99, it reports some errors for "for" statement in `tests/collisions/main.c`.

[Actual GH-Actions error log](https://github.com/t-mat/xxHash/runs/3345023705?check_suite_focus=true#step:11:647).

```
CFLAGS=-Werror make -C tests/collisions
make[1]: Entering directory '/home/runner/work/xxHash/xxHash/tests/collisions'
gcc-4.8 -Werror -O3 -I ./ -I ../../ -I allcodecs/  -c -o main.o main.c
                                                                      
main.c: In function ‘hexRaw’:
main.c:67:5: error: ‘for’ loop initial declarations are only allowed in C99 mode
     for (size_t i=0; i<size; i++) {
     ^
...
```

Cyan4973:
ensure that building test tools preserves compilation flags
including `-std=c99` and `-std=c++11`
when invoked with `CFLAGS` pre-set
so that it properly compiles with `gcc` <= 4.8 .

reported by @t-mat.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,12 +30,15 @@ environment:
     ARCH: "ARM64"
     APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
     # note: ARM64 is not available with Visual Studio 14 2015, which is default for Appveyor
-  - COMPILER: "gcc"
-    PLATFORM: "mingw64"
-  - COMPILER: "gcc"
-    PLATFORM: "mingw32"
-  - COMPILER: "gcc"
-    PLATFORM: "clang"
+# Below tests are now disabled.
+# They are flacky on Appveyor, for various reasons.
+# Moreover, their equivalent already runs correctly on Github Actions.
+#  - COMPILER: "gcc"
+#    PLATFORM: "mingw64"
+#  - COMPILER: "gcc"
+#    PLATFORM: "mingw32"
+#  - COMPILER: "gcc"
+#    PLATFORM: "clang"
 
 install:
   - ECHO Installing %COMPILER% %PLATFORM% %ARCH%
--- a/tests/bench/Makefile
+++ b/tests/bench/Makefile
@@ -29,7 +29,8 @@
 
 CPPFLAGS += -I../..   # directory of xxHash source files
 CFLAGS   ?= -O3
-CFLAGS   += -std=c99 -Wall -Wextra -Wstrict-aliasing=1
+CFLAGS   += -Wall -Wextra -Wstrict-aliasing=1 \
+            -std=c99
 CFLAGS   += $(MOREFLAGS)   # custom way to add flags
 CXXFLAGS ?= -O3
 LDFLAGS  += $(MOREFLAGS)
--- a/tests/collisions/Makefile
+++ b/tests/collisions/Makefile
@@ -26,9 +26,10 @@
 SRC_DIRS = ./ ../../ allcodecs/
 VPATH = $(SRC_DIRS)
 CPPFLAGS += $(addprefix -I ,$(SRC_DIRS))
-CFLAGS   ?= -std=c99 \
-            -Wall -Wextra -Wconversion
-CXXFLAGS ?= -Wall -Wextra -Wconversion -std=c++11
+CFLAGS   += -Wall -Wextra -Wconversion \
+            -std=c99
+CXXFLAGS += -Wall -Wextra -Wconversion \
+            -std=c++11
 LDFLAGS  += -pthread
 TESTHASHES = 110000000
 
--- a/tests/collisions/main.c
+++ b/tests/collisions/main.c
@@ -106,7 +106,7 @@ static uint64_t avalanche64(uint64_t h64)
     return h64;
 }
 
-static unsigned char randomByte(size_t n)
+static unsigned char randomByte(uint64_t n)
 {
     uint64_t n64 = avalanche64(n+1);
     n64 *= prime64_1;
@@ -240,7 +240,7 @@ typedef struct {
     /* slab5 */
     size_t nbSlabs;
     size_t current;
-    size_t prngSeed;
+    uint64_t prngSeed;
 } sampleFactory;
 
 static void init_sampleFactory(sampleFactory* sf, uint64_t htotal)
@@ -282,7 +282,7 @@ static void free_sampleFactory(sampleFactory* sf)
 
 static void flipbit(void* buffer, uint64_t bitID)
 {
-    size_t const pos = bitID >> 3;
+    size_t const pos = (size_t)(bitID >> 3);
     unsigned char const mask = (unsigned char)(1 << (bitID & 7));
     unsigned char* const p = (unsigned char*)buffer;
     p[pos] ^= mask;
@@ -416,7 +416,7 @@ static inline int Filter_insert(Filter* bf, int bflog, uint64_t hash)
      hash >>= 8;
 
      size_t const fclmask = ((size_t)1 << (bflog-6)) - 1;
-     size_t const cacheLineNb = hash & fclmask;
+     size_t const cacheLineNb = (size_t)hash & fclmask;
 
      size_t const pos1 = (cacheLineNb << 6) + (slot1 >> 2);
      unsigned const shift1 = (slot1 & 3) * 2;
@@ -456,7 +456,7 @@ static inline int Filter_check(const Filter* bf, int bflog, uint64_t hash)
      hash >>= 8;
 
      size_t const fclmask = ((size_t)1 << (bflog-6)) - 1;
-     size_t const cacheLineNb = hash & fclmask;
+     size_t const cacheLineNb = (size_t)hash & fclmask;
 
      size_t const pos1 = (cacheLineNb << 6) + (slot1 >> 2);
      unsigned const shift1 = (slot1 & 3) * 2;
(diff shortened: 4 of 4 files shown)


Cyan4973:
The last conversion warning (Appveyor's `mingw`) is pretty strange to me : 
```
main.c: In function 'Filter_insert':
main.c:435:18: error: conversion to 'Filter {aka unsigned char}' from 'int' may alter its value [-Werror=conversion]
      bf[pos1] |= (Filter)(MAX(ex1, nextValue[existing]) << shift1);
                  ^
main.c:436:18: error: conversion to 'Filter {aka unsigned char}' from 'int' may alter its value [-Werror=conversion]
      bf[pos2] |= (Filter)(MAX(ex2, nextValue[existing]) << shift2);
                  ^
```

- It triggers a warning on the conversion, even though the conversion is explicitly set
- I don't know how it could derive that the right side is of type `int`, since all its constituents are of type `unsigned`, thus mandating that the outcome should be of type `unsigned` too.

t-mat:
FYI: [gcc-5.5.0](https://github.com/t-mat/xxHash/runs/3353067994#step:4:13) on Ubuntu-18.04 reports no warning for this code.

https://github.com/t-mat/xxHash/runs/3353067994#step:4:13

```
CFLAGS=-Werror make -C tests/collisions
make[1]: Entering directory '/home/runner/work/xxHash/xxHash/tests/collisions'
gcc-5 -Werror -std=c99 -O3 -I ./ -I ../../ -I allcodecs/  -c -o main.o main.c
gcc-5 -Werror -std=c99 -O3 -I ./ -I ../../ -I allcodecs/  -c -o pool.o pool.c
gcc-5 -Werror -std=c99 -O3 -I ./ -I ../../ -I allcodecs/  -c -o threading.o threading.c
g++-5 -Wall -Wextra -Wconversion -std=c++11 -O3 -I ./ -I ../../ -I allcodecs/  -c -o sort.o sort.cc
gcc-5 -Werror -std=c99 -O3 -I ./ -I ../../ -I allcodecs/  -c -o allcodecs/dummy.o allcodecs/dummy.c
g++-5 -I ./ -I ../../ -I allcodecs/ -Wall -Wextra -Wconversion -std=c++11 -O3 main.o pool.o threading.o sort.o allcodecs/dummy.o -pthread -o collisionsTest
make[1]: Leaving directory '/home/runner/work/xxHash/xxHash/tests/collisions'
```

edit: Compiler option is different.  I'll try to see what happens with same option.

edit2: Compiled with almost equivalent options (not same).  But there's no warning.

https://github.com/t-mat/xxHash/runs/3353255629#step:11:536

```
CFLAGS='-Werror -Wall -Wextra -Wstrict-aliasing=1' make -C tests/collisions
make[1]: Entering directory '/home/runner/work/xxHash/xxHash/tests/collisions'
gcc-5 -Werror -Wall -Wextra -Wstrict-aliasing=1 -std=c99 -O3 -I ./ -I ../../ -I allcodecs/  -c -o main.o main.c
gcc-5 -Werror -Wall -Wextra -Wstrict-aliasing=1 -std=c99 -O3 -I ./ -I ../../ -I allcodecs/  -c -o pool.o pool.c
gcc-5 -Werror -Wall -Wextra -Wstrict-aliasing=1 -std=c99 -O3 -I ./ -I ../../ -I allcodecs/  -c -o threading.o threading.c
g++-5 -Wall -Wextra -Wconversion -std=c++11 -O3 -I ./ -I ../../ -I allcodecs/  -c -o sort.o sort.cc
gcc-5 -Werror -Wall -Wextra -Wstrict-aliasing=1 -std=c99 -O3 -I ./ -I ../../ -I allcodecs/  -c -o allcodecs/dummy.o allcodecs/dummy.c
g++-5 -I ./ -I ../../ -I allcodecs/ -Wall -Wextra -Wconversion -std=c++11 -O3 main.o pool.o threading.o sort.o allcodecs/dummy.o -pthread -o collisionsTest
make[1]: Leaving directory '/home/runner/work/xxHash/xxHash/tests/collisions'
```

Cyan4973:
Yeah, `-Wconversion` is the one that triggers the warning, so it has to be selected.

But I'm puzzled as to why it's a problem for `mingw` of Appveyor, but not `mingw` on GA. Probably due to a difference in version.

Some ideas that come to mind.

- Remove `-Wconversion` flag, on the ground it's just a test tool. Not ideal, more like a quick-fix. It would preferably have to be re-enabled in some future.
- Remove `mingw` tests from Appveyor, on the ground they are already played in GA. On the plus side, it would speed up checks, since Appveyor is the limiting factor. On the other hand, it's just a way to hide the problem, not fix it.
- Rewrite `Filter_insert`, so that it doesn't need this conversion. This is a much larger rewrite though, a complete project. So it's an oversized effort compared to this minimal goal.
- Try to limit the value, with something like `& 255`, so that the compiler get a hint that it cannot overflow the destination variable, assuming that's the problem. Pretty much a trial & error magic. It would be preferable to be able to reproduce the problem first, in order to observe the fix.

t-mat:
> Try to limit the value, with something like & 255

I tried this method.  But there was no luck.

----

I'm still not able to reproduce this warning on GA and Linux environment.  But it seems `-Wconversion` is very fragile.

I made the following test code to reproduce the warning.

```C
/* gcc-9 test-1.c -std=c99 -Wconversion */
#include <stdio.h>
#include <stdint.h>   /* uint64_t */

#define MIN(a,b)   ((a) < (b) ? (a) : (b))
#define MAX(a,b)   ((a) > (b) ? (a) : (b))

typedef unsigned char Filter;

static uint64_t avalanche64(uint64_t h64)
{
    static const uint64_t prime64_2 = 14029467366897019727ULL;
    static const uint64_t prime64_3 =  1609587929392839161ULL;

    h64 ^= h64 >> 33;
    h64 *= prime64_2;
    h64 ^= h64 >> 29;
    h64 *= prime64_3;
    h64 ^= h64 >> 32;
    return h64;
}

static inline int Filter_insert(Filter* bf, int bflog, uint64_t hash)
{
    hash = avalanche64(hash);
    unsigned const slot1 = hash & 255;
    hash >>= 8;
    unsigned const slot2 = hash & 255;
    hash >>= 8;

    size_t const fclmask = ((size_t)1 << (bflog-6)) - 1;
    size_t const cacheLineNb = (size_t) (hash & fclmask);

    size_t const pos1 = (cacheLineNb << 6) + (slot1 >> 2);
    unsigned const shift1 = (slot1 & 3) * 2;
    unsigned const ex1 = (bf[pos1] >> shift1) & 3;

    size_t const pos2 = (cacheLineNb << 6) + (slot2 >> 2);
    unsigned const shift2 = (slot2 & 3) * 2;
    unsigned const ex2 = (bf[pos2] >> shift2) & 3;

    unsigned const existing = MIN(ex1, ex2);

    static const int addCandidates[4] = { 0, 2, 1, 1 };
    static const unsigned nextValue[4] = { 1, 2, 3, 3 };

    bf[pos1] &= (Filter)(~(3 << shift1)); /* erase previous value */
    bf[pos1] |= (Filter)(MAX(ex1, nextValue[existing]) << shift1);
    bf[pos2] |= (Filter)(MAX(ex2, nextValue[existing]) << shift2);

    { /* Additional code for testing */
        unsigned const a = 1;
        unsigned const b = 2;
        bf[pos2] |= (Filter) (MAX(a, b));           /* (A) */
        /*          ^
                    warning: conversion from 'int' to 'Filter'
                    {aka 'unsigned char'} may change value [-Wconversion]
        */

        bf[pos2] |= (Filter) (MAX(a, b) << shift1); /* (B) */ /* no warning */

        bf[pos2] |= (Filter) (MAX(1, 2));           /* (C) */ /* no warning */
    }

    {
        unsigned const a = 1;
        unsigned const b = 2;
        bf[pos2] = (Filter) (bf[pos2] | MAX(a, b)); /* (D) */ /* no warning */
    }

    return addCandidates[existing];
}

int main(int argc, char** argv)
{
    (void) argc;  (void) argv;  /* unused */
}
```

Please take a look at (A), (B), (C) and (D).

- (A) is almost same as previous `bf[pos2] |= ...`.  But without left shift.  All `gcc` shows `[-Wconversion]` warning.
- (B) is previous line with left shift.  There's no warning.
- (C) is same as (A) but immediate number.  There's no warning.
- (D) is the only robust way I could find to do this operation without `[-Wconversion]` warning for all `gcc` versions.

----
(edit) I :+1: for "Remove -Wconversion flag".  I do not understand the difference between (4), (5), (6) and (7) in the following code.

```c
/* gcc-9 test-2.c -std=c99 -Wconversion */
int main(int argc, char** argv)
{
    unsigned const a = 1;
    unsigned const b = 2;
    unsigned const c = 3;
    unsigned char x = 0;
    x |= (unsigned char) (a);                                /* (1) no warning */
    x |= (unsigned char) (a + b);                            /* (2) no warning */
    x |= (unsigned char) (a < b ? a : b);                    /* (3) warning [-Wconversion] */
    x |= (unsigned char) ((a < b ? a : b) << c);             /* (4) no warning */
    x |= (unsigned char) ((a < b ? a : b) << 3);             /* (5) warning [-Wconversion] */
    x |= (unsigned char) ((a < b ? a : b) << 3U);            /* (6) warning [-Wconversion] */
    x |= (unsigned char) ((a < b ? a : b) << (unsigned) 3);  /* (7) warning [-Wconversion] */

    (void) argc;  (void) argv;
}
```

Cyan4973:
I think I will rather remove those tests from Appveyor. Reasons are : 
- The `mingw32` conversion warning isn't reproduced on any other platform beyond Appveyor, making it highly specific to this version. And it doesn't make sense, which rather points to a compiler bug in this specific version.
- The `msys + clang` test fails, for reasons completely internal to its own code base (it can't compile its own `sort` standard function).
- The `mingw64` test "succeeds" because it doesn't do anything at all.

That's too many concerns to make these tests useful. I'd rather cut this flailing dependency.

PR #569, opened 2021-08-21
removing const float*

### LINKED ISSUE #559: Strange `-Wcast-align` warning from `clang-3.9`
I saw the following strange warning (as an error) from `clang-3.9`.  ( Here's [actual GH-Actions log](https://github.com/t-mat/xxHash/runs/3345251571#step:7:23) )

```
clang-3.9 -Werror -Wall -Wextra -Wconversion -Wcast-qual -Wcast-align -Wshadow \
          -Wstrict-aliasing=1 -Wswitch-enum -Wdeclaration-after-statement \
          -Wstrict-prototypes -Wundef -Wpointer-arith -Wformat-security -Wvla \
          -Wformat=2 -Winit-self -Wfloat-equal -Wwrite-strings -Wredundant-decls \
          -Wstrict-overflow=2 -Wno-sign-conversion -DXXHSUM_DISPATCH=1 -c -o xxhash.o xxhash.c

./xxhash.h:3868:56: error:
    cast from 'const xxh_u8 *' (aka 'const unsigned char *') to 'const float *'
    increases required alignment from 1 to 4 [-Werror,-Wcast-align]

    XXH_ALIGN(64)        const float* const src  = (float const*) XXH3_kSecret;
                                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I'd like to know there's good solution for this issue.

Cyan4973:
to fix #559

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -3865,19 +3865,21 @@ XXH_FORCE_INLINE XXH_TARGET_SSE2 void XXH3_initCustomSecret_sse2(void* XXH_RESTR
 #       endif
         int i;
 
-        const __m128i* const src16 = (const __m128i *)XXH3_kSecret;
-        __m128i* dest = (__m128i*) customSecret;
+        const void* const src16 = XXH3_kSecret;
+        __m128i* dst16 = (__m128i*) customSecret;
 #       if defined(__GNUC__) || defined(__clang__)
         /*
          * On GCC & Clang, marking 'dest' as modified will cause the compiler:
          *   - do not extract the secret from sse registers in the internal loop
          *   - use less common registers, and avoid pushing these reg into stack
          */
-        XXH_COMPILER_GUARD(dest);
+        XXH_COMPILER_GUARD(dst16);
 #       endif
+        XXH_ASSERT(((size_t)src16 & 15) == 0); /* control alignment */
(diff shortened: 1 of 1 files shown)


xxhash.h
```diff
@@ -3865,8 +3865,8 @@ XXH_FORCE_INLINE XXH_TARGET_SSE2 void XXH3_initCustomSecret_sse2(void* XXH_RESTR
 #       endif
         int i;
 
-        XXH_ALIGN(64)        const float* const src  = (float const*) XXH3_kSecret;
-        XXH_ALIGN(XXH_SEC_ALIGN) __m128i*       dest = (__m128i*) customSecret;
+        const __m128i* const src16 = (const __m128i *)XXH3_kSecret;
```
Cyan4973: As expected, this variant fails the `-Wcast-align` test on older `clang` version, 
as these compilers fail to notice that `XXH3_kSecret` is correctly aligned, 
since they only take into consideration the source type `const unsigned char*`, but not its alignment `XXH_ALIGN(64)`.

Note that `const float*` generates the same warning, it only differs in the amount of alignment required.

Cyan4973:
This variant should now pass the `-Wcast-align` check of `clang`, including older versions.
Its code is also less obscure.

PR #572, opened 2021-08-23
add memory access method 1 for mips on gcc

Cyan4973:
recommended by info@mobile-stream.com

> This is designed to cover all versions since the original MIPS I ISA up
to MIPS32r5 / MIPS64r5 including microMIPS and MIPS16e2. That is, all
MIPS ISA variants with LWR/LWL/SWR/SWL etc instructions. GCC is able to
use these for efficient unaligned access for many years. For example,
see the "MIPS32 Instruction Set Quick Reference" (MD00565-2B) from 2008
for the recommendation to use method 1 for unaligned access on MIPS.
>
> MIPSXXr6 is explicitly excluded because it allows for software (slow)
implementation of unaligned load/stores.
>
> nanoMIPS is not included too due to lack of testing and because the
"nanoMIPS Subset" (NMS) omits the dedicated instructions for unaligned
access. Note this means __mips should not be used instead of __mips__
and vice versa as the latter is the architecture name and the former
is the ISA feature macro. See 4.2 in "P32 Porting Guide" (DN00184).


Cyan4973:
Btw, does it sound reasonable to add a `qemu` mips test target on GA ?

_edit_ : qemu-mips test added

t-mat:
I'm not sure about the context of this change.
But I think adding `qemu-mips` is good for checking basic runtime error such as `Address error`, etc.

Since we don't have machine-cycle based benchmark, we can't keep verifying the actual performance though.

PR #578, opened 2021-08-25
Add OS/Compiler matrix to GA

### LINKED ISSUE #554: Add more compilers to CI
I think PR #553 hits the difference between `gcc-9` and `gcc-10`.
Since it's difficult to find these differences, test procedure should help contributors.

We can import lz4's matrix as a first step.

t-mat:
This change set adds test for various version of `gcc` and `clang`.  Resolves #554.

- Adds `gcc-[4.8 , 11]`
- Adds `clang-[3.9 , 12]`
- Since `gcc-4.8` doesn't support `-mavx512f`, this change set introduces special matrix parameter `avx512`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,57 +54,104 @@ name: xxHash CI tests
 on: [push, pull_request]
 
 jobs:
+  xxhash-c-compilers:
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):

Cyan4973:
Excellent work @t-mat !
This will definitely help preserving code quality and portability !

PR #579, opened 2021-08-25
WIP: added variants _withSecretandSeed()

Cyan4973:
and a companion `XXH3_generateSecret_fromSeed()`.

The new variant uses `seed` for "small" keys,
and `secret` for "large" keys,
with the cutoff point being `XXH3_MIDSIZE_MAX` = 240 bytes.

When `secret` is produced with newer `XXH3_generateSecret_fromSeed()`,
it becomes exactly the same `secret` as the corresponding `_withSeed()` variant (verified by the sanity checker),
meaning that the new `_withSecretandSeed()` variant provides exactly the same hash values as `_withSeed()`.

So what's the point then ?
Well, the new variant has the benefit of pre-calculated `secret` for "large keys",
and presuming this `secret` is re-used across multiple hash calculations,
it results in a speed boost, notably visible for "not too large" keys (<1 KB).

This is illustrated by the following benchmark, run on a Core i7-9700k with turbo disabled, compiled with `gcc` 9.3.0, and using `SSE2`, comparing the different variants of `XXH3_64bits`.

image

"long" keys start at 240 bytes, so the measurement is taken at 256 bytes, then 512, then 1024, etc.
Circa ~32KB, all these variants have essentially the same performance.
But in the early stage, when keys are "small" (yet > 240 bytes), the `_withSeed()` variants are noticeably slower, by up to 20%.
In contrast, the `_withSecret()` variants remain close to the optimal, which is the "default" variant (`XXH3_64bits()`).
Actually the variant employing a _`static`_ secret gets exactly the same performance as "default", because all pre-calculations can be folded by the compiler at compile time (it's essentially the same thing as "default", just with a different `secret`).
The new variant `XXH3_64bits_withSecretandSeed()` is closer to the scenario involving a "dynamic" `secret`, aka one that is defined at runtime, because in the measured scenario, its `secret` is indeed generated at runtime. 
In a more complex scenario, it could be possible to generate the `secret` ahead of time, then embed the `secret` in the source code, and then compile it, resulting in a `static secret` scenario, providing optimal performance.
As can be seen though, the performance difference isn't that large, so such effort is likely a stretch.

On `AVX2`, the performance difference is even more pronounced.

image

The `seed` variants tends to be slower. But in proportion, the difference is especially large at the beginning : the `secret` variants, including the newer `_withSecretandSeed()`, achieve >+50% faster throughput.
_note_ : the performance is reduced after 256 KB, because that's the size of the L2 cache on this CPU. Yep, the `AVX2` variant is faster than the L3 cache. After that point, all variants have the same bottleneck and tend to perform at the same speed.

Todo : 
- complete other variants (streaming)

This answers #416, by @gzm55.


Cyan4973:
If the goal was merely to provide a variant of `_withSeed()` hash functions,
which produces exactly the same hash results,
but with the benefit of faster speed for medium-size keys,
then the goal would be already fulfilled,
since the streaming variants already cache the `secret` between 2 invocations,
and therefore avoid the issue of rebuilding the `secret` across invocations using the same `seed`.

That being said, the new variant `_withSecretandSeed()` could see legitimate usages,
where neither `secret` nor `seed` are kept in sync.
It would then produce its own set of hash values, similar to `_withSeed()` for small keys, and similar to `_withSecret()` for large keys.
In which case, such a scenario would lack an equivalent streaming implementation.

So I presume that a more complete PR should include the corresponding streaming variants.

Cyan4973:
I believe the more important questions in this PR are about scope and design.

The original request, in #416, was demanding the existence of a derivative of `_withSeed()` which would skip the need to regenerate `secret` for each and every long key (>= 241 bytes), sharing a single `secret` generation over multiple hash calculations, assuming `seed` remains static across them.

There are many ways to reach this objective. 
And the first way would be to create a dedicated variant which would do exactly that.

Obviously, such a variant would need to _store_ the generated `secret` in some kind of `state`, so that it can be shared across multiple invocations.
The downside is the creation of a new `state` type just for this one use case.

#416 improves on the direct solution, by avoiding the need for another type of state, using instead the existing `XXH3_state_t` to store the `secret`. And it works.
The downside is that `XXH3_state_t` is relatively large state, meant for streaming. The #416 variant only uses it very partially, just to store the `secret`, but it doesn't need the accumulators nor the buffer. This could be seen as a kind of "waste" of memory.
Another downside is that the variant feels very specific, allowing usage of `XXH3_state_t` in one-pass hashing scenarios only for `_withSeed()`, which also begs the question: what happens if the `state` was not initialized with `reset_withSeed()` ? It lead to the suggestion by @easyaspi314 of a slightly altered variant with enlarged scope, `_withState()`, which would be compatible with any kind of initialization (default, `_withSeed` and `_withSecret`). However, the speed benefits are only present for the `_withSeed` variant (and even then, only for "large" keys, which are not too large...), so it's dubious if it's worth "generalizing".

The "memory waste" downside could be solved by the newly proposed variant `_withSecretandSeed()`, which, instead of a full `state`, only requires to store and present the `secret`. This reduces memory usage by a factor x4.
But it also enlarges the scope of the proposal, since `_withSecretandSeed()` only coincides with the initial request when paired with `XXH3_generateSecret_fromSeed()`. In all other cases, it's its own thing, generating its own set of hashes, increasing the scope of the library, by providing yet another variant.
This extension could be seen as a good thing (more choices) or as a nuisance (more confusion, more maintenance risks). 

Another possibility could be to use more space provided by `XXH3_state_t`, to reduce the "memory waste". Essentially, the idea would be to use the state to also store the accumulators, on top of the `secret`. This would reduce the stack memory budget to run `XXH3`. The saving is only 64-bytes, but in some environments, that's still worth saving, given that stack memory is very precious.
This would pair well with the idea of a more generalized `_withState()` variant: on top of providing the expected speed benefits for `_withSeed()`, it would also reduce stack usage for all variants, making it a bit more useful. 

So these are in essence 2 different sets of solutions, both answering the initial request, but with additional benefits and side effects. 
And the question is : 
- which direction seems better suited to answer #416 ?

cc @gzm55

gzm55:
> The original request, in #416, was demanding the existence of a derivative of _withSeed() which would skip the need to regenerate secret for each and every long key (>= 241 bytes), sharing a single secret generation over multiple hash calculations, assuming seed remains static across them.

The two pr, this and #416 both solve the performance degrade on [241, 8K] length input. As of design, i hive no preference on  neither one (#416 more o less is choice of modifying less lines to focus on the POC of this issue). This pr seems more clear to understand with little cons.

> This reduces memory usage by a factor x4.
But it also enlarges the scope of the proposal, since _withSecretandSeed() only coincides with the initial request when paired with XXH3_generateSecret_fromSeed(). 

For the scope enlargement, since the xxh3 is essential a piecewise function, how about treating `_withSecretandSeed()` as a lower level definition of xxh3, and the `secret` can not be required to generate by  `XXH3_generateSecret_fromSeed` for `expert` usage, in the meanwhile the previous stable api can also be explained by `_withSecretandSeed`, `XXH3_generateSecret_fromSeed`, and the default secrets. `_withSecretandSeed` can also be marked as INTERNAL or NON-STABLE for further evolutions.

Cyan4973:
With the introduction of streaming variants of `_withSecretandSeed()`, I believe it completes the scope of this PR.

When associated with `XXH3_generateSecret_fromSeed()`, the new variants provide exactly the service requested in #416 , aka a faster `_withSeed()` variant using a pre-calculated `secret`.

But `_withSecretandSeed()` is also more general and can be used with independently produced `secret` and `seed` values. In which case, it generates results equivalent to `_withSeed()` variant for "small" inputs (<= `XXH3_MIDSIZE_MAX` = 240 bytes) and equivalent to `_withSecret()` for "larger" inputs.

PR #581, opened 2021-08-26
Add standard static_assert()

### LINKED ISSUE #567: `XXH_STATIC_ASSERT()` doesn't work with VC++2019 as a C compiler
When we compile a code with Visual C++ 2019 as a *C compiler*, it doesn't interpret `XXH_STATIC_ASSERT(0)` properly.

If we use Visual C++ 2019 as a *C++ compiler*, `XXH_STATIC_ASSERT(0)` works correctly though.

```C
/* C code */
#define XXH_STATIC_ASSERT(c)  do { enum { XXH_sa = 1/(int)(!!(c)) }; } while (0)

int main(int argc, char** argv) {
    XXH_STATIC_ASSERT(0); /* VC++2019 as a C compiler reports no error */
}
```

- [Test at Compiler Explorer](https://godbolt.org/z/Tqc18YPza)

t-mat:
This PR resolves #567.

- Use standard `static_assert()` if available (C11, C++11).
- Change definition of non-standard XXH_STATIC_ASSERT() to support MSVC as a C compiler.

We always need `do { ... } while(0)` to support C11 compilers with `-Wdeclaration-after-statement`.

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -1501,7 +1501,7 @@ static void* XXH_memcpy(void* dest, const void* src, size_t size)
 #  elif defined(__cplusplus) && (__cplusplus >= 201103L)            /* C++11 */
 #    define XXH_STATIC_ASSERT_WITH_MESSAGE(c,m) do { static_assert((c),m); } while(0)
 #  else
-#    define XXH_STATIC_ASSERT_WITH_MESSAGE(c,m) do { char XXH_sa[(c) ? 1 : -1]; (void) XXH_sa; } while(0)
+#    define XXH_STATIC_ASSERT_WITH_MESSAGE(c,m) do { void xxh_sa(int x[1 - 2 * ((c) ? 0 : 1)]); (void) xxh_sa; } while(0)
 #  endif
 #  define XXH_STATIC_ASSERT(c) XXH_STATIC_ASSERT_WITH_MESSAGE(c,#c)
 #endif
--- a/xxhash.h
+++ b/xxhash.h
@@ -1494,7 +1494,17 @@ static void* XXH_memcpy(void* dest, const void* src, size_t size)
 #endif
 
 /* note: use after variable declarations */
-#define XXH_STATIC_ASSERT(c)  do { enum { XXH_sa = 1/(int)(!!(c)) }; } while (0)
+#ifndef XXH_STATIC_ASSERT
+#  if defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L)    /* C11 */
+#    include <assert.h>
+#    define XXH_STATIC_ASSERT_WITH_MESSAGE(c,m) static_assert((c),m)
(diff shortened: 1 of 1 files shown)


xxhash.h
```diff
@@ -1494,7 +1494,17 @@ static void* XXH_memcpy(void* dest, const void* src, size_t size)
...
 /* note: use after variable declarations */
-#define XXH_STATIC_ASSERT(c)  do { enum { XXH_sa = 1/(int)(!!(c)) }; } while (0)
+#ifndef XXH_STATIC_ASSERT
+#  if defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L)    /* C11 */
+#    include <assert.h>
+#    define XXH_STATIC_ASSERT_WITH_MESSAGE(c,m) do { static_assert((c),m); } while(0)
+#  elif defined(__cplusplus) && (__cplusplus >= 201103L)            /* C++11 */
+#    define XXH_STATIC_ASSERT_WITH_MESSAGE(c,m) do { static_assert((c),m); } while(0)
+#  else
+#    define XXH_STATIC_ASSERT_WITH_MESSAGE(c,m) do { char XXH_sa[(c) ? 1 : -1]; (void) XXH_sa; } while(0)
```
Cyan4973: It's minor, but some people believe that this formulation _may_ leave a trace, in the form of unused but allocated memory in the stack.
I presume any decent compiler optimization should be able to remove that, leaving the problem only for poorer compilers (which, in the `C` world, do exist) or potentially debug mode with optimizations off.

Anyway, this is what lead to previous design using a type definition instead, which is guaranteed to leave no trace on the stack.
But my understanding is that you are moving away from that because it doesn't work with MSVC ?

t-mat: > Anyway, this is what lead to previous design using a type definition instead, which is guaranteed to leave no trace on the stack.

Perhaps, exploiting function declaration may be suitable for this case.  For example

https://github.com/libuv/libuv/blob/6530ea2ff243161b4208aed9549fe2e4c6acc563/src/uv-common.h#L60-L61

> But my understanding is that you are moving away from that because it doesn't work with MSVC ?

Right.  MSVC [2015,2019] as a *C* compiler doesn't report any error/warning for `enum { x = 1/0; };`.
MSVC as a *C++* compiler reports proper, expected error though.

Cyan4973: > exploiting function declaration

Good idea ! I've never seen that one.
I presume the scope of the declaration can be restricted,
in which case it's as fine as the `enum` solution.

Cyan4973: So, containing function declaration scope is a problem after all.
That's probably why this variant is not so commonly seen.

One possible work-around is to alter the name of the function by combining with something like `__LINE__`.
It makes the resulting macro formula a tad more complex to setup and read though.

t-mat: I'll change it to local struct declaration.

https://godbolt.org/z/589zT3Wcs

Cyan4973 (approved):

PR #584, opened 2021-08-27
Add Unicode lint

t-mat:
This PR imports Unicode test from lz4/lz4#1020.
We introduce this test due to technical restriction to support older compilers and multi-byte text encodings.

- Original script has written by @servusdei2018.
- See also
  - lz4/lz4#1018
  - lz4/lz4#1019.

This test/rejection for Unicode is only for main source codes.

- `${dirs}` in `unicode_lint.sh` represents target directories.
- `find` command in the script specifies regex globbing for files in each directory.


Cyan4973:
Excellent PR as always @t-mat !

PR #585, opened 2021-09-01
Remove unnecessary XXH_ALIGN() from pointer declaration

t-mat:
Note that we still have assert() for each pointer value.
See also #559 and #569.


Cyan4973 (approved):

Cyan4973:
Indeed, the alignment restriction should already be part of the type definition.

PR #591, opened 2021-10-28
Improve XXH_mult64to128 performance on MSVC ARM64

lhecker:
This commit specializes `XXH_mult64to128` to use `__umulh`
on ARM64/aarch64 when compiled with MSVC.

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -3265,6 +3265,21 @@ XXH_mult64to128(xxh_u64 lhs, xxh_u64 rhs)
     r128.high64 = product_high;
     return r128;
 
+    /*
+     * MSVC for ARM64's __umulh method.
+     *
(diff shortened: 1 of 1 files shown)


lhecker:
To be entirely honest I unfortunately don't have an AARCH64 device with which I could test these changes myself. If anyone is able to and would be kind enough to test this on one of their devices, that'd be super awesome. 🙂
Otherwise I believe that this change is simple enough that it can be merged regardless.
Here's a godbolt link showing the difference in assembly: https://godbolt.org/z/TzT9cPsY7

easyaspi314:
Well I don't have an arm64 windows install, but as someone who knows a good amount of arm64 that is definitely going to be faster. 

I should look into trying to get neon to work on msvc arm64. It might just be a header swap.

Cyan4973 (approved):

Cyan4973:
Thanks for the improvement @lhecker !

PR #597, opened 2021-11-15
Fix test failures on Power when building with GCC 10

### LINKED ISSUE #596: Test fails on Power with GCC 10.3.0
On Power8/Power9 (ppc64le) I see the tests fail when building with GCC 10.3.0:

```
$ make test
...
./xxhsum -bi0
xxhsum 0.8.1 by Yann Collet
compiled as 64-bit ppc64 + POWER8 vector little endian with GCC 10.3.0
Error: 64-bit hash test 100: Internal sanity check failed!
Got 0xBB9D481D33CB87AAULL, expected 0xCDEB804D65C6DEA4ULL.
or temporarily recompile with XSUM_NO_TESTS=1.
make: *** [Makefile:196: check] Error 1
```

The failure is always the same.

If I disable strict aliasing with `-fno-strict-aliasing` all tests pass.

I see there was a pull request a while back from @easyaspi314  to fix some VSX issues: https://github.com/Cyan4973/xxHash/pull/433

But it was never merged in any form as far as I can see.

If I take the latter two commits from that pull request:

  https://github.com/Cyan4973/xxHash/pull/433/commits/2e9033d06d98fb321f42f633189a4a53f22318f5
  https://github.com/Cyan4973/xxHash/pull/433/commits/cf1988f2faa82bb99a44a998e67b0d74b93b0d24

And apply them on HEAD (227fd610778a82c9eab9674420d1704b4574b14a), then the tests pass again.

I have also observed rsync failing to transfer files to Power machines when using xxh128, if I rebuild libxxhash.so using the above patches and LD_PRELOAD it then rsync works again.

Should I submit a pull request with those VSX fixes rebased onto HEAD? I could test it on Power8/Power9 big and little endian, with various compilers.

mpe:
This rebases some VSX fixes from @easyaspi314 onto mainline, which fixes test failures I see on Power when building with GCC 10.

Also adds a GCC 10 build for ppc64le to CI, and adds GCC 9 & 10 builds for ppc64 (big endian) to CI.

Closes #596.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -255,7 +255,7 @@ jobs:
         CFLAGS=-Werror make
 
 
-  # Linux, { ARM, ARM64, PPC64LE, S390X }
+  # Linux, { ARM, ARM64, PPC64LE, PPC64, S390X }
   # All tests are using QEMU and gcc cross compiler.
 
   qemu-consistency:
@@ -269,6 +269,8 @@ jobs:
           { name: ARM64,    xcc_pkg: gcc-aarch64-linux-gnu,     xcc: aarch64-linux-gnu-gcc,     xemu_pkg: qemu-system-arm,    xemu: qemu-aarch64-static },
           { name: PPC64LE-gcc9,  xcc_pkg: gcc-9-powerpc64le-linux-gnu,  xcc: powerpc64le-linux-gnu-gcc-9,  xemu_pkg: qemu-system-ppc, xemu: qemu-ppc64le-static },
           { name: PPC64LE-gcc10, xcc_pkg: gcc-10-powerpc64le-linux-gnu, xcc: powerpc64le-linux-gnu-gcc-10, xemu_pkg: qemu-system-ppc, xemu: qemu-ppc64le-static },
+          { name: PPC64-gcc9,    xcc_pkg: gcc-9-powerpc64-linux-gnu,    xcc: powerpc64-linux-gnu-gcc-9,    xemu_pkg: qemu-system-ppc, xemu: qemu-ppc64-static },
+          { name: PPC64-gcc10,   xcc_pkg: gcc-10-powerpc64-linux-gnu,   xcc: powerpc64-linux-gnu-gcc-10,   xemu_pkg: qemu-system-ppc, xemu: qemu-ppc64-static },
           { name: S390X,    xcc_pkg: gcc-s390x-linux-gnu,       xcc: s390x-linux-gnu-gcc,       xemu_pkg: qemu-system-s390x,  xemu: qemu-s390x-static   },
           { name: MIPS,     xcc_pkg: gcc-mips-linux-gnu,        xcc: mips-linux-gnu-gcc,        xemu_pkg: qemu-system-mips,   xemu: qemu-mips-static    },
         ]
@@ -304,7 +306,7 @@ jobs:
         CPPFLAGS="-DXXH_VECTOR=XXH_NEON" LDFLAGS="-static" CC=$XCC RUN_ENV=$XEMU make clean check
 
     - name: ${{ matrix.name }} (XXH_VECTOR=[ scalar, VSX ])
-      if: ${{ startsWith(matrix.name, 'PPC64LE') }}
+      if: ${{ startsWith(matrix.name, 'PPC64') }}
       run: |
         CPPFLAGS="-DXXH_VECTOR=XXH_SCALAR" LDFLAGS="-static" CC=$XCC RUN_ENV=$XEMU make clean check
         CPPFLAGS="-DXXH_VECTOR=XXH_VSX" CFLAGS="-O3 -maltivec -mvsx -mpower8-vector -mcpu=power8" LDFLAGS="-static" CC=$XCC RUN_ENV=$XEMU make clean check
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -267,7 +267,8 @@ jobs:
         include: [
           { name: ARM,      xcc_pkg: gcc-arm-linux-gnueabi,     xcc: arm-linux-gnueabi-gcc,     xemu_pkg: qemu-system-arm,    xemu: qemu-arm-static     },
           { name: ARM64,    xcc_pkg: gcc-aarch64-linux-gnu,     xcc: aarch64-linux-gnu-gcc,     xemu_pkg: qemu-system-arm,    xemu: qemu-aarch64-static },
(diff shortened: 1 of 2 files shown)


t-mat:
I believe changesets for CI are very good.  I'm gladly watching the following line and tests 😃 

https://github.com/Cyan4973/xxHash/runs/4211012053?check_suite_focus=true#step:7:40

```
compiled as 64-bit ppc64 big endian with GCC 10.3.0 
```
----

I think the following my minor concen is out of the scope of this PR.  But put it here for future improvement.
My minor concern is actual performance on real PPC64/LE (and S390X) machines with GCC 10.  Because sometimes compiler generates (extremely) inefficient code with inline SIMD assembly/intrinsics.
It was imperfect but we had been able to check it with Travis CI.  Now since we're relying GitHub Actions, we can't verify it contiguously.

Cyan4973 (approved):
Thanks @mpe  ! This fix looks good to me

Cyan4973:
> My minor concern is actual performance on real PPC64/LE (and S390X) machines with GCC 10. Because sometimes compiler generates (extremely) inefficient code with inline SIMD assembly/intrinsics.

We might be able to check that manually using [godbolt.org](https://godbolt.org/)

Cyan4973:
So, I've been trying to look at the new code using godbolt.org,
but it appears that the list of compilers available for PPC64
is older than the list of compilers tested with Github Actions.

And as a consequence, it seems that the new code doesn't behave as well on these older compilers.
`gcc` for `ppc64` starts at v4.8.5.

Should we extend the nb of compilers tested for `PPC64` architecture ?

What about `clang` by the way ?

easyaspi314:
As for clang you should be able to do `--target`:
- `--target=powerpc-linux-gnu` -> ppc32
- `--target=powerpc64-linux-gnu` -> ppc64 big endian
- `--target=powerpc64le-linux-gnu` -> ppc64 little endian

mpe:
You're right, performance does seem to take a hit.

This is the results of `make test` before and after this PR, both built with GCC 9.3.0 and run on Power9:

| Test                             | Before | After  | Delta % |
|----------------------------------|--------|--------|---------|
|  1#XXH32                         |  37318 |  40797 |   9.32% |
|  2#XXH32 unaligned               |  41429 |  41376 |  -0.13% |
|  3#XXH64                         | 101396 | 101289 |  -0.11% |
|  4#XXH64 unaligned               |  98794 |  98708 |  -0.09% |
|  5#XXH3_64b                      |  72231 |  73159 |   1.28% |
|  6#XXH3_64b unaligned            |  69206 |  67193 |  -2.91% |
|  7#XXH3_64b w/seed               |  70111 |  76115 |   8.56% |
|  8#XXH3_64b w/seed unaligned     |  66248 |  70019 |   5.69% |
|  9#XXH3_64b w/secret             |  74936 |  66033 | -11.88% |
| 10#XXH3_64b w/secret unaligned   |  70375 |  60974 | -13.36% |
| 11#XXH128                        |  72287 |  73152 |   1.20% |
| 12#XXH128 unaligned              |  69325 |  67159 |  -3.12% |
| 13#XXH128 w/seed                 |  70119 |  76087 |   8.51% |
| 14#XXH128 w/seed unaligned       |  66294 |  70009 |   5.60% |
| 15#XXH128 w/secret               |  74817 |  66003 | -11.78% |
| 16#XXH128 w/secret unaligned     |  70272 |  60954 | -13.26% |
| 17#XXH3_stream                   |  57987 |  47869 | -17.45% |
| 18#XXH3_stream unaligned         |  53987 |  44993 | -16.66% |
| 19#XXH3_stream w/seed            |  58033 |  47836 | -17.57% |
| 20#XXH3_stream w/seed unaligned  |  53975 |  44920 | -16.78% |
| 21#XXH128_stream                 |  57992 |  47854 | -17.48% |
| 22#XXH128_stream unaligned       |  53961 |  44962 | -16.68% |
| 23#XXH128_stream w/seed          |  58009 |  47836 | -17.54% |
| 24#XXH128_stream w/seed unaligne |  53962 |  44908 | -16.78% |


I'll see if I can narrow down what exactly is causing the slow down.

easyaspi314:
What happens if you revert this, and just change
```c
typedef __vector unsigned long long xxh_u64x2;
typedef __vector unsigned char xxh_u8x16;
typedef __vector unsigned xxh_u32x4;
```
To
```c
typedef __vector unsigned long long xxh_u64x2 __attribute__((__may_alias__));
typedef __vector unsigned char xxh_u8x16 __attribute__((__may_alias__));
typedef __vector unsigned xxh_u32x4 __attribute__((__may_alias__));
```
Does that fix the aliasing issue without affecting the performance?

I'll have to dig my ssh keys out though to properly test it.

mpe:
> Does that fix the aliasing issue without affecting the performance?

It doesn't fix the aliasing issue. I see the same test failure as before:

```
$ make test
gcc-10 -O3 -Wall -Wextra -Wconversion -Wcast-qual -Wcast-align -Wshadow -Wstrict-aliasing=1 -Wswitch-enum -Wdeclaration-after-statement -Wstrict-prototypes -Wundef -Wpointer-arith -Wformat-security -Wvla -Wformat=2 -Winit-self -Wfloat-equal -Wwrite-strings -Wredundant-decls -Wstrict-overflow=2 -DXXH_DEBUGLEVEL=1  -DXXH_INLINE_ALL cli/xxhsum.c -o xxhsum_inlinedXXH
...
# internal bench
./xxhsum -bi0
xxhsum 0.8.1 by Yann Collet
compiled as 64-bit ppc64 + POWER8 vector little endian with GCC 10.3.0
Error: 64-bit hash test 100: Internal sanity check failed!
Got 0xBB9D481D33CB87AAULL, expected 0xCDEB804D65C6DEA4ULL.
or temporarily recompile with XSUM_NO_TESTS=1.
make: *** [Makefile:196: check] Error 1
```

And adding `-fno-strict-aliasing` "fixes" it like before.

mpe:
Looks like to get the tests passing with GCC 10 I only need the `xacc` change in `XXH3_accumulate_512_vsx()`.

But that also looks to account for nearly all of the slow down.

Cyan4973:
Trying to get a sense of the situation :
is this PR considered ready for merge ?
or does it need another round of test / optimization ?

Cyan4973:
I'm not sure what's going on.
This PR seems to fix issues with `gcc-10`,
but there are still issues with `gcc-9` : 
https://github.com/Cyan4973/xxHash/runs/4343083642?check_suite_focus=true

mpe:
I asked a GCC developer for some hints and they replied that the code was not valid C due to the strict aliasing violations (accessing an object of type A as type B), they suggested using memcpy or bit_cast (?).

Given the tests pass with the change I think it's an improvement, even if the generated code is slower.

PR #598, opened 2021-11-23
[NEON] Enable XXH_NEON for MSVC ARM/ARM64

easyaspi314:
When MSVC is targeting ARMv7VE or ARM64, it will now use the NEON path.

It actually just works(TM) out of the box.

The code hasn't been tested, as I don't have access to an ARM64 Windows device, but the code appears to be correct judging from the assembly listings. I may be able to get WINE on ARM64 working though.

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -2848,10 +2848,13 @@ enum XXH_VECTOR_TYPE /* fake enum */ {
 #    define XXH_VECTOR XXH_AVX2
 #  elif defined(__SSE2__) || defined(_M_AMD64) || defined(_M_X64) || (defined(_M_IX86_FP) && (_M_IX86_FP == 2))
 #    define XXH_VECTOR XXH_SSE2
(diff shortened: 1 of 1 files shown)


Cyan4973:
Windows on `arm64`, that's indeed a difficult target to test in CI...

Cyan4973 (approved):

easyaspi314:
I can verify that armv7ve binary works (but it only runs at about the speed of XXH32), I managed to get Wine on my Pi4B, but the ancient version of wine64 provided in the Raspbian repos crashed before it would do anything, so I'm installing a different OS.

I'd love to install the custom Win10 IoT build but it is the 2GB model 😂

easyaspi314:
Yeah arm64 wine isn't going to work. Version 5.6 crashes with something about `__C_specific_handler`. I'm not going to worry too much about it (although I'm pretty frustrated that I had to reinstall the system like 5 times, including an in place Ubuntu upgrade which took two hours, all for naught)

PR #601, opened 2021-11-25
fixed -Wdocumentation

### LINKED ISSUE #590: Wdocumentation error during grpc ios compile
the logs are as follow:

../../../../../../third_party/xxhash/xxhash.h:1916:11: error: parameter 'input,' not found in the function declaration [-Werror,-Wdocumentation]
 * @param input, len, seed Directly passed from @ref XXH32().
          ^~~~~~
../../../../../../third_party/xxhash/xxhash.h:1916:11: note: did you mean 'input'?
 * @param input, len, seed Directly passed from @ref XXH32().
          ^~~~~~
          input
../../../../../../third_party/xxhash/xxhash.h:3104:11: error: parameter 'lhs,' not found in the function declaration [-Werror,-Wdocumentation]
 * @param lhs, rhs The 64-bit integers to be multiplied
          ^~~~
../../../../../../third_party/xxhash/xxhash.h:3104:11: note: did you mean 'lhs'?
 * @param lhs, rhs The 64-bit integers to be multiplied
          ^~~~
          lhs
../../../../../../third_party/xxhash/xxhash.h:3222:11: error: parameter 'lhs,' not found in the function declaration [-Werror,-Wdocumentation]
 * @param lhs, rhs The 64-bit integers to multiply
          ^~~~
../../../../../../third_party/xxhash/xxhash.h:3222:11: note: did you mean 'lhs'?
 * @param lhs, rhs The 64-bit integers to multiply
          ^~~~
          lhs
3 errors generated.

Cyan4973:
fix #590,
thanks to @t-mat for providing the solution !

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -334,7 +334,7 @@ jobs:
 
     - name: make
       run: |
-        CFLAGS="-Werror" make clean default
+        CFLAGS="-Werror -Wdocumentation" make clean default
 
     - name: make test
       run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -334,7 +334,7 @@ jobs:
 
     - name: make
       run: |
-        CFLAGS="-Werror -Wdocumentation" make clean default
+        CFLAGS="-Werror" make clean default
 
     - name: make test
       run: |
--- a/xxh_x86dispatch.c
+++ b/xxh_x86dispatch.c
@@ -229,7 +229,7 @@ extern "C" {
  * @internal
  * @brief Runs CPUID.
  *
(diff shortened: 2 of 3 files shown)


Cyan4973:
Damn, why isn't my local `macos` system generating the same error ?
```
./xxhash.h:505:11: error: empty paragraph passed to '@return' command [-Werror,-Wdocumentation]
 * @return @ref XXH_OK on success, @ref XXH_ERROR on failure.
   ~~~~~~~^
```

`clang` versions are _slightly_ different : 
`Apple clang version 12.0.0 (clang-1200.0.32.29)` vs `Apple clang version 12.0.5 (clang-1205.0.22.9)`.
So that's very minor.
But if that's indeed the reason for the difference, it makes this test rather fragile, frivolously changing its outcome at every minor version variations, effectively not usable for CI tests.

PR #608, opened 2021-11-27
Fix XXH3 streaming performance

### LINKED ISSUE #516: Streaming much slower than XXH3_64b for AVX 2 on MSVC
When running "xxhsum -b" I get the following results for SSE2:

xxh-msvc 0.8.0 by Yann Collet
compiled as 64-bit x86_64 + SSE2 little endian with MSVC 19.28.29336.00
Sample of 100 KB...
 5#XXH3_64b                      :     102400 ->   265217 it/s (25900.1 MB/s)
17#XXH3_stream                   :     102400 ->   224415 it/s (21915.5 MB/s)

With AVX2 the performance for non-streaming greatly increases, but streaming does not:

xxh-msvc 0.8.0 by Yann Collet
compiled as 64-bit x86_64 + AVX2 little endian with MSVC 19.28.29336.00
Sample of 100 KB...
 5#XXH3_64b                      :     102400 ->   613544 it/s (59916.4 MB/s)
17#XXH3_stream                   :     102400 ->   219870 it/s (21471.7 MB/s)

Is this expected or am I doing something wrong (compiler switches, ...)?

Tested for master and v0.8.0
CPU: Ryzen 3900x
Compiler switches:
/permissive- /ifcOutput "x64\Release\" /GS /GL /W3 /Gy /Zc:wchar_t /Zi /Gm- /O2 /sdl /Fd"x64\Release\vc142.pdb" /Zc:inline /fp:precise /D "NDEBUG" /D "_CONSOLE" /D "_UNICODE" /D "UNICODE" /errorReport:prompt /WX- /Zc:forScope /arch:AVX2 /Gd /Oi /MD /FC /Fa"x64\Release\" /EHsc /nologo /Fo"x64\Release\" /Fp"x64\Release\xxh-msvc.pch" /diagnostics:column

Cyan4973:
Fix #516 

XXH3 streaming was significantly slower than one-shot hashing, especially for non-`clang` compilers, such as `gcc` and `MSVC`.
This PR largely fixes the performance difference for large ingestion scenarios.

test : `xxhsum -b21` (`XXH3_stream`)

| platform | compiler | **dev** | **this PR** |
| --- | --- | --- | --- |
| macos laptop | Apple Clang 12.0.5 | 25.9 GB/s | 28.3 GB/s |
| macos laptop | GCC 11.2.0 | 18.8 GB/s | 27.7 GB/s |
| WinVM laptop | Clang 13.0.0 | 23.8 GB/s | 26.2 GB/s |
| WinVM laptop | GCC 11.2.0 | 17.4 GB/s | 25.6 GB/s |
| WinVM laptop | MSVC 19.29 | 14.3 GB/s | 23.2 GB/s |
| Ubu20 desktop | GCC 9.3.0 | 26.1 GB/s | 38.2 GB/s |
| Ubu20 desktop | Clang 10.0.0 | 31.1 GB/s | 33.8 GB/s |
| Win10 desktop | GCC 11.2.0 | 26.8 GB/s | 38.8 GB/s |
| Win10 desktop | Clang 13.0.0 | 37.2 GB/s | 39.8 GB/s |
| Win10 desktop | MSVC 19.16 | 21.5 GB/s | 32.2 GB/s |

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -4711,7 +4711,7 @@ XXH3_update(XXH3_state_t* XXH_RESTRICT const state,
          * when operating accumulators directly into state.
          * Operating into stack space seems to enable proper optimization.
          * clang, on the other hand, doesn't seem to need this trick */
-        xxh_u64 acc[8]; memcpy(acc, state->acc, sizeof(acc));
+        XXH_ALIGN(XXH_ACC_ALIGN) xxh_u64 acc[8]; memcpy(acc, state->acc, sizeof(acc));
 #else
         xxh_u64* XXH_RESTRICT const acc = state->acc;
 #endif
--- a/xxhash.h
+++ b/xxhash.h
@@ -4679,6 +4679,11 @@ XXH3_consumeStripes(xxh_u64* XXH_RESTRICT acc,
     }
 }
 
+#ifndef XXH3_STREAM_USE_STACK
+# ifndef __clang__ /* clang doesn't need additional stack space */
+#   define XXH3_STREAM_USE_STACK 1
+# endif
+#endif
 /*
  * Both XXH3_64bits_update and XXH3_128bits_update use this routine.
  */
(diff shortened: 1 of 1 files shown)


Cyan4973:
Strange,
why would this update fail a sanity test
specifically on `PPC64LE` + `Power8 Vector`
and no other tested platform ?

_edit_ : well, trying to reproduce the issue on a linux VM, this test seems to fail not just this PR but also much older commits, including `v0.8.0`.
But the question now becomes : why would it fail on GA only _starting now_ ?
Even this prior PR (#606) has been running and tested fine, and it was just yesterday ??

_note_ : forcing the binary to use the `scalar` code path makes it work correctly. So it seems the issue comes from the `vsx` vector extension ?

_edit 2_ : merging #597 seems to fix it.
So it seems it was a long-standing issue within the `vsx` code path ?
Then again, why has it started failing in GA CI only _now_ ?

_edit 3_ : ok, great, it seems this is now passing `powerpc64le` + `vsx` + `gcc-10` test, 
but failing `powerpc64le` + `vsx` + `gcc-9` one.

_edit 4_ : so, apparently, the `vsx` code path is really sensible to accumulators' alignment, so that should fix it.
Not sure why older commits have started to fail too, but well, since #597 fixes that too ...

easyaspi314:
> edit 4 : so, apparently, the vsx code path is really sensible to accumulators' alignment, so that should fix it.
> Not sure why older commits have started to fail too, but well, since #597 fixes that too ...

GCC for the undermaintained architectures is an untamed beast ¯\\\_(ツ)\_/¯

Issue #620, opened 2021-11-29
Windows ARM build bustage with xxHash 0.8.1


rvandermeulen:
Hi, I'm trying to update Firefox' copy of xxHash from version 0.8.0 to 0.8.1 and our Windows ARM builds are failing with the below error and others along the same line:
`INFO -  /builds/worker/checkouts/gecko/mfbt/lz4/xxhash.h(4039,9): error: unknown type name 'uint64x2_t'; did you mean 'uint64_t'?`

Looks like it's missing the arm_neon.h include needed for that type?

rvandermeulen:
Looks like a regression from #598 (@easyaspi314). I can confirm that the below patch fixes the issue without causing issues for ARM builds on other platforms:
```
diff --git a/mfbt/lz4/xxhash.h b/mfbt/lz4/xxhash.h
--- a/mfbt/lz4/xxhash.h
+++ b/mfbt/lz4/xxhash.h
@@ -4024,16 +4024,18 @@ XXH_FORCE_INLINE XXH_TARGET_SSE2 void XX
             dst16[i] = _mm_add_epi64(_mm_load_si128((const __m128i *)src16+i), seed);
     }   }
 }
 
 #endif
 
 #if (XXH_VECTOR == XXH_NEON)
 
+#include <arm_neon.h>
+
 XXH_FORCE_INLINE void
 XXH3_accumulate_512_neon( void* XXH_RESTRICT acc,
                     const void* XXH_RESTRICT input,
                     const void* XXH_RESTRICT secret)
 {
     XXH_ASSERT((((size_t)acc) & 15) == 0);
     {
         uint64x2_t* const xacc = (uint64x2_t *) acc;
```

easyaspi314:
[Try this](https://github.com/easyaspi314/xxHash/tree/msvc_arm_hotfix)

It should also detect ARM64EC on VS2022 preview and fix `ARMv__ARM_ARCH` in the cli for 32-bit builds.

rvandermeulen:
AFAIK, arm_neon.h detects and includes arm64_neon.h if needed (and arm64_neon.h doesn't include all the definitions that arm_neon.h does), so I think you can unconditionally just use arm_neon.h.

easyaspi314:
This doesn't seem to be the case for some earlier versions of MSVC 2017, where it would error saying it was only for ARM targets. Microsoft always has to keep it interesting. 😅

rvandermeulen:
[...]

PR #621, opened 2021-11-30
[don't merge] Msvc ARM hotfix

easyaspi314:
~~Hotfix for #620. This also detects the ARM64EC target on MSVC 2022 Preview and the output for xxhsum -b on armv7ve.~~ Invalid right now. Had some git issues and apparently it doesn't fix the issue.


easyaspi314:
huh? Looks like I forgot to git pull...

Cyan4973:
Related : we may have to expand `CI` tests to cover the case that was detected in #620

easyaspi314:
Yes, some more clang-cl tests would be welcome.

PR #623, opened 2021-11-30
Fix some ARM/clang-cl feature detection issues

### LINKED ISSUE #620: Windows ARM build bustage with xxHash 0.8.1

easyaspi314:
- Fix #620 where clang-cl did not include `<arm_neon.h>`
- Use `__uint128_t` for clang-cl
- Use NEON when `_M_ARM` is defined, not only `_M_ARM_ARMV7VE`
- Use ARM64 codepaths for the new ARM64EC ABI in VS2022 Preview (which emulates the x64 ABI and is detected as so)
  - TODO: don't dispatch
- Update CLI detection to detect Windows ARM better (although you have to cheat to actually run the program on Windows RT lol, only wine users will see it)

This should probably be merged into a hotfix for v0.8.1.

### DIFF
--- a/cli/xsum_arch.h
+++ b/cli/xsum_arch.h
@@ -79,7 +79,7 @@
 #endif
 
 /* makes the next part easier */
-#if defined(__x86_64__) || defined(_M_AMD64) || defined(_M_X64)
+#if (defined(__x86_64__) || defined(_M_AMD64) || defined(_M_X64)) && !defined(_M_ARM64EC)
 #   define XSUM_ARCH_X64 1
 #   define XSUM_ARCH_X86 "x86_64"
 #elif defined(__i386__) || defined(_M_IX86) || defined(_M_IX86_FP)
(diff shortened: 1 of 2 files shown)


Cyan4973 (approved):
The fix itself is good.

I'm just wondering if we should pair it with some kind of extended test coverage, triggering the problem that's been fixed in this PR.

t-mat:
> I'm just wondering if we should pair it with some kind of extended test coverage, triggering the problem that's been fixed in this PR.

#637 may mitigate your concern.  But since the root cause of #620 is related to specific version of MSVC (and its header), #637 doesn't reproduce the issue.

As for MSVC 2019 and later, we can download specific version of it (as a buildtools).  Though it takes some amount of time for downloading.
But MSVC 2017 provides a few (incomplete) versions of their releases.
Perhaps they provides complete list in the Azure archive.  But I don't know anything about it.

PR #628, opened 2021-11-30
Fix an assert comparison the same values (flagged by PVS Studio in 0.8.1)

### LINKED ISSUE #625: 0.8.1: PVS Studio flagged buffer underflows in the state reset functions
My guess is it's detecting the use of the reserved portion, which if coming from malloc (for the memory behind the incoming state pointer), might contain garbage.

### LINKED ISSUE #626: 0.8.1: PVS Studio flagged a copy and paste error asserting the same values

### LINKED ISSUE #627: 0.8.1: PVS Studio flagged a potential micro-optimization using a const-reference argument

kcgen:
- Fixes #626

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -4843,7 +4843,7 @@ XXH3_update(XXH3_state_t* XXH_RESTRICT const state,
             XXH_ASSERT(state->nbStripesPerBlock >= state->nbStripesSoFar);
             /* join to current block's end */
             {   size_t const nbStripesToEnd = state->nbStripesPerBlock - state->nbStripesSoFar;
-                XXH_ASSERT(nbStripes <= nbStripes);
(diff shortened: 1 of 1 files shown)


Cyan4973 (changes requested):
Only the second issue `nbStripesToEnd` is fine in this PR.

I'll proceed differently for the other 2 issues.

kcgen:
> Only the second issue `nbStripesToEnd` is fine in this PR.

Thanks for the quick review. I've now dropped the other two commits. 
 
> I'll proceed differently for the other 2 issues.

Thanks for addressing these @Cyan4973.
Happy to re-run more PVS scans to help wrap these up.

kcgen:
> I'll proceed differently for the other 2 issues.

Dropped the other two commits.

Cyan4973 (approved):

PR #635, opened 2021-12-01
Add macOS matrix to ci.yml

t-mat:
This PR introduces matrix for multiple macOS systems.

For difference between `macos-10.15` and `macos-11`, see
https://github.com/actions/virtual-environments/issues/4060

Here's [actual build log on `macos-11`](https://github.com/t-mat/xxHash/runs/4376263530?check_suite_focus=true#step:3:6)


Cyan4973 (approved):
Nice one !

PR #638, opened 2021-12-01
fix #625

### LINKED ISSUE #625: 0.8.1: PVS Studio flagged buffer underflows in the state reset functions

Cyan4973:
PVS Studio static analyzer doesn't like that the structure is not entirely initialized.
This was intentional, and supposed to allow, in some future, to reduce the size of the state,
while minimizing impact with library's installed based (in case of version mismatch).

This is way too complex and risky for such a tiny benefit.
So revert to the classical "init everything" method, which should please PVS Studio.
The last member will remain "reserved" and unused.
Who knows, maybe in some future, it may be of some use, for some purpose yet to determine.

Fix #625, reported by @kcgen .

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -978,7 +978,7 @@ struct XXH32_state_s {
    XXH32_hash_t v[4];         /*!< Accumulator lanes */
    XXH32_hash_t mem32[4];     /*!< Internal buffer for partial reads. Treated as unsigned char[16]. */
    XXH32_hash_t memsize;      /*!< Amount of data in @ref mem32 */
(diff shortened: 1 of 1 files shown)


kcgen:
Rescan in progress.

kcgen:
Confirmed good! [pvs-analysis-report.zip](https://github.com/dosbox-staging/dosbox-staging/suites/4535503456/artifacts/121527352)

PR #639, opened 2021-12-01
fix #627

### LINKED ISSUE #627: 0.8.1: PVS Studio flagged a potential micro-optimization using a const-reference argument

Cyan4973:
attempt to fix a minor PVS Studio warning
which claims that passing a value by copy is worse for performance than a value by reference.
Given that the value is 16 bytes long, the claim is quite debatable, if not already silly from the start,
but it is even more so in this case because the function is very short and designed to be inlined by the compiler anyway.

Made the inlining objective more explicit, in an attempt to silence the static analyzer objection.

This is a blind fix, as I don't have PVS Studio to check.

fix #627 (hopefully)

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -5572,7 +5572,7 @@ XXH128_hashFromCanonical(const XXH128_canonical_t* src)
  */
 #define XXH_MIN(x, y) (((x) > (y)) ? (y) : (x))
 
-static void XXH3_combine16(void* dst, XXH128_hash_t h128)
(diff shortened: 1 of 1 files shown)


kcgen:
Rescan in progress.

kcgen:
Confirmed good! [pvs-analysis-report.zip](https://github.com/dosbox-staging/dosbox-staging/suites/4535503456/artifacts/121527352)

PR #640, opened 2021-12-01
Add QEMU/gcc matrix to ci.yml

### LINKED ISSUE #624: CI: Add test matrix of gcc versions and QEMU architectures
https://github.com/Cyan4973/xxHash/issues/622#issuecomment-982237636
> should we test PPC64LE + gcc v4.8.5 in CI ? and is that even possible btw ?

- It seems we can easily install `gcc-[5,11]` cross compiler via `apt-get`.
- I don't know much about `gcc-4.8` cross compiler so far.
- We should watch total time of tests.

t-mat:
This PR fixes #624 and adds QEMU test with gcc-{6,7,8,9,10} cross compiler for { arm, aarch64, ppc64, ppc64le, s390x, mips }.

## Implementation
For further tweaking, I leave the matrix as just a bulky text table.
We'll be able to refactor it after we'll implement stable test.

## Known issue
For some reason, ubuntu-20.04 fails to retrieve `gcc-8-mips-linux-gnu`.
As a workaround, we use ubuntu-18.04 for it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -260,32 +260,73 @@ jobs:
 
   qemu-consistency:
     name: QEMU ${{ matrix.name }}
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false  # 'false' means Don't stop matrix workflows even if some matrix failed.
       matrix:
         include: [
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):

Cyan4973:
Excellent ! Thanks @t-mat !

PR #650, opened 2021-12-07
Use __attribute__((aligned)) instead of packed

Hello71:
It improves gcc code generation on armv6 and armv7:
https://gcc.godbolt.org/z/nfnEKsjzc. It has some minor effect on
aarch64: +4 instructions with -O2, -156 instructions with -O3. Overall
it is clear that while gcc emits proper instructions in isolation with
packed structures, it often gets confused after inlining, and using
explicit alignment value with integer mode fixes this issue.

See also https://github.com/facebook/zstd/pull/2881.

I have not runtime tested this change, since I don't have suitable ARM testing environment. I checked that it produces identical code on x86_64.

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -1292,7 +1292,7 @@ XXH3_128bits_reset_withSecretandSeed(XXH3_state_t* statePtr,
  *     Use `memcpy()`. Safe and portable. Note that most modern compilers will
  *     eliminate the function call and treat it as an unaligned access.
  *
- *  - `XXH_FORCE_MEMORY_ACCESS=1`: `__attribute__((packed))`
+ *  - `XXH_FORCE_MEMORY_ACCESS=1`: `__attribute__((aligned(1)))`
  *   @par
  *     Depends on compiler extensions and is therefore not portable.
  *     This method is safe _if_ your compiler supports it,
@@ -1403,20 +1403,7 @@ XXH3_128bits_reset_withSecretandSeed(XXH3_state_t* statePtr,
 
 #ifndef XXH_FORCE_MEMORY_ACCESS   /* can be defined externally, on command line for example */
    /* prefer __packed__ structures (method 1) for gcc on armv7+ and mips */
-#  if !defined(__clang__) && \
-( \
-    (defined(__INTEL_COMPILER) && !defined(_WIN32)) || \
-    ( \
-        defined(__GNUC__) && ( \
-            (defined(__ARM_ARCH) && __ARM_ARCH >= 7) || \
-            ( \
-                defined(__mips__) && \
-                (__mips <= 5 || __mips_isa_rev < 6) && \
-                (!defined(__mips16) || defined(__mips_mips16e2)) \
-            ) \
-        ) \
-    ) \
-)
+#  ifdef __GNUC__
 #    define XXH_FORCE_MEMORY_ACCESS 1
 #  endif
 #endif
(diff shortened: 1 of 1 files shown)


Cyan4973:
cc @easyaspi314

xxhash.h
```diff
@@ -1403,20 +1403,7 @@ XXH3_128bits_reset_withSecretandSeed(XXH3_state_t* statePtr,
...
-            (defined(__ARM_ARCH) && __ARM_ARCH >= 7) || \
-            ( \
-                defined(__mips__) && \
-                (__mips <= 5 || __mips_isa_rev < 6) && \
-                (!defined(__mips16) || defined(__mips_mips16e2)) \
-            ) \
-        ) \
-    ) \
-)
+#  ifdef __GNUC__
```
Cyan4973, line 1406: This will impact both `gcc` and `clang`. 
Previously, this mode was only activated for `gcc`.
Is it an intentional objective ? 
Is this mode desirable only for `gcc`, or for both `gcc` and `clang` even though `clang` seems to work fine with `memcpy()` (method 0)?

easyaspi314:
Was about to say "but `aligned` can only *increase* alignment" but then I read the comment XD

xxhash.h
Hello71: afaik there is no effect for clang, it consistently treats memcpy, packed, and aligned(1) in the same way. icc before 18 is slightly worse with memcpy (it generates real load/store) but 18 is almost 5 years old already, and icc is "legacy" anyways. it is arguably "safer" to use it for gcc only, but this condition is simpler, and icc and clang behavior will almost certainly not change in the future (the former because it is no longer developed, and the latter for backwards compatibility). i don't object to changing it to something else though.

Cyan4973:
It seems that `__attribute__((aligned(1)))` wasn't meant to decrease alignment,
but in practice, it does, and apparently, it now does it officially (documented).
It also seems to result in some performance improvement, at least on ARM platforms, 
so, well, okay.

My main concern here is to make sure that it combines well with #651 , hence the cc @easyaspi314 .

Cyan4973 (approved):

easyaspi314:
There do not seem to be any conflicts with various ARM options. GCC seems to emit byteshift when it needs byteshift and normal unaligned loads when it needs unaligned loads. It also fixes actual calls to `memcpy` on ARM strict alignment mode.

Clang sees almost all of these as the same so I would go with the always portable `memcpy`.

For the reference, here are GCC and Clang's IR representations
<details>

Clang -emit-llvm
```llvm
; memcpy
define i32 @read_memcpy(i8* nocapture readonly %0) local_unnamed_addr #0 {
  %2 = bitcast i8* %0 to i32*
  %3 = load i32, i32* %2, align 1
  ret i32 %3
}
; union { uint32_t x; } __attribute__((packed))
define i32 @read_packed_union(i8* nocapture readonly %0)  {
  %2 = bitcast i8* %0 to i32*
  %3 = load i32, i32* %2, align 1
  ret i32 %3
}
; uint32_t __attribute__((aligned(1)))
define i32 @read_aligned_attr(i8* nocapture readonly %0) {
  %2 = bitcast i8* %0 to i32*
  %3 = load i32, i32* %2, align 1
  ret i32 %3
}
; byteshift (clang does not recognize this as an unaligned load)
define i32 @byteshift(i8* nocapture readonly %0) {
  %2 = load i8, i8* %0, align 1
  %3 = zext i8 %2 to i32
  %4 = getelementptr inbounds i8, i8* %0, i32 1
  %5 = load i8, i8* %4, align 1
  %6 = zext i8 %5 to i32
  %7 = shl nuw nsw i32 %6, 8
  %8 = or i32 %7, %3
  %9 = getelementptr inbounds i8, i8* %0, i32 2
  %10 = load i8, i8* %9, align 1
  %11 = zext i8 %10 to i32
  %12 = shl nuw nsw i32 %11, 16
  %13 = or i32 %8, %12
  %14 = getelementptr inbounds i8, i8* %0, i32 3
  %15 = load i8, i8* %14, align 1
  %16 = zext i8 %15 to i32
  %17 = shl nuw i32 %16, 24
  %18 = or i32 %13, %17
  ret i32 %18
}
; raw pointer cast (the only unsafe one)
define i32 @raw_cast(i8* nocapture readonly %0) {
  %2 = bitcast i8* %0 to i32*
  %3 = load i32, i32* %2, align 4 ; <<< align 4 == bad
  ret i32 %3
}
```
gcc -fdump-tree-optimized
```c
read_memcpy (const uint8_t * ptr)
{
  uint32_t x;
  uint32_t _4;

  memcpy (&x, ptr_2(D), 4);
  _4 = x;
  x ={v} {CLOBBER};
  return _4;
}

read_packed_union (const uint8_t * ptr)
{
  uint32_t _3;

  _3 = MEM[(const union unalign *)ptr_2(D)].x;
  return _3;
}

read_aligned_attr (const uint8_t * ptr)
{
  uint32_t _3;

  _3 = MEM[(const unalign *)ptr_2(D)];
  return _3;
}

byteshift (const uint8_t * ptr)
{
  uint32_t _16;

  _16 = MEM <unsigned int> [(const uint8_t *)ptr_15(D)];
  return _16;

}

raw (const uint8_t * ptr)
{
  uint32_t _3;

  _3 = MEM[(const uint32_t *)ptr_2(D)];
  return _3;
}
```
</details>

Cyan4973:
> Clang sees almost all of these as the same so I would go with the always portable `memcpy`.

If it comes to this we could always change [this line](https://github.com/Cyan4973/xxHash/pull/650/files#diff-343c0b024b7734da2547ae51a138702bd0bb2f2adc2bc7fb7eb97b6d543e9785R1406) to 
```
#if defined(__GNUC__) && !defined(__clang__)
```
though it does not seem absolutely necessary since codegen seems unaffected, hence I can merge this PR as is for the time being.

PR #654, opened 2021-12-08
new build macro XXH_NO_STDLIB

Cyan4973:
Useful for embedded / limited environments : 
ensures there is no usage / invocation of `<stdlib.h>` within `libxxhash`, notably no `malloc()` nor `free()`,
at the expense of `XXH*_createState()` functions, which now always return `NULL`.
This is only useful in situations where state is always allocated statically,
or only one-shot functions (like `XXH64()`) are used.


easyaspi314:
free may not have a leading underscore. That's only really the case on macOS.

PR #661, opened 2021-12-13
Documentation update

easyaspi314:
Here we go again.

I added a few code examples. It still needs some organization improvements and XXH3 is still a little barren, but I think that the doxygen can be added to xxhash.com now.

It also has the sleek, updated benchmark and the primary examples on the front page. (Excuse the weird missing lines on the table, that's just Firefox being jank)
Screenshot_20211212-215406


Cyan4973 (approved):

Cyan4973:
Great update @easyaspi314 !

PR #667, opened 2021-12-23
Introduce XXH_SIZE_OPT and XXH_NO_STREAM

easyaspi314:
- `XXH_SIZE_OPT` is a value from 0-2 which indicates how much xxHash should care about code size, default 1 for -Os/-Oz, default 0 otherwise. It chooses a set of defaults. 
   - `==1`: conservative size optimizations (basically the same as what we were already doing with -Oz)
   - `==2`: performance may cry
 - `XXH_NO_STREAM` disables the streaming API.
 - These two interact: if `XXH_SIZE_OPT == 2` and `XXH_NO_STREAM` is not defined, XXH32 and XXH64 use the streaming API for single shot
    - TODO: apply this to XXH3 as well

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -3512,7 +3512,9 @@ XXH_FORCE_INLINE xxh_u64x2 XXH_vec_mule(xxh_u32x4 a, xxh_u32x4 b)
 #if defined(XXH_NO_PREFETCH)
 #  define XXH_PREFETCH(ptr)  (void)(ptr)  /* disabled */
 #else
-#  if defined(_MSC_VER) && (defined(_M_X64) || defined(_M_IX86))  /* _mm_prefetch() not defined outside of x86/x64 */
+#  if XXH_SIZE_OPT >= 1
+#    define XXH_PREFETCH(ptr) (void)(ptr)
+#  elif defined(_MSC_VER) && (defined(_M_X64) || defined(_M_IX86))  /* _mm_prefetch() not defined outside of x86/x64 */
 #    include <mmintrin.h>   /* https://msdn.microsoft.com/fr-fr/library/84szxsww(v=vs.90).aspx */
 #    define XXH_PREFETCH(ptr)  _mm_prefetch((const char*)(ptr), _MM_HINT_T0)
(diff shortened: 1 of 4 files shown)


Cyan4973:
> if XXH_SIZE_OPT == 2 and XXH_NO_STREAM is not defined, XXH32 and XXH64 use the streaming API for single shot

How much binary size is saved by this optimization ?

> TODO: apply this to XXH3 as well

Do you want to do it as part of this PR ?

xxhash.h
```diff
@@ -1591,6 +1629,17 @@ XXH3_128bits_reset_withSecretandSeed(XXH3_state_t* statePtr,
  */
 #  define XXH_OLD_NAMES
 #  undef XXH_OLD_NAMES /* don't actually use, it is ugly. */
+
+/*!
+ * @def XXH_NO_STREAM
+ * @brief Disablea the streaming API.
```
Cyan4973: nit: Disable

easyaspi314:
I checked it. While Clang seems to do very slightly better doing streamed single shot XXH3, GCC emits larger code. However that might be AArch64 Clang's identical code outliner copying the stack setup.

So, not worth the 40-50 lines to implement it.

XXH3 has a lot more code reuse than XXH32/64 so it isn't too surprising that this is the case. (Which is why I expect some more improvement once I clean up the modern XXH32/XXH64 refactor). 

I also rerolled the midrange loops on >= 1.

Cyan4973 (approved):

Cyan4973:
While I believe that the objectives of the PR are fine,
I sometimes worry about the added maintenance complexity.

I see nothing "wrong" in this PR.
Just another set of build variables and `#ifdef` to maintain,
and while they only add a little bit of complexity,
there is a sense that little by little, as PR stack up, following the code is getting more difficult.

And unfortunately, I also have no simple counter proposal.
Sometimes, making the code more maintainable requires sweeping non-local changes for the sake of small complexity benefits.

So take it more as a "general feeling" when building future evolutions (not something specific to this PR).
I think code maintenance is a topic we must keep in mind and try to improve.

PR #670, opened 2021-12-26
do no longer depend on `<assert.h>` for XXH_STATIC_ASSERT

### LINKED ISSUE #664: xxhash.h->XXH_STATIC_ASSERT: potential build failure due to missing static_assert macro
I just reported this to lighttpd; they directed me to this project. 

Cross-compiling with a uClibc toolchain results in a build failure, because uClibc does not define the static_assert macro. 
Including the macro definition in case it is not defined by <assert.h> fixes this issue

```
/* note: use after variable declarations */
#ifndef XXH_STATIC_ASSERT
#  if defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L)    /* C11 */
#    include <assert.h>

/* uClibc does not define static_assert, but provides _Static_assert */
#    ifndef static_assert
#      define static_assert _Static_assert
#    endif

#    define XXH_STATIC_ASSERT_WITH_MESSAGE(c,m) do { static_assert((c),m); } while(0)
#  elif defined(__cplusplus) && (__cplusplus >= 201103L)            /* C++11 */
#    define XXH_STATIC_ASSERT_WITH_MESSAGE(c,m) do { static_assert((c),m); } while(0)
#  else
#    define XXH_STATIC_ASSERT_WITH_MESSAGE(c,m) do { struct xxh_sa { char x[(c) ? 1 : -1]; }; } while(0)
#  endif
#  define XXH_STATIC_ASSERT(c) XXH_STATIC_ASSERT_WITH_MESSAGE((c),#c)
#endif

```

Cyan4973:
since some versions of this library are buggy (ex: `uclibc`).

use `_Static_assert` instead, which is directly part of the `C11` language.

fix #664 .

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -1749,8 +1749,7 @@ static void* XXH_memcpy(void* dest, const void* src, size_t size)
 /* note: use after variable declarations */
 #ifndef XXH_STATIC_ASSERT
 #  if defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L)    /* C11 */
-#    include <assert.h>
-#    define XXH_STATIC_ASSERT_WITH_MESSAGE(c,m) do { static_assert((c),m); } while(0)
+#    define XXH_STATIC_ASSERT_WITH_MESSAGE(c,m) do { _Static_assert((c),m); } while(0)
 #  elif defined(__cplusplus) && (__cplusplus >= 201103L)            /* C++11 */
 #    define XXH_STATIC_ASSERT_WITH_MESSAGE(c,m) do { static_assert((c),m); } while(0)
 #  else


ThomasWaldmann:
Had a lot of troubles due to this when trying to build borgbackup 1.1.18 release on FreeBSD 12.

The xxhash package in ports does not have this patch and the build is stumbling over "undefined symbol static_assert". Not sure why this worked ok for 1.1.17 release.

Would be cool to have a new xxhash release with all the fixes since last release.

ThomasDevoogdt:
I also see the same problem on uClibc toolchains. This fix seems to be fine.

PR #668, opened 2021-12-24
Fix build with clang

pkubaj:
Currently build fails with error on FreeBSD/powerpc64le:
```
In file included from /wrkdirs/usr/ports/mail/rspamd-devel/work/rspamd-93430bb/contrib/xxhash/xxhash.h:3065:
/usr/lib/clang/11.0.1/include/altivec.h:55:19: error: unknown type name 'vector'
static __inline__ vector bool char __ATTRS_o_ai
                  ^
/usr/lib/clang/11.0.1/include/altivec.h:56:10: error: unknown type name 'vector'
vec_perm(vector bool char __a, vector bool char __b, vector unsigned char __c);
         ^
/usr/lib/clang/11.0.1/include/altivec.h:56:32: error: unknown type name 'vector'
vec_perm(vector bool char __a, vector bool char __b, vector unsigned char __c);
                               ^
/usr/lib/clang/11.0.1/include/altivec.h:66:19: error: unknown type name 'vector'
static __inline__ vector bool short __ATTRS_o_ai vec_perm(
                  ^
/usr/lib/clang/11.0.1/include/altivec.h:67:5: error: unknown type name 'vector'
    vector bool short __a, vector bool short __b, vector unsigned char __c);
    ^
/usr/lib/clang/11.0.1/include/altivec.h:67:28: error: unknown type name 'vector'
    vector bool short __a, vector bool short __b, vector unsigned char __c);
                           ^
/usr/lib/clang/11.0.1/include/altivec.h:80:19: error: unknown type name 'vector'
static __inline__ vector bool int __ATTRS_o_ai
                  ^
/usr/lib/clang/11.0.1/include/altivec.h:81:10: error: unknown type name 'vector'
vec_perm(vector bool int __a, vector bool int __b, vector unsigned char __c);
         ^
/usr/lib/clang/11.0.1/include/altivec.h:81:31: error: unknown type name 'vector'
vec_perm(vector bool int __a, vector bool int __b, vector unsigned char __c);
                              ^
/usr/lib/clang/11.0.1/include/altivec.h:96:19: error: unknown type name 'vector'
static __inline__ vector bool long long __ATTRS_o_ai
                  ^
/usr/lib/clang/11.0.1/include/altivec.h:97:10: error: unknown type name 'vector'
vec_perm(vector bool long long __a, vector bool long long __b,
         ^
/usr/lib/clang/11.0.1/include/altivec.h:97:37: error: unknown type name 'vector'
[...]

PR #672, opened 2021-12-30
Properly fix altivec namespace collisions

easyaspi314:
Uses `#pragma push_macro`/`#pragma pop_macro` instead of assuming that`__APPLE_ALTIVEC__` can be defined safely.

Safer and less hacky fix for #668.

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -3350,22 +3350,32 @@ XXH_FORCE_INLINE uint64x2_t XXH_vld1q_u64(void const* ptr)
  * inconsistent intrinsics, spotty coverage, and multiple endiannesses.
  */
 #if XXH_VECTOR == XXH_VSX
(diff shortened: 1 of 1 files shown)


easyaspi314:
CC @pkubaj
can you try this?

pkubaj:
Yes, that works as well, thanks!

Cyan4973:
minor concern : 
how certain are we that compilers following this code branch (`#if XXH_VECTOR == XXH_VSX`)
necessarily support the `#pragma push` statement ?

easyaspi314:
> minor concern : 
> how certain are we that compilers following this code branch (`#if XXH_VECTOR == XXH_VSX`)
> necessarily support the `#pragma push` statement ?

Because the VSX path is only enabled on GCC compatible compilers.

Cyan4973 (approved):

PR #683, opened 2022-02-18
Add the assembly implementation for aarch64 SVE

hzhuang1:
It's the result of bench test among SCALAR and SVE implementation.
The test machine is based on Fujitsu A64FX SoC with SVE 512-bit vector.

The only difference between SVE (C) and SVE (ASSEMBLY) is on the
frequency of loading and storing ACC variable. ACC variable is 512 bits
long.

In SVE (ASSEMBLY) implementation, ACC variable is only stored into
memory while calculation is finished. It could avoid to load and store
ACC variable frequently. The performance is improved on nearly 10%-70%
based on SCALAR implementation.

According to the microarchitecture spec of A64FX in
https://github.com/fujitsu/A64FX/, loading and storing data by SVE
instructions cost much more cycles than SCALAR instructions.

======================================================================
| XXH3       | 512B |  1KB |  2KB |  4KB |  8KB | 16KB | 32KB | 64KB |
+------------+------+------+------+------+------+------+------+------+
| SCALAR     | 2688 | 3084 | 3268 | 3406 | 3487 | 3524 | 3543 | 3509 |
| (aarch64)  |      |      |      |      |      |      |      |      |
+------------+------+------+------+------+------+------+------+------+
| SVE        | 1895 | 2106 | 2174 | 2207 | 2227 | 2236 | 2227 | 2236 |
| (C)        |      |      |      |      |      |      |      |      |
+------------+------+------+------+------+------+------+------+------+
| SVE        | 2858 | 4012 | 4943 | 5522 | 5868 | 6053 | 6142 | 6106 |
| (assembly) |      |      |      |      |      |      |      |      |
======================================================================

Build instructions:
$export CC=clang
$export CPP_FLAGS="-DXXH_VECTOR=XXH_SVE -DXXH_IMPL=XXH_IMPL_ASSEMBLY"
$export CFLAGS="-O3 -march=armv8-a+sve -fPIC -DXXH_IMPL=XXH_IMPL_ASSEMBLY"
$export AARCH64_DISPATCH=1
$make


easyaspi314:
How does this compare to vanilla NEON? 

Also do you think you could make some asm macros for that? Scanning over it, there seems to be a lot of copy/pasted blocks.

hzhuang1:
> How does this compare to vanilla NEON?
> 
> Also do you think you could make some asm macros for that? Scanning over it, there seems to be a lot of copy/pasted blocks.

The performance of NEON is worse than SCALAR. So I didn't list it in the comment.

Let me check whether I can use macro to simplify code.

easyaspi314:
> The performance of NEON is worse than SCALAR. 

I might have to look into the microarchitecture again to see if there is a better NEON strat. I seem to recall seeing something similar on one of the compiler farm CPUs (Lenovo HR350A I think).

There's like 15 ways of doing it, especially when you combine the NEON and scalar paths. It isn't obvious like SSE2 thanks to the janky `umlal` instruction.

But that is a different issue so ignore me 😅

hzhuang1:
@Cyan4973 @easyaspi314 Excuse me. Is there any comments on this pull request? Thanks

easyaspi314:
Let me take a deeper look at it. I can't test anything properly as I don't have access to SVE. 

One thing I noticed was that the code isn't entirely position-independent. Can you please fix that?

**edit:** Sorry for the double post.
It seems that there is no reason for the `xxh3_prime_1`, it can just be converted to `mov/movk`.

easyaspi314:
Let me take a deeper look at it. 

One thing I noticed was that the code isn't entirely position-independent. Can you please fix that?

**Edit:** Can't this just be `mov/movk`?

easyaspi314:
Ok I scanned over the code. 

My comments
 1. I am very concerned about the complexity of this assembly, as it could easily become write-only code. 
   - Can I have a benchmark of the various vector lengths with these unrolled functions vs a straightforward VL-agnostic path? If the difference is negligible, I don't want this complexity.
   -  There already seems to be some inconsistency in the various code blocks. For example, `XXH3_aarch64_sve512_internal_loop` seems to incorrectly set the frame pointer.
 2. Can a dispatcher be added to select between SVE and NEON? This should be a simple `getauxval(AT_HWCAP)` check for `HWCAP_SVE`. That is basically what the dispatcher does on x86, except it uses `cpuid`.
   - As far as I know, there is no simple way to check if scalar is faster than NEON without a micro bench. In general, most mobile devices prefer the NEON path, with only some server-class devices preferring scalar.
   - In that case, the xxhash.h file should be compiled without SVE enabled. This means that the dispatch must be made in assembly. 
 3. As stated before, please convert the constant load to a mov/movk to avoid PIC issues. (And even still, that should be in `.rodata`, not `.data`)
 4. Is this code safe if the vectors are larger than 512?

xxh_aarch64dispatch.S
```diff
@@ -0,0 +1,1043 @@
...
+	addvl	sp, sp, #-6
+	mov	x29, sp
+	/*
+	 * prepare index
+	 * p7 and z7 are reserved to use.
+	 */
+	ptrue	p7.d
+	init_index
+
+	st1d	{z8.d}, p7, [sp, #1, MUL VL]	// save z8
```
easyaspi314, line 57: If I'm not mistaken, z16-z31 are caller saved.

hzhuang1:
> Ok I scanned over the code.
> 
> My comments
> 
>     1. I am very concerned about the complexity of this assembly, as it could easily become write-only code.
> 
> 
>     * Can I have a benchmark of the various vector lengths with these unrolled functions vs a straightforward VL-agnostic path? If the difference is negligible, I don't want this complexity.
> 
Of course, I don't like to handle various vector lengths too. Since it's a bit complex. The main optimization point is to avoid to save ACC from Z register into memory in the calculation.
For SVE-512, SVE-1024 & SVE-2048, the V register could hold the 512-bit ACC without any issue.
For SVE-128, the V register could only hold one 128-bit ACC. So I have to use 4 Z registers to store the whole ACC. For SVE-256, I have to use 2 Z registers to store the whole ACC. So it's a bit complex, and I don't have any better solution.

>     * There already seems to be some inconsistency in the various code blocks. For example, `XXH3_aarch64_sve512_internal_loop` seems to incorrectly set the frame pointer.
> 
Let me fix it.
> 
>     2. Can a dispatcher be added to select between SVE and NEON? This should be a simple `getauxval(AT_HWCAP)` check for `HWCAP_SVE`. That is basically what the dispatcher does on x86, except it uses `cpuid`.
> 
Let me check.
> 
>     * As far as I know, there is no simple way to check if scalar is faster than NEON without a micro bench. In general, most mobile devices prefer the NEON path, with only some server-class devices preferring scalar.
> 
Oh, yes. I just tried the xxhash bench on HiKey960 platform with mobile SoC. NEON gains 30% improvement that SCALAR.
But on server, it's different.
Currently, we can use macro to decide whether use NEON/SVE or SCALAR. It's totally decided by user.
By the way, what's the definition on "micro bench"?

>     * In that case, the xxhash.h file should be compiled without SVE enabled. This means that the dispatch must be made in assembly.
> 
It's fine to change the dispatch to assembly.
> 
>     3. As stated before, please convert the constant load to a mov/movk to avoid PIC issues. (And even still, that should be in `.rodata`, not `.data`)
> 
OK. Let me follow it up.

>     4. Is this code safe if the vectors are larger than 512?

Yes, I verified it on both SVE-1024 and SVE-2048 with virtual machine.

xxh_aarch64dispatch.S
hzhuang1: It's mentioned in AAPCS64.

`z0-z7 are used to pass scalable vector arguments to a subroutine, and to return scalable vector results from a function. If a subroutine takes at least one argument in scalable vector registers or scalable predicate registers, or if it is a function that returns results in such registers, it must ensure that the entire contents of z8-z23 are preserved across the call. In other cases it need only preserve the low 64 bits of z8-z15, as described in SIMD and Floating-Point registers.`

So I have to save z8. And I also tried in program, I may meet some software error without saving z8.

easyaspi314: I mean why not just use z16 instead of z8?

If I am reading this correctly, since this falls in the latter case, z16-z31 are fair game and don't need to be saved.

hzhuang1: I can try to use z16. I don't know z16 is saved by caller before, and I couldn't find it from AAPCS64. I'll try it.

xxh_aarch64dispatch.S
```diff
@@ -0,0 +1,1043 @@
...
+	tbl	z5.d, {z1.d}, z7.d
+	// mixed = input EOR secret
+	eor	z1.d, p7/m, z1.d, z2.d
+	mov	z6.d, z1.d
+	// mixed_lo = mixed AND 0xffffffff
+	and	z1.d, p7/m, z1.d, z3.d
+	// mixed_hi = mixed >> 32
+	lsr	z6.d, p7/m, z6.d, z4.d
+	// mixed_lo = mixed_hi * acc + mixed_lo
+	mad	z1.d, p7/m, z6.d, \acc
```
easyaspi314, line 28: You should be able to avoid the temporary with the distributive property like this:
```asm
        mad z1.d, p7/m, z6.d, z5.d
        add \acc, p7/m, \acc, z1.d
```

easyaspi314:
@hzhuang1, can you please test this simplified version I wrote? I think this would be more viable and easy to convert back to C once compilers get better at SVE.

It only replaces `XXH3_accumulate()` (not the relatively minor scramble and cleanup paths which can just use scalar), but *in theory* it should be about the same speed as the full ASM version.

I also found a few optimizations to the accumulate loop which allow me to fit it into `z0-z7`, letting me completely avoid the stack, and I made it theoretically compatible with SVE384 if anyone is silly enough to use it.

[xxh3_aarch64_sve.S](https://gist.github.com/easyaspi314/b9257ba72853e9c4391aad7ee8cd9ca2)

easyaspi314:
@Cyan4973  Should we dispatch at `XXH3_accumulate()` instead of `XXH3_accumulate_512()`? That would make incorporating this much easier and probably improve performance of the dispatch code as well.

Cyan4973:
> @Cyan4973 Should we dispatch at `XXH3_accumulate()` instead of `XXH3_accumulate_512()`? That would make incorporating this much easier and probably improve performance of the dispatch code as well.

I'm unable to tell.
This _sounds_ like a good idea, given the list of merits you provide,
but I'm not able to fully grasp the consequences of such a change.

I guess the better way forward is to try it.

easyaspi314:
Hmm, I translated it directly to C and even GCC 10 and Clang 11 seem to compile it fairly well AFAICT. We might not even need assembly.

However, it seems that Clang doesn't support the intrinsics with --target overridden properly on the x86_64 version before 13.0.

Additionally, Clang 11 whines a lot about the code even though it seems to compile fine. Since this comes from LLVM instead of Clang, this can't be supressed (and LLVM can be compiled in a way that forces an error).

Godbolt: https://godbolt.org/z/9v1fjEnfa

easyaspi314:
Granted, all of this is untested.

As for the accumulate cutoff, this will avoid reloads since the function won't be called via function pointer, and writing `XXH3_accumulate` entirely in SVE will avoid the major problem where Clang and GCC are both incapable of load-store optimization as of trunk. Unlike x86, ARM does not like working from memory.

Specifically, GCC and Clang are currently unable to optimize something like this
```c
void foo(uint64_t *xacc, size_t n)
{
    for (size_t i = 0; i < n; i++) {
        svuint64_t acc = svld1(svptrue_b64(), xacc); // load
        // .... do stuff with acc with no side effects to memory
        svst1(svptrue_b64(), xacc, acc); // store
    }
}
```
to this
```c
void foo(uint64_t *xacc, size_t n)
{
    svuint64_t acc = svld1(svptrue_b64(), xacc); // load
    for (size_t i = 0; i < n; i++) {
        // .... do stuff with acc with no side effects to memory
    }
    svst1(svptrue_b64(), xacc, acc); // store
}
```
which is *critical* in the performance of XXH3.

*This would be soooo much easier to do if C had templates*

xxh_aarch64dispatch.S
hzhuang1: Thanks a lot. It could save one instruction.

hzhuang1:
> @hzhuang1, can you please test this simplified version I wrote? I think this would be more viable and easy to convert back to C once compilers get better at SVE.
> 
> It only replaces `XXH3_accumulate()` (not the relatively minor scramble and cleanup paths which can just use scalar), but _in theory_ it should be about the same speed as the full ASM version.
> 
> I also found a few optimizations to the accumulate loop which allow me to fit it into `z0-z7`, letting me completely avoid the stack, and I made it theoretically compatible with SVE384 if anyone is silly enough to use it.
> 
> [xxh3_aarch64_sve.S](https://gist.github.com/easyaspi314/b9257ba72853e9c4391aad7ee8cd9ca2)

[SVE xxh3_aarch64_sve.S] (https://github.com/hzhuang1/xxHash/tree/dirty_v0.3.2)

benchmarking large inputs : from 512 bytes (log9) to 128 MB (log27) 
xxh3   ,  4131,  6638,  7159,  7492,  7896,  8395,  8601,  8665,  8853,  8891,  8920,  8943,  8933,  7939,  3125,  2684,  2705,  3198,  3203
XXH32  ,  1326,  1440,  1485,  1518,  1534,  1540,  1547,  1536,  1503,  1506,  1506,  1502,  1503,  1466,  1263,  1208,  1210,  1208,  1209
XXH64  ,  2497,  2801,  2937,  3063,  3121,  3138,  3146,  3107,  3028,  3042,  3049,  3047,  3044,  2929,  2190,  2021,  2012,  2010,  2013
XXH128 ,  3681,  6029,  6780,  7276,  7753,  8264,  8631,  8640,  8851,  8933,  8964,  8984,  8979,  8086,  4149,  3661,  3189,  3204,  3204
Throughput small inputs of fixed size (from 1 to 127 bytes): 
xxh3   ,  72726845,  72795674,  72790111,  68536502,  68567226,  68542624,  68527008,  68546252,  74380651,  74322456,  74384399,  74238448,  74326510,  74369229,  74356894,  74320430,  39162000,  39150303,  39167118,  39148461,  39182573,  39182213,  39131125,  43928151,  39169789,


[SVE xxh_aarch64dispatch.S] (https://github.com/hzhuang1/xxHash/tree/drity_v0.3)
 ===  benchmarking 4 hash functions  === 
benchmarking large inputs : from 512 bytes (log9) to 128 MB (log27) 
xxh3   ,  3885,  6154,  9175, 11539, 13250, 14228, 14718, 12839, 11377, 11104, 11158, 11123, 11159, 10228,  4226,  2788,  2794,  2801,  3753
XXH32  ,  1326,  1438,  1494,  1522,  1533,  1542,  1545,  1529,  1487,  1486,  1488,  1487,  1487,  1443,  1249,  1199,  1196,  1196,  1188
XXH64  ,  2509,  2786,  2971,  3071,  3119,  3134,  3151,  3112,  2981,  2991,  2994,  2994,  2994,  2862,  2136,  1993,  1989,  1985,  1996
XXH128 ,  3710,  6134,  8888, 11407, 12941, 14092, 14624, 12518, 10886, 11019, 11093, 11124, 11135, 10038,  4732,  4113,  3760,  3739,  3785
Throughput small inputs of fixed size (from 1 to 127 bytes): 
xxh3   ,  72690593,  72692222,  72599159,  68504118,  68482115,  68505888,  68488296,  68398111,  74437184,  74445105,  74282793, 

From the above two log, it works better with original assembly code. This performance data is better than initial code since your advice is adopted.

easyaspi314:
> From the above two log, it works better with original assembly code.

I expected it to be a slower, but not *that* much slower. I wonder if this has to do with the prefetch, can you try it without the `prfm` instructions?

Also, how does my C version in the godbolt link above perform? Make sure to make it `XXH_FORCE_INLINE` and remove the target attribute.

BTW, here is a visual of that benchmark:
xxh3 and xxh3 full asm(1)
You can clearly see the cache lines (matches the 64KiB L1 and 8MiB L2 on the spec sheet), which makes me think the prefetch is causing the slowdown.

**Edit: I believe I used the wrong prefetch instruction.** Apparently, SVE has its own prefetch instruction.

hzhuang1:
> 
> I expected it to be a slower, but not _that_ much slower. I wonder if this has to do with the prefetch, can you try it without the `prfm` instructions?
>
The prefetch instruction doesn't help. Here's the data without prefetch.
 ===  benchmarking 4 hash functions  === 
benchmarking large inputs : from 512 bytes (log9) to 128 MB (log27) 
xxh3   ,  4144,  7005,  7588,  8022,  8340,  8925,  9311,  8622,  7828,  7403,  7497,  7510,  7465,  6390,  3280,  3209,  3223,  3247,  3267
XXH32  ,  1322,  1434,  1490,  1517,  1529,  1543,  1545,  1522,  1490,  1497,  1495,  1498,  1497,  1442,  1254,  1201,  1207,  1199,  1204
XXH64  ,  2502,  2802,  2976,  3070,  3120,  3135,  3151,  3112,  3005,  3012,  3016,  3016,  3016,  2828,  2142,  1988,  2007,  2015,  2011
XXH128 ,  3631,  6320,  7145,  7782,  8299,  8954,  9325,  8610,  7690,  7750,  7787,  7803,  7799,  7025,  4112,  3665,  3277,  3307,  3270
Throughput small inputs of fixed size (from 1 to 127 bytes): 
xxh3   ,  72769981,  72770840,  72554169,  67856810^C
 
>  You can clearly see the cache lines (matches the 64KiB L1 and 8MiB L2 on the spec sheet), which makes me think the prefetch is causing the slowdown.
>
Yes, the prefetch instruction hurted the performance. But it only impacted a little.
 
> **Edit: I believe I used the wrong prefetch instruction.** Apparently, SVE has its own prefetch instruction.

Let me check whether there's any SVE prefetch instruction. But I'm not sure whether prefetch instruction could gain a lot of performance even without SVE.

hzhuang1:
> Also, how does my C version in the godbolt link above perform? Make sure to make it `XXH_FORCE_INLINE` and remove the target attribute.
> 

It's good. (https://github.com/hzhuang1/xxHash/tree/dirty_v0.3.3) The performance data is in below.

 ===  benchmarking 4 hash functions  === 
benchmarking large inputs : from 512 bytes (log9) to 128 MB (log27) 
xxh3   ,  4099,  6926,  8365,  9887, 10811, 11311, 11555, 10982, 10438,  9755,  9788,  9763,  9820,  8668,  4709,  4143,  4194,  4166,  4191
XXH32  ,  1326,  1441,  1496,  1523,  1534,  1544,  1547,  1535,  1501,  1503,  1503,  1504,  1503,  1461,  1253,  1205,  1209,  1206,  1207
XXH64  ,  2510,  2803,  2976,  3073,  3122,  3140,  3155,  3123,  3032,  3027,  3043,  3044,  3042,  2909,  2149,  2004,  2015,  1995,  2009
XXH128 ,  3501,  6098,  8131,  9596, 10682, 11227, 11537, 11017, 10444, 10635, 10694, 10711, 10728,  9699,  4890,  4440,  4222,  4203,  4199
Throughput small inputs of fixed size (from 1 to 127 bytes): 
xxh3   ,  74940682,  74940242,  74905421,  29463309,  29447211,

hzhuang1:
> Also, how does my C version in the godbolt link above perform? Make sure to make it `XXH_FORCE_INLINE` and remove the target attribute.

I checked your code. The main difference in each acc512 is focus on "extw" instruction. I tried it before and I failed to use it. How to understand the "extw" instruction?

easyaspi314:
> I checked your code. The main difference in each acc512 is focus on "extw" instruction. I tried it before and I failed to use it. How to understand the "extw" instruction?

EXTract Word to double word. `x = (uint64_t)(uint32_t)y;`, doing the same thing as `x = y & 0xFFFFFFFF` but not requiring a `MOVPRFX` to be non-destructive.

It is an alias to `uxtw` which is what I use in the assembly version and the more natural mnemonic in relation to the rest of the ISA.

If I make a final version, I will comment everything.

easyaspi314:
~~thanks github for double posting~~

Does this pass the self-tests btw?

hzhuang1:
If I apply your improvement code on assembly code, the best performance could be beyond 15000. And your intrinsic code could make performance reach 11xxx. Whatever, both of them are better than SCALAR implementation. So what's the next plan? Do we keep to use assembly code or switch to intrinsic code?

hzhuang1:
> 

Sure. I always run it with "make test" to pass self-test first.

easyaspi314:
> If I apply your improvement code on assembly code, the best performance could be beyond 15000. And your intrinsic code could make performance reach 11xxx. 

I honestly wonder how much of that is just compilers not being able to optimize SVE yet. My phone gets 10-11 GB/s on the hybrid NEON path 😂

Are there any platforms that use SVE256? The only ones I know of using SVE are either 128-bit (Cortex-X2, Neoverse) or 512-bit (A64FX).

If so, we could instead have dispatch targets for 128-bit (also compatible with 256- and 384-bit, but trimmed to a 128-bit mask) and 512-bit and larger, and instead of branching each time, similar to AVX dispatching on x86.

hzhuang1:
> > If I apply your improvement code on assembly code, the best performance could be beyond 15000. And your intrinsic code could make performance reach 11xxx.
> 
> I honestly wonder how much of that is just compilers not being able to optimize SVE yet. My phone gets 10-11 GB/s on the hybrid NEON path joy
> 
What's the hybrid NEON path? The code in mainstream is either to use NEON or not. How to make it happen with the hybrid path?

> Are there any platforms that use SVE256? The only ones I know of using SVE are either 128-bit (Cortex-X2, Neoverse) or 512-bit (A64FX).
>
https://www.techinsights.com/blog/graviton3-debuts-neoverse-v1
It seems that Graviton3 supports SVE256. But I can only access A64FX.

> If so, we could instead have dispatch targets for 128-bit (also compatible with 256- and 384-bit, but trimmed to a 128-bit mask) and 512-bit and larger, and instead of branching each time, similar to AVX dispatching on x86.

So you're talking on XXH_PREFETCH_DIST, right? I can only find it related to 384.

easyaspi314:
> What's the hybrid NEON path? The code in mainstream is either to use NEON or not. How to make it happen with the hybrid path?

Ah, you aren't using the latest `dev`? The latest `dev` does 6 lanes NEON and 2 lanes scalar for AArch64 and it is configurable at compile time.

This leverages the execution units better as most ARM chips can only dispatch N of each *type* of instruction, and NEON and scalar multiplication have a lot of *independent* latency. On most Cortex-based designs I tested, it is either very beneficial, or at the very worst case only slightly slower than full NEON.

Theoretically, 4/4 is better, but GCC and Clang both haven't fully understood how `umaddl` is better than `and+madd`, especially when the cycle counts clearly indicate a 64-bit multiply being 3 32-bit multiplies in microcode on most of the Cortex ARMv8 processors.

> So you're talking on XXH_PREFETCH_DIST, right? I can only find it related to 384.

SVE can technically be set to 384 bits. Granted it is incredibly stupid to do so, but the code should still support it. Mine just puts 384 on 256 with a mask.

hzhuang1:
@easyaspi314 Let me summarize the next goal. Please help to check whether my understanding is correct.

Since xxHash should be built as inline, intrinsic code should be used at this time. For it's hard to embed assembly code into head file.
If dispatch mode is enabled (detecting different arm architecture dynamically), assembly code could be used at this time.

Is it right?

hzhuang1:
@Cyan4973 @easyaspi314 Please help to review the prefetch parameter in (https://github.com/hzhuang1/xxHash/tree/up_sve).

We discussed prefetch before that is based on __builtin_prefetch(). Now I try to switch to SVE prefetch instruction. And get the result in below.
```
    prefetch (128):
     ===  benchmarking 4 hash functions  ===
    benchmarking large inputs : from 512 bytes (log9) to 128 MB (log27)
    xxh3   ,  3939,  6547,  8155,  9428, 10133, 10541, 10710, 10529, 10579, 10811, 10827, 10787, 10830, 10019,  6371,  5965,  5986,  5922,  5851
    XXH32  ,  1326,  1440,  1495,  1523,  1534,  1543,  1545,  1535,  1494,  1494,  1495,  1494,  1493,  1454,  1255,  1183,  1185,  1181,  1175
    XXH64  ,  2509,  2796,  2974,  3051,  3122,  3128,  3145,  3122,  3000,  3008,  3011,  3009,  3010,  2882,  2145,  1965,  1958,  1942,  1927
    XXH128 ,  3614,  5995,  7645,  9090,  9955, 10448, 10525, 10508, 10531, 10600, 10637, 10659, 10663,  9905,  6132,  5741,  5937,  5894,  5865
    
    prefetch (320):
     ===  benchmarking 4 hash functions  ===
    benchmarking large inputs : from 512 bytes (log9) to 128 MB (log27)
    xxh3   ,   721,   665,  7697,  6332,  7810,  8703,  9219,  9393,  9806,  9971,  9968, 10039, 10024,  9348,  7109,  7005,  7047,  7102,  7163
    XXH32  ,  1492,  1523,  1534,  1546,  1544,  1546,  1548,  1537,  1502,  1504,  1504,  1505,  1503,  1456,  1253,  1198,  1178,  1184,  1181
    XXH64  ,  2631,  2872,  3017,  3092,  3131,  3141,  3152,  3135,  3043,  3049,  3048,  3051,  3046,  2904,  2165,  1991,  1963,  1947,  1936
    XXH128 ,  3854,  6339,  7226,  8198,  8960,  9402,  9659,  9791,  9969, 10052, 10154, 10082, 10147,  9419,  7161,  7068,  7092,  7148,  7183
```

While prefetch parameter is increased from 128 to 320, the larger data block could gain better performance while the smaller data block drop a lot.
I could understand why the result are different. How to understand the prefetch?

hzhuang1:
> 
> Since xxHash should be built as inline, intrinsic code should be used at this time. For it's hard to embed assembly code into head file. If dispatch mode is enabled (detecting different arm architecture dynamically), assembly code could be used at this time.
>
@Cyan4973 @easyaspi314 
I think it's a bit contradictory.
In xxhash.h, XXH_VECTOR is used to make sure only one architecture implementation built at one time. For example, XXH_VECTOR should be either XXH_SCALAR or XXH_NEON. So either XXH3_accumulate_scalar() or XXH3_accumulate_neon() is compiled.
In dispatch, we hope it could detect different architecture automatically. We hope that XXH3_accumulate_scalar() and XXH3_accumulate_neon() should be both ready. It caused the compiling issue.

easyaspi314:
Well with an ugly hack (manually defining `__ARM_FEATURE_SVE`) it is 100% possible to conditionally enable SVE on a per-function basis.

hzhuang1:
Close since #697 could replace it.

PR #688, opened 2022-03-10
tipi.build instructions

pysco68:
Hey!  I Love your library! I was asked by a customer if/how they would use it in a [tipi.build](https://tipi.build/) project amidst a demo session. I thought you might be happy to add the instructions to the readme etc.

I went ahead and added a CI step to test is building the project (and depending on it) still work as expected. Feel free to remove that, here's how the result would look like: https://github.com/tipi-deps/xxHash/actions/runs/1962088487

Best, 
- Yannic


Cyan4973:
That's fine @pysco68 ,
I like that most changes are within a `.tipi` sub-directory,
so the only "visible" part affecting non-tipi users is the documentation part.

Let's get these CI tests working, and it will be good to go.

pysco68:
Looks like this time it was fine. I saw that one of the Linux images failed on an apt install which looked like a broken package mirror. 

How should we move on?

Cyan4973:
Let's merge it.
I'll take care of the rest on my side.

Cyan4973 (approved):

PR #694, opened 2022-03-15
Fix neon

hzhuang1:
Fix build issues related to NEON

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -2998,9 +2998,7 @@ XXH_PUBLIC_API XXH64_hash_t XXH64_hashFromCanonical(const XXH64_canonical_t* src
 #endif
 
 #if defined(__GNUC__) || defined(__clang__)
-#  if defined(__ARM_NEON__) || defined(__ARM_NEON) \
-   || defined(__aarch64__)  || defined(_M_ARM) \
-   || defined(_M_ARM64)     || defined(_M_ARM64EC)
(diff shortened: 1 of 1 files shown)


easyaspi314:
Isn't building for AArch64 without NEON a little odd? Aside from maybe osdev where you can't use FPU registers?

If you want to build the scalar version, just define `-DXXH_VECTOR=XXH_SCALAR`.

easyaspi314:
> `Macro "__arm__" indicates armv7. Without it, "arm_neon.h" isn't included
for NEON.`

**Incorrect.** `__arm__` indicates ARM, not ARMv7. (It also doesn't detect Thumb mode).

If you compile for ARMv6, this would error, as `arm_neon.h` cannot be included without NEON support, unlike the SSE intrinsic headers.

The correct way to detect ARM NEON for ARMv7 is either `__ARM_NEON` or `__ARM_NEON__`, although `__aarch64__`, `_M_ARM` (MSVC ARMv7), and `_M_ARM64` indicate NEON support by default.

hzhuang1:
> 
> The correct way to detect ARM NEON for ARMv7 is either `__ARM_NEON` or `__ARM_NEON__`, although `__aarch64__`, `_M_ARM` (MSVC ARMv7), and `_M_ARM64` indicate NEON support by default.

OK. Let me try.

hzhuang1:
It's not urgent. Close it first.

PR #696, opened 2022-03-29
Build fix for Haiku

Begasus:


Begasus:
This is actually for the first failing test:
```
cli/xsum_os_specific.c:78:35: error: unused parameter 'stream'
```

Begasus:
I've added a new version to our haikuports based on the current commit, part of the patch is here the other one can be found here: https://github.com/Begasus/haikuports/blob/52f725755a883cca5153dc2ae177aaa651b77f13/dev-libs/xxhash/patches/xxhash-0.8.1%7Egit.patchset
Not sure if the other 2 are fine to be upstreamed (well maybe a ifneq for Haiku for -pthread could be added, will hear from you) :)

Cyan4973 (approved):

Begasus:
Thanks! :+1:

PR #697, opened 2022-04-01
Up sve 3

hzhuang1:
Implement arm64 dispatch with assembly SVE routines.


hzhuang1:
It's a new version of #683 
In this version, both intrinsic code and assembly code are included.

hzhuang1:
The first two commits come from Devin. And the third commit is used to fix the CI errors.
Let's review the last 3 commits.

hzhuang1:
@Cyan4973 @easyaspi314 How about this patch set? Could you have any comments to share? Thanks

hzhuang1:
> @Cyan4973 @easyaspi314 How about this patch set? Could you have any comments to share? Thanks

@Cyan4973 @easyaspi314 What should I do for next? Thanks

Cyan4973:
Hi @hzhuang1 

I don't think you need to do anything else.
It's mostly a matter of time, we need to find enough availability to make a proper review of your code.
You did as much and as well as you could up to now.
Now the ball is on our side.

Makefile
```diff
@@ -97,6 +97,11 @@ all: lib xxhsum xxhsum_inlinedXXH
 ifeq ($(DISPATCH),1)
 xxhsum: CPPFLAGS += -DXXHSUM_DISPATCH=1
 xxhsum: xxh_x86dispatch.o
+DISPATCH_HEADERS = xxh_x86dispatch.h
+else ifeq ($(ARM64_DISPATCH),1)
```
Cyan4973: `ARM64_DISPATCH` provides a way to manually set dispatch for ARM64.
This works if `DISPATCH` is not set, and needs to be documented, otherwise it's unlikely to be discovered.

Another alternative could be to automate `ARM64` target detection.
This would require calling the compiler with its intended set of compilation flags, and make it tell which platform it's targeting, then trigger `x86` or `arm64` dispatch depending on the detected platform, or even opt out if detection fails or feels fishy. 
Ideally, it should be possible to override automated detection with manual settings, just in case, and also to satisfy requirements from package managers.

I understand this is additional nice-to-have, and makes more sense to invest into it once the main patch is accepted.

xxh_arm64dispatch.h
```diff
@@ -0,0 +1,86 @@
...
+#ifndef XXH_ARM64DISPATCH_H
+#define XXH_ARM64DISPATCH_H
+
+#include "xxhash.h"  /* XXH64_hash_t, XXH3_state_t */
+
+#if defined (__cplusplus)
+extern "C" {
+#endif
+
+XXH64_hash_t  XXH3_64bits_dispatch(const void* input, size_t len);
```
Cyan4973: This seems to be exactly the same interface contract as `xxh_x86dispatch.h` ?
In which case, maybe it's better to combine them ?

xxh_arm64sve.S
```diff
@@ -0,0 +1,475 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * You can contact the author at:
+ *   - xxHash homepage: https://www.xxhash.com
+ *   - xxHash source repository: https://github.com/Cyan4973/xxHash
+ */
+
+	.arch armv8-a+sve
+
+#include "asmdefs.h"
```
Cyan4973, line 38: This is an assembly file, not a C file with some assembly.
It means the compilation pipeline will be different for this file.

This may look like a minor detail, but I've experienced critical issues doing the same thing in other projects.

Probably the most critical one is the impact this can have on the executable stack flag, which then trigger security related issues.
It needs to be taken care of at compiler level.
This can be an annoyance to setup. You may refer to https://github.com/facebook/zstd/pull/2964/files#diff-a19342636d0d39fbbfc6ebf4e3839ed2ae4e043e7c9e37ca96226c0b7ad3a499R101 for an example.

xxhash.h
```diff
@@ -4861,20 +5077,20 @@ XXH3_hashLong_internal_loop(xxh_u64* XXH_RESTRICT acc,
...
     {   size_t const nbStripes = ((len - 1) - (block_len * nb_blocks)) / XXH_STRIPE_LEN;
         XXH_ASSERT(nbStripes <= (secretSize / XXH_SECRET_CONSUME_RATE));
-        XXH3_accumulate(acc, input + nb_blocks*block_len, secret, nbStripes, f_acc512);
+        f_acc(acc, input + nb_blocks*block_len, secret, nbStripes);
 
         /* last stripe */
         {   const xxh_u8* const p = input + len - XXH_STRIPE_LEN;
 #define XXH_SECRET_LASTACC_START 7  /* not aligned on 8, last secret is different from acc & scrambler */
-            f_acc512(acc, p, secret + secretSize - XXH_STRIPE_LEN - XXH_SECRET_LASTACC_START);
+            XXH3_accumulate_512(acc, p, secret + secretSize - XXH_STRIPE_LEN - XXH_SECRET_LASTACC_START);
```
Cyan4973, line 5093: Question :
how can you guarantee that `XXH3_accumulate_512()` will trigger the right variant
in the context of dynamic dispatch ?

Cyan4973:
Could you document, in the PR, the performance implications of this patch ?

Cyan4973:
I was wondering if this PR could be made smaller by separating 2 components : 
1) introduce a new `SVE` micro-op support (triggered through static selection only)
2) then introduce a dynamic dispatcher for `arm64`

Unless I'm missing something, it seems this could help to review each part specifically.

Cyan4973:
Summary : it looks like great work,
but it also looks like a lot of code to add, 
so we have to be cautious about its potential side effects,
both for existing user base, and for future maintenance.

Anything that can help make this PR smaller will be welcome.
Hence the suggestion to split it into 2 parts.

It may also trigger deeper questions about the repository structure.
The core library `xxhash` was built around the requirement of being a "single file library" (not counting here the CLI, nor the dispatcher).
This PR seems to change that requirement, by introducing an additional assembly file for `SVE`.
And of course, more files for the dispatcher.

So it's a direct contributor to increased repository complexity.
This could nonetheless be worth it, but this complexity must be brought under control, in order to keep this repository usable for "casual" users that just want to grab `"xxhash.h"` and maybe `"xxhash.c"` and call it a day.
I think it's still possible for non `SVE` users, but it's less and less clear that it is, so even discoverability shall be taken seriously, both at documentation and repository organization levels.

I don't claim to have all answers right now, but these are the topics to consider to make this contribution part of mainline (as opposed to be maintained into its own fork, where of course such requirements can be relaxed).

Makefile
hzhuang1: I rechecked the Makefile. Maybe it could be implemented by `uname -p` command. The command result is in below.
| Command | Mac on x86 | Mac on arm | Linux on x86 | Linux on arm |
|---------------| ---------------: | ---------------: | -----------------: | ----------------: |
| uname -p  | i386             | arm              | x86_64           | aarch64        | 

And I need to make sure that Makefile is only used in MacOS/Linux/UNIX in this project. CMAKE is used for Windows OS. Is it right? If so, I could create a macro to distinguish architecture.

xxh_arm64dispatch.h
hzhuang1: There's a bit difference. In `xxh_x86dispatch.h`, the interfaces are defined with XXH_PUBLIC_API. I removed these for `xxh_arm64dispatch.h` since it would result in some build warnings.
```
clang -O3 -march=armv8-a+sve -fPIC -Wall -Wextra -Wstrict-aliasing=1 -std=c99     -DARM64_DISPATCH=1 -I../..     -c -o main.o main.c
In file included from main.c:36:
In file included from ./hashes.h:68:
./../../xxh_arm64dispatch.h:45:30: warning: function 'XXH3_64bits_dispatch' has internal linkage but is not defined [-Wundefined-internal]
XXH_PUBLIC_API XXH64_hash_t  XXH3_64bits_dispatch(const void* input, size_t len);
                             ^
./hashes.h:91:21: note: used here
    return (size_t) XXH3_64bits(src, srcSize);
                    ^
./../../xxh_arm64dispatch.h:65:22: note: expanded from macro 'XXH3_64bits'
# define XXH3_64bits XXH3_64bits_dispatch
                     ^
./../../xxh_arm64dispatch.h:50:30: warning: function 'XXH3_128bits_dispatch' has internal linkage but is not defined [-Wundefined-internal]
XXH_PUBLIC_API XXH128_hash_t XXH3_128bits_dispatch(const void* input, size_t len);
                             ^
./hashes.h:98:21: note: used here
    return (size_t) XXH3_128bits(src, srcSize).low64;
                    ^
./../../xxh_arm64dispatch.h:76:23: note: expanded from macro 'XXH3_128bits'
# define XXH3_128bits XXH3_128bits_dispatch
                      ^
2 warnings generated.
```
I'm not sure whether I could remove these for x86. So I just created a copy with a few differences.

xxh_arm64sve.S
hzhuang1: Does it mean that I could avoid executable stack flag if I don't include `asmdefs.h`?

Makefile
Cyan4973: Sounds like a plan

xxh_arm64dispatch.h
Cyan4973: It's not clear to me why there are these differences compared to the `x86` dispatcher.
Would be worth an investigation.

xxh_arm64sve.S
Cyan4973: > Does it mean that I could avoid executable stack flag if I don't include `asmdefs.h`?

No, the issue is about `xxh_arm64sve.S` itself

xxhash.h
hzhuang1: Thanks for catching this issue. I should define `XXH3_accumulate_512()` to `XXH3_accumulate_512_scalar()` for ARM64_DISPATCH case. Since it's only directly invoked for the last strip.

hzhuang1:
> Could you document, in the PR, the performance implications of this patch ?

Sure. I'll document it. The problem is that only performance of small data is recorded in `README.md`. My optimization is mainly focus on large block data. Which documentation should I use? or just fill them in the comments of the PR?

hzhuang1:
> I was wondering if this PR could be made smaller by separating 2 components :
> 
>     1. introduce a new `SVE` micro-op support (triggered through static selection only)
> 
>     2. then introduce a dynamic dispatcher for `arm64`
> 
OK. I'll split it into two patches.

Cyan4973:
> > Could you document, in the PR, the performance implications of this patch ?
> 
> Sure. I'll document it. The problem is that only performance of small data is recorded in `README.md`. My optimization is mainly focus on large block data. Which documentation should I use? or just fill them in the comments of the PR?

Yes, in the PR, that's fine

hzhuang1:
Close it since new pull request (https://github.com/Cyan4973/xxHash/pull/713) is created.

PR #708, opened 2022-05-04
Use inline assembler for Power/IBM XL Compiler

### LINKED ISSUE #707: xxHash fails to compile with IBM XL Compilers due to usage of clang builtin
The XXH_vec_mulo macro is defined for clang as __builtin_altivec_vmulouw, see https://github.com/Cyan4973/xxHash/blob/bcc4c6a0a2e25223fde0e6bb8a24946586e00771/xxhash.h#L3498.

However, this is non-standard, see page 151 of the Vector Intrinsics Programming Reference Specification (https://openpowerfoundation.org/specifications/vectorintrinsicprogrammingreference).

The IBM XL Compilers identifies itself as clang, but does not implement `__builtin_altivec_vmulouw`. Instead, it only supports `vec_mulo`. 

I think the easiest fix is to use the standardized vec_mulo if `__ibmxl__` is defined, and i will supply a PR for that in a moment. However, I am unsure whether this whole ifdef is still required with modern compilers...

MaxiBoether:
Fixes #707. With that fix, the IBM XL Compiler can successfully compile xxHash.

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -3489,13 +3489,13 @@ XXH_FORCE_INLINE xxh_u64x2 XXH_vec_loadu(const void *ptr)
  * These intrinsics weren't added until GCC 8, despite existing for a while,
  * and they are endian dependent. Also, their meaning swap depending on version.
  * */
-# if defined(__s390x__) || defined(__ibmxl__)
+# if defined(__s390x__) 
  /* s390x is always big endian, no issue on this platform */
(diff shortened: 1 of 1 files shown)


Cyan4973:
Looks good to me.
A second opinion from @easyaspi314 would be welcome.

easyaspi314:
Lgtm 👍

Cyan4973 (approved):

PR #713, opened 2022-05-09
Up sve 4

hzhuang1:
Implement ARM SVE optimization on xxHash. There're two implementations are included. One is instrinsic implementation that is the default implementation. The other is assembly implemenation that is enabled by setting "DISPATCH=1". The benchmark is in below.

    SCALAR implementation (default)
    benchmarking large inputs : from 512 bytes (log9) to 128 MB (log27)
    xxh3   ,  2587,  3015,  3152,  3296,  3383,  3425,  3449,  3425,  3431,  3457,  3467,  3467,  3466,  3254,  2400,  2151,  2134,  2116,  2119
    XXH32  ,  1498,  1524,  1537,  1544,  1543,  1546,  1548,  1536,  1491,  1490,  1493,  1493,  1490,  1441,  1242,  1186,  1181,  1184,  1188
    XXH64  ,  2630,  2874,  3017,  3092,  3131,  3142,  3154,  3130,  3013,  3020,  3021,  3017,  3017,  2870,  2034,  1834,  1853,  1854,  1842
    XXH128 ,  2353,  2711,  3057,  3244,  3352,  3409,  3443,  3435,  3431,  3441,  3396,  3441,  3431,  3224,  2299,  2125,  2145,  2164,  2192
    
    intrinsic SVE512 implementation with prefetch instruction
    benchmarking large inputs : from 512 bytes (log9) to 128 MB (log27)
    xxh3   ,  3684,  5967,  7804,  8945,  9866, 10406, 10691, 10656, 10854, 10925, 10922, 10949, 10960, 10029,  6347,  6034,  5958,  5991,  6059
    XXH32  ,  1326,  1440,  1495,  1523,  1534,  1543,  1544,  1535,  1505,  1507,  1508,  1506,  1507,  1460,  1257,  1199,  1189,  1192,  1198
    XXH64  ,  2510,  2803,  2976,  3067,  3120,  3138,  3155,  3130,  3051,  3058,  3056,  3063,  3055,  2932,  2163,  1990,  1962,  1971,  1981
    XXH128 ,  3421,  5798,  7500,  8881,  9790, 10348, 10669, 10592, 10855, 10968, 11021, 11029, 11020,  9988,  6156,  5821,  5937,  5994,  6079
    assembly SVE512 implementation with prefetch instruction
    benchmarking large inputs : from 512 bytes (log9) to 128 MB (log27)
    xxh3   ,  4040,  6610,  9690, 12158, 14054, 15201, 15891, 15506, 15384, 13975, 14045, 14059, 14070, 11838,  4572,  3953,  3997,  3999,  3990
    XXH32  ,  1321,  1439,  1494,  1521,  1534,  1541,  1546,  1534,  1502,  1505,  1493,  1504,  1505,  1444,  1249,  1195,  1195,  1191,  1195
    XXH64  ,  2502,  2799,  2975,  3070,  3115,  3136,  3152,  3127,  3040,  3038,  3051,  3052,  3036,  2857,  2142,  1975,  1980,  1980,  1979
    XXH128 ,  3842,  6341,  9479, 11898, 13938, 15005, 15819, 15245, 15278, 15505, 15630, 15668, 15712, 12437,  4912,  4148,  4020,  4028,  4010


Cyan4973:
Thanks for this updated PR,
and thanks for presenting benchmark numbers,
they are important for the rationale of importing this improvement.

Now, in previous PR, I suggested to split the new PR in multiple smaller ones.
I see you have split the PR into multiple commits, though they still form a single PR.

The benefit of having multiple PRs is to be able to merge them more gradually, as opposed to all-or-nothing.
There are multiple topics in this PR, each one with its own points of attention.
I believe we can be more focused if we take care of them one by one.

So, essentially, create new PR from intermediate commits.
(I might be able to do it myself from this PR, I'll have to look around.)

Here are the proposed milestones : 
1) Change the dispatch point from `XXH3_accumulate_512` to `XXH3_accumulate`, as it seems important for `SVE` speed. When done first, it's actually impacting all other variants, since SVE is not inserted yet. This is a good opportunity to check that performance for other platforms, especially `x64` and its vector variants, is either neutral or improved thanks to this change. And a good opportunity to evaluate the maintenance complexity of this stage. My understanding is that such step would supersede #692.
2) Introduce `SVE` using intrinsics, as a new vector variant, selected manually.
What will matter in this case is to **test** the variant in CI, typically using `qemu`.
Looking around, it seems that [`qemu` supports `SVE` now](https://wiki.qemu.org/Features/ARM/SVE).
3) Introduce ability to automatically select `SVE` at compile time, through detection of feature macros.
Check this detection works well in a CI test, typically using `qemu` and `xxhsum`'s verbose welcome message.
4) Introduce `SVE` as _assembly_. You made a good point that the assembly version can be measurably faster in some cases (when data is already in cache), so it seems to be worth it.
However, since it also requires an additional `*.S` file, it changes the source layout, and the build recipes.
There is also a potential issue around the "executable stack bit-flag" that we will have to understand and tame.
In short, integrating this variant cleanly is more complex than it seems.
5) Introduce `arm64dispatch`. This will add a few more questions around cross-platform API stability for users, automated selection, clean build, relation with the dynamic library, etc. It's a whole topic.

I think it doesn't change the end goal. All the elements present in this PR should end up being merged. But it allows us to be more focused at each stage.

hzhuang1:
> 
> I think it doesn't change the end goal. All the elements present in this PR should end up being merged. But it allows us to be more focused at each stage.

I agree that we need to work it out with milestones. Let's focus on stage 1 (#692) first. Currently this patch set isn't merged because there's a build issue if macro doesn't contain parameter. I figured out a workaround to avoid this case. What's our plan on stage 1?

hzhuang1:
It's replaced by #748. So close it now.

Issue #717, opened 2022-06-07
XXH_FALLTHROUGH macros generates warning with ICC compilation


rohanaggarwal7997:
Compiling with ICC, leads to evaluation of XXH_FALLTHROUGH as empty string and we get the following warning message 

xxhash.h(2344): warning #169: expected a declaration
             case 11:      XXH_PROCESS4; XXH_FALLTHROUGH

Cyan4973:
I presume that `XXH_FALLTHROUGH` definition reaches this line :  
https://github.com/Cyan4973/xxHash/blame/dev/xxhash.h#L743

It's questionable if this macro definition does anything. I wouldn't expect it to transport the comment statement at the target place.

What does ICC require for explicit fall-through ?
And also, which version of ICC are you using ?

Cyan4973:
Could you try to patch in #718 ?
It might help for `icc`, at least temporarily.

PR #718, opened 2022-06-07
blind fix for fallthrough on icc

Cyan4973:
answering #717


t-mat:
Intel has a public repo which contains `icc` setup for GA.
https://github.com/oneapi-src/oneapi-ci

I'll try it later.

PR #723, opened 2022-06-15
Fix double prefix when building with cmake

ilya-fedin:


ilya-fedin:
@Cyan4973 ping

cmake_unofficial/CMakeLists.txt
```diff
@@ -167,8 +167,8 @@ if(NOT XXHASH_BUNDLED_MODE)
   # configure and install pkg-config
   set(PREFIX ${CMAKE_INSTALL_PREFIX})
   set(EXECPREFIX "\${prefix}")
-  set(INCLUDEDIR "${CMAKE_INSTALL_INCLUDEDIR}")
-  set(LIBDIR "${CMAKE_INSTALL_LIBDIR}")
+  set(INCLUDEDIR "${CMAKE_INSTALL_FULL_INCLUDEDIR}")
```
tristan957: This is not right. You seem to be setting an absolute path. It needs to be relative to the `prefix`

ilya-fedin: cmake allows the user to set absolute path to CMAKE_INSTALL_LIBDIR (what is used in a couple of distros), so assuming the path will always be relative to prefix is not reliable with cmake.

tristan957: @ilya-fedin look at the cmake changes here https://github.com/pybind/pybind11/pull/4077/files

tristan957: or here https://github.com/HdrHistogram/HdrHistogram_c/blob/main/CMakeLists.txt

tristan957: you can pretty easily do this to support both ways. cc @eli-schwartz since you have have submitted a few PRs in this arena

ilya-fedin: :thinking: is there any difference though? It will be the same string after all while absolute path allows just to use the builtin variable existing in cmake for a long time.

eli-schwartz: > cmake allows the user to set absolute path to CMAKE_INSTALL_LIBDIR (what is used in a couple of distros), so assuming the path will always be relative to prefix is not reliable with cmake.

This is why you MUST use a smart path-joining function that takes CMAKE_INSTALL_LIBDIR without "FULL", checks if it is an absolute path, and:
- if absolute, uses it as-is to create an inferior pkg-config file
- if relative, joins it with `${prefix}`

eli-schwartz: In short, please use @jtojnar's https://github.com/jtojnar/cmake-snips/

As is in fact recommended for NixOS. :p

eli-schwartz: > :thinking: is there any difference though? It will be the same string after all while absolute path allows just to use the builtin variable existing in cmake for a long time.

That is broken in various situations, please don't break it. pkg-config needs to know the installation prefix in various cases more complicated than just running `pkgconf --cflags --libs libfoobar` on a Linux (and only Linux) system where the software is installed as a distro package.

It's the same reason that Meson handles GNU directory variables specially (guaranteeing they are always relativized even when defined as full paths if that full path is prefix-relative, providing `join_paths()`) and has a builtin pkg-config generator (builds variables for you, supports relocatable pkg-config files).

ilya-fedin: Thanks for explaining, done

t-mat:
@ilya-fedin, I'm sorry about these annoying errors!
#736 will fix this issue.

cmake_unofficial/CMakeLists.txt
eli-schwartz: Great, thanks for the update. :)

eli-schwartz (approved):
From a quick eyeball this change looks correct to me and properly handles the various edge cases under cmake.

Cyan4973:
I'm trying to understand what was the objective of this PR.

The title mentions "Fix double prefix when building with cmake",
which implies there's an issue when building with `cmake`.
It's probably not a trivial scenario, since I could not detect any issue from basic `cmake` tests.

If that's the case, I would expect it should be possible to create a test case that reproduces this issue. Such test case could then be integrated into CI.

Looking at the PR's content, it appears focused on pkg-config generation.
More specifically `includedir` and `libdir` are now set entirely from the build system, either `make` or `cmake`, instead of being derived from `prefix` within the pkgconfig file. The new design seems superficially similar to the original one, but I presume it differs in subtle ways, in order to fix some edge cases that were not properly covered.
And since the new design is also a bit more complex, which is a cost to maintenance, I would like to understand precisely what it does better.

Note : I see that one test was failing, but my understanding is that the problem was on the test side, not the PR. We can ignore it.

ilya-fedin:
Try to build xxHash with the following cmake arguments:
```
cmake -Bbuild -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_INCLUDEDIR=/usr/include -DCMAKE_INSTALL_LIBDIR=/usr/lib
```
This is valid for cmake but xxHash's build scripts generate pkg-config file in a way it gets `/usr/usr/include`, `/usr/usr/lib`

Cyan4973:
OK !
Following the proposed test,
I get the following content in `libxxhash.pc` : 
```
prefix=/usr
exec_prefix=${prefix}
includedir=${prefix}//usr/include
libdir=${exec_prefix}//usr/lib
```
which is obviously incorrect, and will lead to duplicated prefix `/usr`.

In contrast, after your patch, the content of `libxxhash.pc` becomes : 
```
prefix=/usr
exec_prefix=${prefix}
includedir=/usr/include
libdir=/usr/lib
```

It's probably worth making a test from this example.

ilya-fedin:
Sorry, but I don't have that much interest to create a test. If this is required to get the PR merged, I'd just close it and let someone else to do the work.

Cyan4973:
Don't worry, I'm taking care of it (#845).

PR #726, opened 2022-07-09
Remove stream loads and slightly improve avx512 seed generation

goldsteinn:
Streamed loads (`vmovntdqa`) don't provide any benefit over normal
aligned loads (`vmovdqa{8|16|32|64}`) but prevent GCC/Clang from
micro-fusing the loads with the ALU wasting code size.

Replace lane cross `vpbroadcastq` with vpsubq`

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -4179,7 +4179,8 @@ XXH3_initCustomSecret_avx512(void* XXH_RESTRICT customSecret, xxh_u64 seed64)
     XXH_ASSERT(((size_t)customSecret & 63) == 0);
     (void)(&XXH_writeLE64);
     {   int const nbRounds = XXH_SECRET_DEFAULT_SIZE / sizeof(__m512i);
-        __m512i const seed = _mm512_mask_set1_epi64(_mm512_set1_epi64((xxh_i64)seed64), 0xAA, (xxh_i64)(0U - seed64));
+        __m512i const seed_pos = _mm512_set1_epi64(seed64);
+        __m512i const seed     = _mm512_mask_sub_epi64(seed_pos, 0xAA, _mm512_set1_epi8(0), seed_pos);
 
         const __m512i* const src  = (const __m512i*) ((const void*) XXH3_kSecret);
               __m512i* const dest = (      __m512i*) customSecret;
@@ -4187,14 +4188,7 @@ XXH3_initCustomSecret_avx512(void* XXH_RESTRICT customSecret, xxh_u64 seed64)
         XXH_ASSERT(((size_t)src & 63) == 0); /* control alignment */
         XXH_ASSERT(((size_t)dest & 63) == 0);
         for (i=0; i < nbRounds; ++i) {
-            /* GCC has a bug, _mm512_stream_load_si512 accepts 'void*', not 'void const*',
-             * this will warn "discards 'const' qualifier". */
-            union {
-                const __m512i* cp;
-                void* p;
-            } remote_const_void;
-            remote_const_void.cp = src + i;
-            dest[i] = _mm512_add_epi64(_mm512_stream_load_si512(remote_const_void.p), seed);
+            dest[i] = _mm512_add_epi64(_mm512_load_si512(src + i), seed);
     }   }
 }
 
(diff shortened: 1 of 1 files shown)


Cyan4973:
I presume this change is expected to bring some improvements, either speed or binary size ?
Is there any way to measure such benefit ?

Also : `_mm256_stream_load_si256` was brought by @gzm55 as part of #351 . @gzm55: do you remember any specific reason to select this intrinsic variant, instead of the now proposed `_mm256_load_si256` ?

goldsteinn:
> I presume this change is expected to bring some improvements, either speed or binary size ? Is there any way to measure such benefit ?



Size:
`-O3 -march=x86-64-v3`: Saves 124 bytes of `.text`
`-O3 -march=x86-64-v4`: Saves 88 bytes of `.text. 

Perf:

Intuitively there is no difference. The actual backend change (mask generation) is not in the loop.

`micro-fusion` doesn't save any backend uops. On some x86-64 machine it might be the difference
makers for a loop fitting in the DSB/LSD, but that's really not something you can control at the `C` level. 
The best I can say is it does give the compiler the flexibility to make those optimization as based on its 
model of the FE it can decide whether `micro-fusion` is worth it (although I don't know of any compiler 
that does that). Maybe it was save an icache miss occasionally but that's a stretch.



> 
> Also : `_mm256_stream_load_si256` was brought by @gzm55 as part of #351 . @gzm55: do you remember any specific reason to select this intrinsic variant, instead of the now proposed `_mm256_load_si256` ?

Cyan4973 (approved):

gzm55:
> I presume this change is expected to bring some improvements, either speed or binary size ? Is there any way to measure such benefit ?
> 
> Also : `_mm256_stream_load_si256` was brought by @gzm55 as part of #351 . @gzm55: do you remember any specific reason to select this intrinsic variant, instead of the now proposed `_mm256_load_si256` ?

Logically two serials of load instructions are the same. `*_stream_*` intrinsics try to load with  non-temporal memory. When choosing load instructions, `micro-fusion` is not considered much, and these are the main reasons i can remember:
- the loaded data here should be read and process once, and
- the data will rarely be cached before load, so the non-temporal memory should be suiltable
- `_mm512_stream_load_si512` has a smaller latency than `_mm512_load_si512 ` from https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html
- many time benchmarks for avx2 and avx512, and the  `*_stream_*` intrinsics seams to be more stable (now the data were lost)

goldsteinn:
> > I presume this change is expected to bring some improvements, either speed or binary size ? Is there any way to measure such benefit ?
> > Also : `_mm256_stream_load_si256` was brought by @gzm55 as part of #351 . @gzm55: do you remember any specific reason to select this intrinsic variant, instead of the now proposed `_mm256_load_si256` ?
> 
> Logically two serials of load instructions are the same. `*_stream_*` intrinsics try to load with non-temporal memory. When choosing load instructions, `micro-fusion` is not considered much, and these are the main reasons i can remember:
> 
> * the loaded data here should be read and process once, and
> * the data will rarely be cached before load, so the non-temporal memory should be suiltable

AFAIK this is only potentially meaningful for non-cacheable memory.
See: https://stackoverflow.com/questions/32103968/non-temporal-loads-and-the-hardware-prefetcher-do-they-work-together
and: https://stackoverflow.com/questions/35516878/acquire-release-semantics-with-non-temporal-stores-on-x64

> * `_mm512_stream_load_si512` has a smaller latency than `_mm512_load_si512 ` from https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html

This is an error in the intrinsics guides' documentation.

See https://uops.info/ for a generally more reliable source on these things.

Seems to also be an extra `uop` for the NT hint.


> * many time benchmarks for avx2 and avx512, and the  `*_stream_*` intrinsics seams to be more stable (now the data were lost)

gzm55:
> > > I presume this change is expected to bring some improvements, either speed or binary size ? Is there any way to measure such benefit ?
> > > Also : `_mm256_stream_load_si256` was brought by @gzm55 as part of #351 . @gzm55: do you remember any specific reason to select this intrinsic variant, instead of the now proposed `_mm256_load_si256` ?
> > 
> > 
> > Logically two serials of load instructions are the same. `*_stream_*` intrinsics try to load with non-temporal memory. When choosing load instructions, `micro-fusion` is not considered much, and these are the main reasons i can remember:
> > 
> > * the loaded data here should be read and process once, and
> > * the data will rarely be cached before load, so the non-temporal memory should be suiltable
> 
> AFAIK this is only potentially meaningful for non-cacheable memory. See: https://stackoverflow.com/questions/32103968/non-temporal-loads-and-the-hardware-prefetcher-do-they-work-together and: https://stackoverflow.com/questions/35516878/acquire-release-semantics-with-non-temporal-stores-on-x64
> 
> > * `_mm512_stream_load_si512` has a smaller latency than `_mm512_load_si512 ` from https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html
> 
> This is an error in the intrinsics guides' documentation.
> 
> See https://uops.info/ for a generally more reliable source on these things.
> 
> Seems to also be an extra `uop` for the NT hint.
> 
> > * many time benchmarks for avx2 and avx512, and the  `*_stream_*` intrinsics seams to be more stable (now the data were lost)

@goldsteinn thanks for your explantation

PR #729, opened 2022-07-29
Fix: brace expansion

t-mat:
Currently, `make trailingWhitespace` fails with the following error (but `!` passes it).

```
$ make trailingWhitespace
! grep -E "`printf '[ \\t]$'`" cli/*.{c,h,1} *.c *.h LICENSE Makefile cmake_unofficial/CMakeLists.txt
grep: cli/*.{c,h,1}: No such file or directory
```

This error is caused by `cli/*.{c,h,1}`, because GNU `make` invokes `shell` which doesn't support brace expansion.

This PR fixes it with plain wildcards.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -433,7 +433,7 @@ listL120:  # extract lines >= 120 characters in *.{c,h}, by Takayuki Matsuoka (n
 
 .PHONY: trailingWhitespace
 trailingWhitespace:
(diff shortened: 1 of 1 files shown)


Cyan4973:
Great find ! Thanks for fixing this issue @t-mat !

Cyan4973 (approved):

PR #730, opened 2022-07-29
Fix issue #724

### LINKED ISSUE #724: xxh*sum aren't Behaving Like sha*sum on Checking
I.g. sha512sum will not check sha256sum's checksums, it'll output an error. xxHash checks all, not following Coreutils' behaviour.

Please would it be possible for "xxh64sum" to reject "xxh128sum" checksums and vice versa? For example, a symbolic link or the "-H" option modifies the behaviour, but "xxhsum" called alone behaves broadly like usual.

The step additionally might help with #645 in some way.

t-mat:
This PR fixes #724.

(1) Introduces new `enum AlgoBitmask` which represents xxh algorithm(s) as a bitmask(s).  With this type, each `xxhsum` symlink variant indicates their acceptance.
(2) Adds all possible 4x4 test matrix to the `Makefile`.  *4* means `{ xxhsum, xxh32sum, xxh64sum, xxh128sum }`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -427,6 +427,33 @@ test-tools:
 	CFLAGS=-Werror $(MAKE) -C tests/bench
 	CFLAGS=-Werror $(MAKE) -C tests/collisions
 
+.PHONY: test-xxh-nnn-sums
+test-xxh-nnn-sums: xxhsum_and_links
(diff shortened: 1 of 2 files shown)


Cyan4973 (approved):

Cyan4973:
Thanks @t-mat ! That's a great addition !

PR #731, opened 2022-08-22
Minor optimizations for medium / large hashes.

goldsteinn:


goldsteinn:
The test failure seems to be because of the changed hash result. I think this is expected because the loop count change but someone should double check.

Cyan4973:
Changing the value of the hash will indeed fail the tests.

goldsteinn:
> Changing the value of the hash will indeed fail the tests.

Hi, can you comment on whether its an issue in this case?

I believe the hash value changes are because the loop bound was changed
to avoid rehashing **duplicate** bytes in the second commit.

Cyan4973:
> Hi, can you comment on whether its an issue in this case?

It is. `xxHash` is labelled "stable" for a reason : so the produced hashes can be stored and compared later. Therefore, any change of value is a big problem.

goldsteinn:
> > Hi, can you comment on whether its an issue in this case?
> 
> It is. `xxHash` is labelled "stable" for a reason : so the produced hashes can be stored and compared later. Therefore, any change of value is a big problem.

I see. Are you open to a patch that puts this behind a preprocessor? Something like `ifdef XXH_UNSTABLE`? Think for many users we don't need stability (or can accept rehashing) so would prefer unstable changes for the sake of perf.

Cyan4973:
I'm concerned about the confusion to the ecosystem that such a double-allegiance would create.
I suspect that a fork, featuring a clearly named difference, would be more appropriate for this project.

goldsteinn:
@Cyan4973 I'm going to start maintaining a fork [xxHash-Unstable](https://github.com/goldsteinn/xxHash-Unstable). Aside from the inclusive calculating of `nbRounds` in the `129to240` case, do you know of any other perf gains that have been dropped / discarded because they where unstable?

Issue #733, opened 2022-08-31
Potential AArch64 Optimization


dougallj:
Currently AArch64 uses `XTN` and `SHRN` (via `XXH_SPLIT_IN_PLACE`) once every two lanes:

https://github.com/Cyan4973/xxHash/blob/c4359b17db082888fdc18371eba918b957a6baaa/xxhash.h#L4456-L4466

This is a total of 6 SIMD operations to accumulate two lanes.

By unrolling the loop once, to do four lanes per iteration, I think both copies of the `XTN` and `SHRN` operations could be replaced with a single `UZP1` and `UZP2` per iteration (and `UMLAL2` for the second multiply-accumulate). This would save 2 SIMD operations (out of 12), every four lanes.

(As there are only six SIMD lanes, this can only save 2 of 18 SIMD operations per accumulate.)

dougallj:
Oh, and clang seems to be turning this into a ~5c latency chain, rather the intended 2c latency chain, which may make this latency bound on M1:
```
ADD   V1.2D, V6.2D, V1.2D
UMLAL V1.2D, V5.2S, V4.2S
```
Adding an `__asm__("" : "+w"(acc_vec_2));` before the `acc_vec = vaddq_u64 (acc_vec, acc_vec_2);` might force it to not do that.

PR #734, opened 2022-08-31
Optimize XXH3_accumulate_512_neon

dougallj:
Curiosity got the better of me, so I made the changes described in #733.

In a custom benchmark on M1, those are ~4% speedup (34.6GB/s -> 36.1GB/s). Bumping `XXH3_NEON_LANES` to 8 reaches 37.3GB/s for a ~7.7% speedup overall.

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -1840,6 +1840,12 @@ static void* XXH_memcpy(void* dest, const void* src, size_t size)
 #  define XXH_COMPILER_GUARD(var) ((void)0)
 #endif
 
+#if defined(__GNUC__) || defined(__clang__)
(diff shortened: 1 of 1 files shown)


Cyan4973:
cc @easyaspi314 

I can confirm the performance gains on M1.

easyaspi314:
Slight boost on my Pixel 4a (10.4->10.8)

I get 10.9 (no clue why) if I do
```c
uint32x4x2_t zipped = vuzpq_u32(vreinterpretq_u32_u64(data_key1), vreinterpretq_u32_u64(data_key2));
uint32x4_t data_key_lo = zipped.val[0];
uint32x4_t data_key_hi = zipped.val[1];
```
Plus this also makes this block compatible with armv7-a.

Cyan4973 (approved):

PR #736, opened 2022-09-02
Remove macOS-10.15 from ci.yml

t-mat:
This PR fixes the following error

image

See also:
- Actual error in CI : https://github.com/Cyan4973/xxHash/actions/runs/2979849990
- Announce from GitHub: https://github.com/actions/runner-images/issues/5583

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -407,7 +407,7 @@ jobs:
         LDFLAGS="-static" CC=$XCC RUN_ENV=$XEMU make clean check
 
 
-  # macOS, { 10.15, 11 }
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks @t-mat ,
that's fine, we still have `macos-11` in CI tests.

Maybe we should also consider adding `macos-12` tests since it's the current version,
though I don't know if it's supported by Github Actions.

t-mat:
Yes, it seems `macos-12` is out of beta.

https://github.blog/changelog/2022-06-13-github-actions-macos-12-for-github-hosted-runners-is-now-generally-available/

PR #740, opened 2022-09-11
Fix xxhash.com address

mykaul:
Should be https://cyan4973.github.io/xxHash/ I reckon.

### DIFF
--- a/doc/xxhash_spec.md
+++ b/doc/xxhash_spec.md
@@ -339,7 +339,7 @@ On 32-bit systems though, positions are reversed: `XXH64` performance is reduced
 Reference Implementation
 ----------------------------------------
 
-A reference library written in C is available at https://www.xxhash.com.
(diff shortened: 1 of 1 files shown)


Cyan4973:
https://cyan4973.github.io/xxHash/ is the correct value today.
But the point of having an umbrella URL is that its underlying representation can change overtime.

mykaul:
> https://cyan4973.github.io/xxHash/ is the correct value today. But the point of having an umbrella URL is that its underlying representation can change overtime.

Understood - but right now I'm getting 404 for it - perhaps it should be http://www.xxhash.com ?

Cyan4973:
Indeed,
https redirection seems broken again. :(

Yes, `http://` redirection works better.
The `https` one is more finicky ...

Cyan4973:
`https://www.xxhash.com` is now accessible again

PR #746, opened 2022-10-19
fix issues reported by cppcheck

hzhuang1:
Run command "make clean cppcheck".

---- static analyzer - cppcheck ----
cppcheck . --force --enable=warning,portability,performance,style --error-exitcode=1 > /dev/null cli/xsum_os_specific.c:220:9: style: The scope of the variable 'ret' can be reduced. [variableScope]
    int ret;
        ^
cli/xsum_sanity_check.c:571:19: style: Condition 'len>0' is always true [knownConditionTrueFalse]
    } else if (len>0) {
                  ^
cli/xsum_sanity_check.c:569:13: note: Assuming that condition 'len==0' is not redundant
    if (len == 0) {
            ^
cli/xsum_sanity_check.c:571:19: note: Condition 'len>0' is always true
    } else if (len>0) {
                  ^
tests/bench/benchHash.c:112:55: error: Uninitialized variables: result.internal_never_ever_use_directly, result.error_result_never_ever_use_directly, result.error_tag_never_ever_use_directly [uninitvar]
    BMK_runTime_t const runTime = BMK_extract_runTime(result);
                                                      ^
tests/bench/benchHash.c:107:12: note: Assuming condition is false
    while (!BMK_isCompleted_TimedFn(txf)) {
           ^
tests/bench/benchHash.c:107:12: note: Assuming condition is false
    while (!BMK_isCompleted_TimedFn(txf)) {
           ^
tests/bench/benchHash.c:107:12: note: Assuming condition is false
    while (!BMK_isCompleted_TimedFn(txf)) {
           ^
tests/bench/benchHash.c:112:55: note: Uninitialized variables: result.internal_never_ever_use_directly, result.error_result_never_ever_use_directly, result.error_tag_never_ever_use_directly
    BMK_runTime_t const runTime = BMK_extract_runTime(result);
                                                      ^
tests/bench/benchfn.c:95:12: error: Uninitialized variable: outcome.error_result_never_ever_use_directly [uninitvar]
    return outcome;
           ^
tests/collisions/main.c:209:14: style: Same expression on both sides of '<='. [duplicateExpression]
    assert(m <= m);
             ^
make: *** [Makefile:387: cppcheck] Error 1

### DIFF
--- a/cli/xsum_os_specific.c
+++ b/cli/xsum_os_specific.c
@@ -217,7 +217,6 @@ static int XSUM_stat(const char* infilename, XSUM_stat_t* statbuf)
 XSUM_ATTRIBUTE((__format__(__printf__, 2, 0)))
 static int XSUM_vasprintf(char** strp, const char* format, va_list ap)
 {
-    int ret;
     int size;
     va_list copy;
     /*
@@ -235,6 +234,7 @@ static int XSUM_vasprintf(char** strp, const char* format, va_list ap)
         *strp = NULL;
         return size;
     } else {
+        int ret;
         *strp = (char*) malloc((size_t)size + 1);
         if (*strp == NULL) {
             return -1;
--- a/tests/collisions/main.c
+++ b/tests/collisions/main.c
@@ -206,7 +206,7 @@ static double Cnm(int n, int m)
 {
     assert(n > 0);
     assert(m > 0);
-    assert(m <= m);
+    assert(m <= n);
     double acc = 1;
     for (int i=0; i<m; i++) {
         acc *= n - i;
--- a/cli/xsum_sanity_check.c
+++ b/cli/xsum_sanity_check.c
@@ -568,7 +568,7 @@ static void XSUM_testXXH128_withSecret(const void* data, const void* secret, siz
     XXH128_hash_t Nresult = testData->Nresult;
     if (len == 0) {
         data = NULL;
-    } else if (len>0) {
+    } else {
         assert(data != NULL);
     }
     {   XXH128_hash_t const Dresult = XXH3_128bits_withSecret(data, len, secret, secretSize);
--- a/tests/bench/benchHash.c
+++ b/tests/bench/benchHash.c
@@ -28,6 +28,7 @@
 
 #include <stdlib.h>   // malloc
 #include <assert.h>
+#include <string.h>
 
 #include "benchHash.h"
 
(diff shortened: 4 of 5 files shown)


cli/xsum_os_specific.c
```diff
@@ -235,6 +234,7 @@ static int XSUM_vasprintf(char** strp, const char* format, va_list ap)
         *strp = NULL;
         return size;
     } else {
+        int ret;
```
Cyan4973, line 237: 👍

tests/collisions/main.c
```diff
@@ -206,7 +206,6 @@ static double Cnm(int n, int m)
 {
     assert(n > 0);
     assert(m > 0);
-    assert(m <= m);
```
Cyan4973, line 209: I think the correct fix here is : 
```
assert(m <= n);
```

cli/xxhsum.c
```diff
@@ -59,7 +59,8 @@
 
 static unsigned XSUM_isLittleEndian(void)
 {
-    const union { XSUM_U32 u; XSUM_U8 c[4]; } one = { 1 };   /* don't use static: performance detrimental  */
+    union { XSUM_U32 u; XSUM_U8 c[4]; } one;   /* don't use static: performance detrimental  */
```
Cyan4973: You might want to check that the compiler still generates the same code for that variant.
Not concerned about correctness : it should be fine in both cases.
But we want to be sure that performance is also not impacted.

hzhuang1:
OK. Let me check the instructions.

On 19 Oct 2022, at 14:38, Yann Collet ***@***.******@***.***>> wrote:



@Cyan4973 commented on this pull request.

________________________________

In cli/xxhsum.c<https://github.com/Cyan4973/xxHash/pull/746#discussion_r998997881>:

> @@ -59,7 +59,8 @@

 static unsigned XSUM_isLittleEndian(void)
 {
-    const union { XSUM_U32 u; XSUM_U8 c[4]; } one = { 1 };   /* don't use static: performance detrimental  */
+    union { XSUM_U32 u; XSUM_U8 c[4]; } one;   /* don't use static: performance detrimental  */


You might want to check that the compiler still generates the same code for that variant.
Not concerned about correctness : it should be fine in both cases.
But we want to be sure that performance is also not impacted.

Cyan4973:
> OK. Let me check the instructions.

oh, nervermind.
I initially reacted thinking this endian detection function was in `xxhash.c`,
but it's in `xxhsum.c`,
and where it's employed, it's not performance critical.
So performance is not an issue.

I confused it for `XXH_isLittleEndian()`,
which is way more performance critical.

Interestingly, note that both `XXH_isLittleEndian()` and `XSUM_isLittleEndian()` are identical.
I presume this change is meant to please `cppcheck` static analyzer.
Since I don't see an equivalent change on `XXH_isLittleEndian()`, I presume it did not flag anything there.
It's strange that it seems to complain about `XSUM_isLittleEndian()`, but not `XXH_isLittleEndian()`,
given that both functions are identical...

hzhuang1:
Let me remove the patch for unusedStructMember first.

Cyan4973 (approved):

Issue #675, opened 2022-01-16
Update xxhash_spec.md with the details for XXH3 ?


palladium-coder:
Hello,

Can xxhash_spec.md doc be updated to include the details of XXH3 algorithm . It will help the folks who are implementing it from scratch.

Cyan4973:
Note : this is a `v1.0` release topic.

Also : partially redundant with #589.

t-mat:
Since #589 has been closed, we can also close this issue via #750.  @Cyan4973

PR #750, opened 2022-10-24
Add algorithm specification of XXH3

### LINKED ISSUE #589: Specify the 128-bit variant
As things are, a public domain implementation would require a third party to [extract a specification](https://en.wikipedia.org/wiki/Clean_room_design) from the current code.

In #479 it has been mentioned that we can't expect this anytime soon.

I'm creating a standalone issue, because that one is a "question".

adrien1018:
Update `xxhash_spec.md` to include XXH3 specification. Closes #589 and #675.

### DIFF
--- a/doc/xxhash_spec.md
+++ b/doc/xxhash_spec.md
@@ -24,6 +24,10 @@ Table of Contents
 - [Introduction](#introduction)
 - [XXH32 algorithm description](#xxh32-algorithm-description)
 - [XXH64 algorithm description](#xxh64-algorithm-description)
+- [XXH3 algorithm description](#xxh3-algorithm-overview)
+   - [Small inputs](#xxh3-algorithm-description-for-small-inputs)
+   - [Medium inputs](#xxh3-algorithm-description-for-medium-inputs)
+   - [Large inputs](#xxh3-algorithm-description-for-large-inputs)
 - [Performance considerations](#performance-considerations)
 - [Reference Implementation](#reference-implementation)
 
@@ -232,7 +236,7 @@ Each lane read its associated 64-bit value using __little-endian__ convention.
 For each {lane, accumulator}, the update process is called a _round_, and applies the following formula:
 
 ```c
-  round(accN,laneN):
+round(accN,laneN):
   accN = accN + (laneN * PRIME64_2);
   accN = accN <<< 31;
   return accN * PRIME64_1;
@@ -250,7 +254,7 @@ All 4 lane accumulators from previous steps are merged to produce a single remai
 Note that accumulator convergence is more complex than 32-bit variant, and requires to define another function called _mergeAccumulator()_:
 
 ```c
-  mergeAccumulator(acc,accN):
+mergeAccumulator(acc,accN):
   acc  = acc xor round(0, accN);
   acc  = acc * PRIME64_1;
   return acc + PRIME64_4;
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks @adrien1018 ! This is great work !

I think this documentation is in good shape, just give me a bit more time to read it in detail.

doc/xxhash_spec.md
```diff
@@ -43,13 +47,18 @@ However, a given variant shall produce exactly the same output, irrespective of
 ### Operation notations
 
 All operations are performed modulo {32,64} bits. Arithmetic overflows are expected.
-`XXH32` uses 32-bit modular operations. `XXH64` uses 64-bit modular operations.
+`XXH32` uses 32-bit modular operations. `XXH64` and `XXH3` uses 64-bit modular operations.
```
Cyan4973, line 50: minor nit: use_ 64-bit modular (plural)

doc/xxhash_spec.md
```diff
@@ -43,13 +47,18 @@ However, a given variant shall produce exactly the same output, irrespective of
...
 
 - `+`: denotes modular addition
+- `-`: denotes modular subtraction
 - `*`: denotes modular multiplication
+    - **Exception:** In `XXH3`, if it is in the form `(u128)x * (u128)y`, it denotes 64-bit by 64-bit normal multiplication into a full 128-bit result.
```
Cyan4973, line 55: 👍 
It's interesting, the width of the modular arithmetic may be not obvious, making `*` alone possibly insufficient to describe the operation.

doc/xxhash_spec.md
```diff
@@ -324,6 +333,461 @@ The `XXH64()` function produces an unsigned 64-bit value as output.
...
+  static const u64 PRIME32_1 = 0x9E3779B1U;  // 0b10011110001101110111100110110001
+  static const u64 PRIME32_2 = 0x85EBCA77U;  // 0b10000101111010111100101001110111
+  static const u64 PRIME32_3 = 0xC2B2AE3DU;  // 0b11000010101100101010111000111101
+  static const u64 PRIME64_1 = 0x9E3779B185EBCA87ULL;  // 0b1001111000110111011110011011000110000101111010111100101010000111
+  static const u64 PRIME64_2 = 0xC2B2AE3D27D4EB4FULL;  // 0b1100001010110010101011100011110100100111110101001110101101001111
+  static const u64 PRIME64_3 = 0x165667B19E3779F9ULL;  // 0b0001011001010110011001111011000110011110001101110111100111111001
+  static const u64 PRIME64_4 = 0x85EBCA77C2B2AE63ULL;  // 0b1000010111101011110010100111011111000010101100101010111001100011
+  static const u64 PRIME64_5 = 0x27D4EB2F165667C5ULL;  // 0b0010011111010100111010110010111100010110010101100110011111000101
+  static const u64 PRIME_MX1 = 0x165667919E3779F9ULL;  // 0b0001011001010110011001111001000110011110001101110111100111111001
+  static const u64 PRIME_MX2 = 0x9FB21C651E98DF25ULL;  // 0b1001111110110010000111000110010100011110100110001101111100100101
```
Cyan4973, line 355: kudos for giving this constant a name

doc/xxhash_spec.md
```diff
@@ -324,6 +333,461 @@ The `XXH64()` function produces an unsigned 64-bit value as output.
...
+### Step 2. Process the input
+
+This step is further divided into two cases: one for 17-128 bytes of input, and one for 129-240 bytes of input.
+
+#### Mixing operation
+
+This step uses a mixing operation that mixes a 16-byte segment of data, a 16-byte segment of secret and the seed into a 64-bit value as a building block. This operation treat the segment of data and secret as little-endian 64-bit values.
+
+```c
+mixStep(u8 data[16], size secretOffset, u64 seed):
```
Cyan4973, line 576: minor: this interface likely miss the `secret` pointer parameter,
the following code would only make sense if `secret` was a global variable only.

doc/xxhash_spec.md
```diff
@@ -324,6 +333,461 @@ The `XXH64()` function produces an unsigned 64-bit value as output.
...
+
+```c
+u64 acc[8] = {
+  PRIME32_3, PRIME64_1, PRIME64_2, PRIME64_3,
+  PRIME64_4, PRIME32_2, PRIME64_5, PRIME32_1};
+```
+
+### Step 2. Process blocks
+
+The input is consumed and processed one full block at a time. The size of the block depends on the length of the secret. Specifically, a block consists of several 64-byte stripes. The number of stripes per block is `floor((secretLength-64)/8)` . For the default 192-byte secret, there are 16 stripes in a block, and thus the block size is 1024 bytes.
```
Cyan4973, line 684: 👍

Cyan4973 (approved):
This is great work.
I only noticed a couple of nits, don't worry I'll fix them directly in my branch.
I'm even likely to update the source in a couple of ways to better follows the text of this specification.

Thanks @adrien1018 !

doc/xxhash_spec.md
Cyan4973: After reviewing the rest of the specification, I believe that the propose notification for 128-bit modular arithmetic does the job correctly.

doc/xxhash_spec.md
Cyan4973: I realize that not providing `secret` as an input parameter is a good way to improve brevity, which is beneficial for readability.
I think I'll keep it this way,
I just want to make it clear that `secret` is not _necessarily_ the _standard_ `defaultSecret`, it can also be any blob of bytes set by the user.
There's already a paragraph about this point, which is probably clear enough.
Let's just ensure it.

PR #759, opened 2022-11-10
remove tipi build temporary

hzhuang1:
/usr/local/share/.tipi/w/cli-281f677/xsum_sanity_check.c:31:10: error: '../xxhash.h' file not found, did you mean 'xxhash.h'? 10081
10082
         ^~~~~~~~~~~~~
10083
         "xxhash.h"
10084
1 error generated.

Since there are always some tipi build issues, remove them temporary until it could be resolved.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -549,38 +549,38 @@ jobs:
           file ./xxhsum.exe | grep -q '80386' || $(exit 1)
           ./xxhsum.exe --version
 
-  tipi-build-linux:
-    name: tipi.build project build and dependency resolution
-    runs-on: ubuntu-latest
-    container: tipibuild/tipi-ubuntu
-
-    env:
-      HOME: /root
-
-    steps:
-      - uses: actions/checkout@v3
-        # FIX: we currently need a full clone
-        with:
-          fetch-depth: '0'
-      - run: |
-          mkdir -p /usr/local/share/.tipi
-          # FIX: Hack for github action
-          git config --global --add safe.directory /usr/local/share/.tipi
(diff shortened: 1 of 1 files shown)


pysco68:
Hey @hzhuang1 this is another bug that was triggered by a change in the mode of operation (basically now "clean and isolated" vs. what it was previously) of tipi.

Some explainations:

We have released a new automatic caching feature in tipi that, in order to guarantee consistency accross different machines etc, required us to implement some sort of "constant and canonical path to dependency sources". Basically we mirror the sources to a fixed and known location on the machine running the build and compile it there. We do so from the "project root" e.g. the working directory of the build.

Now, it happens that the CI integration I originally submitted included one build step to verify that users could depend on xxHash as a library in their own projects (the step currently failing), but I oversaw that the "xxHash library" includes were parent-relative (e.g. `#include "../xxhash.h"`).

While this was no issue initially (because we built "in-place"), with the source mirroring the path becomes unreachable (because now we compile somewhere in `/usr/local/share/.tipi/w/`). Hence the error message.

I added a fix that I will submit in a few minutes (given the CI run passes - https://github.com/tipi-deps/xxHash/actions/runs/3438511224) which creates a copy of the `./cli` folder and applies some SED magic during the tipi build CI job to simulated a "proper" library-like inclusion. 

Again, sorry a thousand times for the disruption. This one was really bad.

We'd like to make this square somehow, so if you want to get in touch with us we'll figure something out (@Cyan4973 - same for you, we're really really sorry about how this went)

Cyan4973:
There is small conflict to solve after merging #760
but I'm still supportive of this PR @hzhuang1 .

hzhuang1:
Close it since tipi build issue is fixed now.

PR #754, opened 2022-10-31
Small improvement to x86 vectorized hashes and medium-sizes hash.

goldsteinn:

PR #765, opened 2022-11-30
added macos-12 test to GH CI

Cyan4973:


Cyan4973:
The `macos-12` test employs `clang-14`, which seems able to reproduce the compilation error experienced while testing #754. 
So that's all good !

Issue #766, opened 2022-11-30
GH-Actions: QEMU and compiler matrix test are broken


t-mat:
Currently, almost all expression for platform detection is written with `==`

```
    - name: ARM64 (XXH_VECTOR=[ scalar, NEON, SVE ])
      if: ${{ matrix.name == 'ARM64' }}
```

But it must be replaced with `startsWith()`

```
    - name: ARM64 (XXH_VECTOR=[ scalar, NEON, SVE ])
      if: ${{ startsWith(matrix.name, 'ARM64,') }}
```

Because our `name` format is `PLATFORM, compiler`, first entry passes the condition but all other entries are skipped.

```
  qemu-consistency:
    strategy:
      matrix:
        include: [
          { name: 'ARM64',           ... },
          { name: 'ARM64, gcc-10',   ... },
          { name: 'ARM64, gcc-9',    ... },
```

t-mat:
First of all, this is my mistake :facepalm:.

I made potential patch for `ci.yml` : https://github.com/t-mat/xxHash/commit/e6bee1dd9fe9b7304f89b376b613883627ee331c
and here's its full test log https://github.com/t-mat/xxHash/actions/runs/3583516752/jobs/6029038872

We also need

- Investigate platform=S390X, XXH_VECTOR=VSX
- Investigate platform=ARM64, XXH_VECTOR=SVE

t-mat:
As for "S390X, VSX", some versions are failed, but other versions pass the test.

<img width="228" alt="image" src="https://user-images.githubusercontent.com/898396/204806896-707b6eea-a32d-4efd-82c3-a9a568247ceb.png">

For example, in above case, `gcc-8` pass the test.
- gcc-8 : https://github.com/t-mat/xxHash/actions/runs/3583516752/jobs/6029045125#step:10:303

But `gcc-10` failed
- gcc-10 https://github.com/t-mat/xxHash/actions/runs/3583516752/jobs/6029043494#step:10:303

Perhaps, QEMU and/or gcc may have issue.

t-mat:
## QEMU ARM64 : Summary

I think it's okay to disable ARM64/SVE test for old `gcc` and `QEMU`s.  Because they won't have further support or fix.

Errors:

<img width="212" alt="image" src="https://user-images.githubusercontent.com/898396/204812951-a0b692fe-76b1-4c89-99cc-9bf2fa27bdbf.png">
[...]

Issue #763, opened 2022-11-27
dispatch regression with recent commit


tycho:
Commit 91788f1677f5a8449ae57b825009781f4d9eec73 seems to have caused a regression with dispatch, causing it to unconditionally use AVX512 on my system -- while my toolchain supports AVX512, my host CPU does not.

My system hits an illegal instruction through here: https://github.com/Cyan4973/xxHash/blob/91788f1677f5a8449ae57b825009781f4d9eec73/xxhash.h#L5028

Call stack:
```
>	game.exe!XXH3_accumulate_512_avx512(void * acc, const void * input, const void * secret) Line 4195	C
 	game.exe!XXH3_hashLong_internal_loop(unsigned __int64 * acc, const unsigned char * input, unsigned __int64 len, const unsigned char * secret, unsigned __int64 secretSize, void(*)(unsigned __int64 *, const unsigned char *, const unsigned char *, unsigned __int64) f_acc, void(*)(void *, const void *) f_scramble) Line 5093	C
 	game.exe!XXH3_hashLong_64b_internal(const void * input, unsigned __int64 len, const void * secret, unsigned __int64 secretSize, void(*)(unsigned __int64 *, const unsigned char *, const unsigned char *, unsigned __int64) f_acc, void(*)(void *, const void *) f_scramble) Line 5144	C
 	game.exe!XXHL64_default_avx2(const void * input, unsigned __int64 len) Line 555	C
 	game.exe!XXH3_hashLong_64b_defaultSecret_selection(const void * input, unsigned __int64 len, unsigned __int64 seed64, const unsigned char * secret, unsigned __int64 secretLen) Line 688	C
 	game.exe!XXH3_64bits_internal(const void * input, unsigned __int64 len, unsigned __int64 seed64, const void * secret, unsigned __int64 secretLen, unsigned __int64(*)(const void *, unsigned __int64, unsigned __int64, const unsigned char *, unsigned __int64) f_hashLong) Line 5244	C
 	game.exe!XXH3_64bits_dispatch(const void * input, unsigned __int64 len) Line 693	C
```

**EDIT**: For clarity, note that dispatch correctly selected AVX2 (`XXHL64_default_avx2`) but `XXH3_hashLong_internal_loop` directly uses the `XXH3_accumulate_512` macro which resolved to `XXH3_accumulate_512_avx512` (which was the highest toolchain-supported ISA at compile-time).

I'm not sure what the correct line would look like.

[...]

PR #767, opened 2022-11-30
Use ubuntu-22.04 for QEMU tests

t-mat:
This patch replaces ubuntu VM images for QEMU test matrix from `ubuntu-latest` (ubuntu-20.04) to `ubuntu-22.04`.

----
Context: @hzhuang1 pointed out that we're using too old version of QEMU and gcc cross compiler in the `ubuntu-latest` (ubuntu-20.04) environment.
https://github.com/Cyan4973/xxHash/issues/763?notification_referrer_id=NT_kwDOAA21XLE0OTM4ODk5OTQ2Ojg5ODM5Ng#issuecomment-1332080365

Also there're additional issues

(1) QEMU S390X VSX test failed.  https://github.com/Cyan4973/xxHash/actions/runs/3584033303/jobs/6030189599#step:10:319
(2) QEMU matrix tests for ohter versions of S390X are broken.  See #766.

We may need to fix them in the following order

(1) Investigation for S390X VSX.  Where's actual issue? (xxhash.h, gcc, QEMU)
(2) Use the latest version of QEMU and gcc if they are okay to use.  (This PR)
(3) Fix QEMU test matrix (#766)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -302,13 +302,13 @@ jobs:
       fail-fast: false  # 'false' means Don't stop matrix workflows even if some matrix failed.
       matrix:
         include: [
-          { name: 'ARM',             xcc_pkg: gcc-arm-linux-gnueabi,        xcc: arm-linux-gnueabi-gcc,        xemu_pkg: qemu-system-arm,   xemu: qemu-arm-static,     os: ubuntu-latest, },
-          { name: 'ARM64',           xcc_pkg: gcc-aarch64-linux-gnu,        xcc: aarch64-linux-gnu-gcc,        xemu_pkg: qemu-system-arm,   xemu: qemu-aarch64-static, os: ubuntu-latest, },
-          { name: 'PPC64LE',         xcc_pkg: gcc-powerpc64le-linux-gnu,    xcc: powerpc64le-linux-gnu-gcc,    xemu_pkg: qemu-system-ppc,   xemu: qemu-ppc64le-static, os: ubuntu-latest, },
-          { name: 'PPC64',           xcc_pkg: gcc-powerpc64-linux-gnu,      xcc: powerpc64-linux-gnu-gcc,      xemu_pkg: qemu-system-ppc,   xemu: qemu-ppc64-static,   os: ubuntu-latest, },
-          { name: 'S390X',           xcc_pkg: gcc-s390x-linux-gnu,          xcc: s390x-linux-gnu-gcc,          xemu_pkg: qemu-system-s390x, xemu: qemu-s390x-static,   os: ubuntu-latest, },
-          { name: 'MIPS',            xcc_pkg: gcc-mips-linux-gnu,           xcc: mips-linux-gnu-gcc,           xemu_pkg: qemu-system-mips,  xemu: qemu-mips-static,    os: ubuntu-latest, },
-          { name: 'M68K',            xcc_pkg: gcc-m68k-linux-gnu,           xcc: m68k-linux-gnu-gcc,           xemu_pkg: qemu-system-m68k,  xemu: qemu-m68k-static,    os: ubuntu-latest, },
+          { name: 'ARM',             xcc_pkg: gcc-arm-linux-gnueabi,        xcc: arm-linux-gnueabi-gcc,        xemu_pkg: qemu-system-arm,   xemu: qemu-arm-static,     os: ubuntu-22.04, },
+          { name: 'ARM64',           xcc_pkg: gcc-aarch64-linux-gnu,        xcc: aarch64-linux-gnu-gcc,        xemu_pkg: qemu-system-arm,   xemu: qemu-aarch64-static, os: ubuntu-22.04, },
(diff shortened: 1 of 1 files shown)


Cyan4973:
Do I understand that this PR is not final, and there will be some follow-up,
at least to understand why the s390x test is failing ?

t-mat:
Yes, I think s390x VSX issue must be understood.

So far, its random failure is mystery, frankly. 👻 
When I have free time, I'll post my observation to #766.
And will come back to this PR when we understand what's happening to s390x VSX test and/or code.

hzhuang1:
I've tried these configuration.

| Configuration  |  Optimization  | Use f_acc |  CI  Status | Github Action Link  |
|---------------|---------------|-----------|-----------|----------|
| ubuntu-latest | -O2                 | No            | Pass         | https://github.com/hzhuang1/xxHash/actions/runs/3598167899 |
| ubuntu-22.04 | -O2                | No            | Pass         | https://github.com/hzhuang1/xxHash/actions/runs/3598320811 |
| ubuntu-latest | -O2                | Yes           | Pass         | https://github.com/hzhuang1/xxHash/actions/runs/3598401045 |
| ubuntu-22.04 | -O2               | Yes            | Pass         | https://github.com/hzhuang1/xxHash/actions/runs/3598457435 |
| ubuntu-latest | -O3                | No             | Pass         | Default |
| ubuntu-22.04 | -O3                | No             | Fail           | As this pull request |
| ubuntu-latest | -O3                | Yes            | Fail           | https://github.com/hzhuang1/xxHash/actions/runs/3579213227 |
| ubuntu-22.04 | -O3               | Yes            | Fail           | https://github.com/hzhuang1/xxHash/actions/runs/3598511680 |

I don't have the real S390X machine. But I don't have the same issue on my S390X VM or container. Up to now, I have to say the result is obviously affected by gcc optimization.

t-mat:
I wrote simple test program for s390x VSX issue : https://gist.github.com/t-mat/f23db167b55adda8fc4bbdbef89b1879

To help investigation in assembly level, I removed almost everything from `xsum_bench.c`.

note : Test case number 16 is the first failure case.  You can tweak `#if` in the `main()` function to create more simpler code for investigation.

----
## Usage

```sh
# Setup
cd
git clone https://github.com/Cyan4973/xxHash.git xxhash-issue-767
cd xxhash-issue-767
curl -LOJ https://gist.githubusercontent.com/t-mat/f23db167b55adda8fc4bbdbef89b1879/raw/0e486194b54c9d2de255a351e3d6ed4874274ac5/xxhsum-issue-767.c
```

```sh
# Host environment

gcc -O2 xxhsum-issue-767.c && ./a.out
# PASS

gcc -O3 xxhsum-issue-767.c && ./a.out
# PASS
```

```sh
# s390x QEMU environment

sudo apt install gcc-s390x-linux-gnu qemu-system-s390x
s390x-linux-gnu-gcc --version   # s390x-linux-gnu-gcc (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0
qemu-s390x-static --version     # qemu-s390x version 6.2.0 (Debian 1:6.2+dfsg-2ubuntu6.5)

s390x-linux-gnu-gcc -O2 -DXXH_VECTOR=XXH_VSX -march=arch11 -mzvector -static xxhsum-issue-767.c
qemu-s390x-static ./a.out
# PASS

s390x-linux-gnu-gcc -O3 -DXXH_VECTOR=XXH_VSX -march=arch11 -mzvector -static xxhsum-issue-767.c
qemu-s390x-static ./a.out
# FAIL
```

hzhuang1:
Ah, I can reproduce this issue on my s390x container now. Because I missed those vector parameters for gcc. It went into scalar mode instead.

I tried gcc-10, gcc-11 & gcc-12. The issue still exists. I don't know what will happen on real S390X machine.

Up to now, the easiest workaround is to downgrade the gcc optimization from O3 to O2.

hzhuang1:
I’ve tried to checkout v0.7.4. I can still reproduce this issue on S390X with your test case. The issues exists for a long time, but it hasn’t been figured out by CI. Should we make CI to figure it out first?

hzhuang1:
I found another workaround to fix this issue by adding memory barrier.


https://github.com/hzhuang1/xxHash/pull/new/issue_767_barrier

I guess there should be some synchronization between vector instruction stream line and generic instruction stream line. With this workaround, it could pass the test with gcc O3 optimization.

hzhuang1:
Excuse me that the above link is wrong. I just copied it from git.

Please check the commit directly instead.
https://github.com/hzhuang1/xxHash/commit/2046a2257d6a9b6e02028373cb27fdbbfede0542

I only add "__sync_synchronize()" to make it working.

hzhuang1:
Now there are two workarounds. One is using O2 optimization instead. And the other is using __sync_synchronize() with O3 optimization. __sync_synchronize() is only available in GCC or CLANG.

So which one is preferred? From my view, I prefer the second one.

hzhuang1:
So what's the status on this one? I haven't heard any feedback on this issue. I thought that this issue on S390X should be fixed. Otherwise, it'll cause CI error at any time.

Cyan4973:
Yes, we should fix this lingering issue.

One possible solution could be to statically set Ubuntu version to 20.04 for the `s390x` test.

It actually depends if we believe that the issue comes from a `qemu` bug,
or if it comes from the `s390x` code path within `xxhash`.
In the second case, we should rather fix, or at least disable, the buggy code path.

hzhuang1:
It could be closed now since s390x issue is fixed by #785.

PR #769, opened 2022-12-02
Don't compute `nbRounds` and use `len` as upper bound in 129to240

goldsteinn:
Codegen was previously using the divide to truncate the lower-bits but was shifting back for the actual loop.

Previous Codegen:
    ```
    // XXH128_129to240:
    shr    $0x5,%r15d
    lea    -0x5(%r15),%r13d
    shl    $0x5,%r13
    ...
    // XXH64_129to240:
    shr    $0x4,%r9d
    lea    -0x9(%r9),%eax
    shl    $0x4,%rax
    ```

which is totally unnecessary.

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -1803,7 +1803,7 @@ static void* XXH_memcpy(void* dest, const void* src, size_t size)
 #  include <assert.h>   /* note: can still be disabled with NDEBUG */
 #  define XXH_ASSERT(c)   assert(c)
 #else
-#  define XXH_ASSERT(c)   ((void)0)
+#  define XXH_ASSERT(c)   XXH_ASSUME(c)
 #endif
 
 /* note: use after variable declarations */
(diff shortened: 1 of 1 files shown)


goldsteinn:
I made two changes from the last version that *may* fix the unknown regression on aarch64.

1. Leave `len` unchanged (instead of trying to pull of a cast to `u32` to save a touch of code size).
2. Make non-debug `XXH_ASSERT` use `XXH_ASSUME` so that hopefully no information is missing to allow the compiler to optimize the code as best it can.

Cyan4973:
Unfortunately, this variant still craters the bandwidth test for len [129-240]
by about the same amount as before (~-20%)

goldsteinn:
> Unfortunately, this variant still craters the bandwidth test for len [129-240] by about the same amount as before (~-20%)

Got it. Thanks for checking.

Will close, if I get my hands on an m1 in the future may re-open.

PR #773, opened 2022-12-13
dispatch: Use __attribute__((constructor)) on XXH_setDispatch

goldsteinn:
Minimal change but eliminates the NULL check if the attribute is supported.

### DIFF
--- a/xxh_x86dispatch.c
+++ b/xxh_x86dispatch.c
@@ -207,6 +207,23 @@ extern "C" {
 #define XXH_X86DISPATCH
 #include "xxhash.h"
 
+#ifndef XXH_HAS_ATTRIBUTE
+#  ifdef __has_attribute
+#    define XXH_HAS_ATTRIBUTE(...) __has_attribute(__VA_ARGS__)
+#  else
+#    define XXH_HAS_ATTRIBUTE(...) 0
+#  endif
+#endif
+
+#if XXH_HAS_ATTRIBUTE(constructor)
+#  define XXH_CONSTRUCTOR __attribute__((constructor))
+#  define XXH_DISPATCH_MAYBE_NULL 0
+#else
+#  define XXH_CONSTRUCTOR
+#  define XXH_DISPATCH_MAYBE_NULL 1
+#endif
+
+
 /*
  * Support both AT&T and Intel dialects
  *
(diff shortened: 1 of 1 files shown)


goldsteinn:
NB: We could also do this with `ifunc` (on GNU/Linux targets) and save a few hundred bytes of rodata but doesn't seem worth the added code complexity b.c the `ifunc` impl wouldn't live along side the current one so cleanly.

Cyan4973:
Looks good to me.
Would `has_attribute()` only work on `clang` ?
If yes, would it make sense to also get the `constructor` attribute working on `gcc` ?

goldsteinn:
> Looks good to me. Would `has_attribute()` only work on `clang` ? If yes, would it make sense to also get the `constructor` attribute working on `gcc` ?

[GCC has it too](https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fattribute.html)

Cyan4973 (approved):

Cyan4973:
> [GCC has it too](https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fattribute.html)

I completely missed this one !

Seems to be a `gcc-9` thing : https://gcc.gnu.org/gcc-9/changes.html

goldsteinn:
> > [GCC has it too](https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fattribute.html)
> 
> I completely missed this one !
> 
> Seems to be a `gcc-9` thing : https://gcc.gnu.org/gcc-9/changes.html

By the way, docs reminded me that the prettying idomatic way of checking
if a macro is defined + its value (i.e `#if (defined __STDC__VERSION && __STDC__VERSION >= 199901L`)
is not technically portable.
Its used all over the place in xxhash. Worth fixing or non-issue?

Cyan4973:
It's indeed used everywhere, and never was a problem, on any target.
And that's hardly an `xxhash` thing, it's used pretty much in every C OSS I know of.

Why do you think it's not technically portable ?

goldsteinn:
> It's indeed used everywhere, and never was a problem, on any target. And that's hardly an `xxhash` thing, it's used pretty much in every C OSS I know of.
> 
> Why do you think it's not technically portable ?

if [this commit](https://sourceware.org/git/?p=glibc.git;a=commit;h=ddcf5a9170a0db5055259118f3090debab248126) is to be trusted its `p4` of the Processor Standard. The example seems to say:
```
#if (defined __has_attribute && __has_attribute(xxx))
```
is non-portable (I assume b.c its not explicitly standard). It may be different when checking a value though
b.c non-defined PP tokens have implicit zero IIRC.

Although FWIW we have this in a bunch of other places in GLIBC and think its mostly likely a non-issue.

Just something the `__has_attribute` in GCC reminded me about so thought I'd bring it up.

Edit: I misread the [GCC defined doc](https://gcc.gnu.org/onlinedocs/cpp/Defined.html#Defined). It doesn't say its UB there.

PR #779, opened 2023-01-03
VS2019 fix on xxhash.h

aismann:
Issue on axmol (https://github.com/axmolengine/axmol): VS2019: thirdparty\xxhash: Cannot open include file: 'stdalign.h': No such file or directory  See also the solution:
https://github.com/axmolengine/axmol/issues/991

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -1161,7 +1161,7 @@ struct XXH64_state_s {
 
 #ifndef XXH_NO_XXH3
 
-#if defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L) /* >= C11 */
+#if !defined(_MSC_VER) && defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L) /* >= C11 */
 #  include <stdalign.h>
 #  define XXH_ALIGN(n)      alignas(n)
 #elif defined(__cplusplus) && (__cplusplus >= 201103L) /* >= C++11 */


Cyan4973 (approved):

aismann:
The problem is a mistake in xxhash.h lines 1164. The #include <stdalign.h> is done whenever the __STDC_VERSION__ variable is larger than 201112L. A way to fix it is to add
` ... !defined(_MSC_VER) ...`
to line 1164. Then the #include will never be attempted with Visual Studio and VC and the #else case `#elif defined(_MSC_VER) ...` will work.

Cyan4973:
To be more complete, I don't think that the `xxhash` code is at fault here.

`<stdalign.h>` is a mandatory header file for any compiler claiming `C11` compliance,
and advertising `__ STDC_VERSION__ >= 201112L` means exactly that.
Therefore, if `VS2019` claims C11 compliance, it _should_ ship with the `<stdalign.h>` include.

Alas, we live in an imperfect world, and it wouldn't be the first time that a compiler _claims_ compliance but actually only delivers 99% of the experience. In which case, we need to carve out exceptions, one case at a time, to compile nonetheless.

So that's what the proposed patch fixes : a carved-out exception for Visual Studio compilers. And it's fine, because they have their own non-standard compiler-specific equivalent.

That being said : 
Visual Studio 2019 is part of the targets that are tested by xxHash CI test suite.
And it has been compiling completely fine so far.
See [this log](https://github.com/Cyan4973/xxHash/actions/runs/3805252647/jobs/6473062250) for a recent example.
Therefore, it's quite surprising that it would fail specifically for your use case.
This suggest there is a bit more to look into...

Wild guess : are you compiling in C++ mode ?

aismann:
@Cyan4973 you are right.
confirmed, required windows sdk version: 10.0.22000+

PR #784, opened 2023-01-09
s390x: add memory barrier to avoid fault issue

hzhuang1:
The issue link is in below.
https://github.com/Cyan4973/xxHash/issues/766

This issue occurs randomly. If we downgrade gcc optimization from "O3" to "O2", this issue is gone.

Now use memory barrier to implement the same effect.


hzhuang1:
It's a similar issue as the previous pull request. It always report to fail installing package and result missing clang.
Does it mean that it's a network issue?

PR #785, opened 2023-01-10
s390x: add memory barrier to avoid fault issue

hzhuang1:
The issue link is in below.
https://github.com/Cyan4973/xxHash/issues/766

This issue occurs randomly. If we downgrade gcc optimization from "O3" to "O2", this issue is gone.

Now use memory barrier to implement the same effect.

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -4698,6 +4698,7 @@ XXH3_accumulate_512_vsx(  void* XXH_RESTRICT acc,
         /* xacc[i] = acc_vec; */
         vec_xst((xxh_u32x4)acc_vec, 0, xacc + 4 * i);
     }
+    __sync_synchronize();
 }
 XXH_FORCE_INLINE XXH3_ACCUMULATE_TEMPLATE(vsx)
 


xxhash.h
```diff
@@ -4698,6 +4698,7 @@ XXH3_accumulate_512_vsx(  void* XXH_RESTRICT acc,
         /* xacc[i] = acc_vec; */
         vec_xst((xxh_u32x4)acc_vec, 0, xacc + 4 * i);
     }
+    __sync_synchronize();
```
Cyan4973, line 4701: How guaranteed is this `__sync_synchronize()` symbol ?
Is it necessarily part of `<altivec.h>` and `<s390intrin.h>` ?
Is it `gcc` builtin ?

hzhuang1: The synchronization function doesn't belong to s390x intrinsic code. It belongs to gcc/clang builtin code for different architectures, including s390x.

Optimization flag "-O2/-O3" would result in instructions out of order. "-O3" optimizes more than "-O2". The synchronization function tries to add memory barrier that make sure instructions before it all executed. It could avoid instructions out of order at that point.

Cyan4973: Do we have any guarantee that, if we reach this code, then it's necessarily `gcc` or `clang`, and therefore the `__sync_synchronize()` builtin is necessarily valid ?
Or could another compiler reach this code path, and choke on the builtin ?

hzhuang1: Now xxhash on s390 could only be built for linux. We either choose gcc or clang. We don't have a third choice.

__sync_synchronize() is based on instructions on different architectures. We can refer to $GCC/gcc/config/aarch64/atomics.md. The DMB instruction is used to implement memory barrier. For $GCC/gcc/config/s390/s390.md, we could also find memory barrier is supported.

__sync_synchronize() just makes use of the instructions provided by different architectures. And they're not vector instructions, they're just normal instructions.

If a third compiler want to build it without __sync_synchronization, it can even use those instructions directly just as gcc.

hzhuang1: Is there any concern on my comments?

hzhuang1: Could it be merged?

Cyan4973: Yes.
I guess if it breaks compilation for some unforeseen compiler / platform, we could always add some `#ifdef __GNUC__` around to gate the symbol invocation.

Cyan4973 (approved):

Cyan4973:
Thanks @hzhuang1 !
This PR has fixed the recurrent `s390x` CI failure issue.

PR #786, opened 2023-01-13
Annotate xxh_x86dispatch with XXH_NOESCAPE

goldsteinn:
The pointers don't escape.

### DIFF
--- a/xxh_x86dispatch.c
+++ b/xxh_x86dispatch.c
@@ -472,7 +472,8 @@ static int XXH_featureTest(void)
 /* ===   XXH3, default variants   === */                                      \
                                                                               \
 XXH_NO_INLINE target XXH64_hash_t                                             \
-XXHL64_default_##suffix(const void* XXH_RESTRICT input, size_t len)           \
+XXHL64_default_##suffix(XXH_NOESCAPE const void* XXH_RESTRICT input,          \
+                        size_t len)                                           \
 {                                                                             \
     return XXH3_hashLong_64b_internal(                                        \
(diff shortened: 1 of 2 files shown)


Cyan4973 (approved):

Cyan4973:
While I'm fine with adding this annotation,
I start to be concerned about brevity.
All these annotations make declarations more verbose,
which in turn makes reading declarations more difficult.

I was wondering if there would be a way to simplify this.
Maybe some `typedef` with inherited attributes ?

goldsteinn:
> While I'm fine with adding this annotation, I start to be concerned about brevity. All these annotations make declarations more verbose, which in turn makes reading declarations more difficult.
> 
agreed.

> I was wondering if there would be a way to simplify this. Maybe some `typedef` with inherited attributes ?

`typedef` does't seem to work: https://godbolt.org/z/xh9r1brcq and warning is pretty explicit it's only for parameters.
A `define` would work though, something like:
```
#define xxh_void_arg XXH_NOESCAPE void
#define xxh_u8_arg XXH_NOESCAPE xxh_u8
...
void XXH3_64bits(xxh_void_arg const * input);
```
I can fixup this one and update the previous patch if its preferable.

Cyan4973:
Thanks @goldsteinn ,
let's take this patch as is.
It hasn't yet reached the point of being too cumbersome to read,
it's only trending in this direction.

If another annotation needs to be added, then probably a `#define` to summarize multiple annotations and qualifiers (and type maybe) would become useful.

PR #790, opened 2023-01-29
cmake: support dispatch mode

hzhuang1:
Add dispatch mode with option DISPATCH.

Build with DISPATCH:
$cmake ../cmake_unofficial -DDISPATCH=ON
$cmake --build ./


hzhuang1:
It depends on cmake 3.10 to recognize architecture.

Cyan4973:
> It depends on cmake 3.10 to recognize architecture.

You have designed a fallback in case `cmake` doesn't have a good enough version, so I think this is good.

Cyan4973 (approved):

Issue #791, opened 2023-01-31
Compile issue with gcc 11 -std=c++23


eseiler:
Compiling with `gcc-11 -std=c++23`:
```
xxhash.h:2052:34: error: ‘unreachable’ is not a member of ‘std’
 2052 | #  define XXH_UNREACHABLE() std::unreachable()
```

#782 introduced
https://github.com/Cyan4973/xxHash/blob/6dba9abfaae75a4b70a5c53d7e5eb3dd554973d1/xxhash.h#L2049-L2052

The problem with `__cplusplus` is that it does not necessarily mean that all features are available.
It's usually more robust to use the [feature testing macros](https://en.cppreference.com/w/cpp/feature_test) (`__cpp_lib_unreachable`). 
In this case, it would mean using
```cpp
#elif defined(__cpp_lib_unreachable) && (__cpp_lib_unreachable >= 202202L)
```

See also https://godbolt.org/z/5Ys77oM9j

I am not sure if this also applies to the C version. [GCC](https://github.com/gcc-mirror/gcc/blob/b2ec2504af77b35e748067eeb846821d12a6b6b4/gcc/ginclude/stddef.h#L457) defines `__STDC_VERSION_STDDEF_H__`, but I haven't found a reference for this.

Cyan4973:
cc @goldsteinn

goldsteinn:
> Compiling with `gcc-11 -std=c++23`:
> 
> ```
> xxhash.h:2052:34: error: ‘unreachable’ is not a member of ‘std’
>  2052 | #  define XXH_UNREACHABLE() std::unreachable()
> ```
> 
> #782 introduced
> 
> https://github.com/Cyan4973/xxHash/blob/6dba9abfaae75a4b70a5c53d7e5eb3dd554973d1/xxhash.h#L2049-L2052
> 
> The problem with `__cplusplus` is that it does not necessarily mean that all features are available. It's usually more robust to use the [feature testing macros](https://en.cppreference.com/w/cpp/feature_test) (`__cpp_lib_unreachable`). In this case, it would mean using
> 
> ```c++
> #elif defined(__cpp_lib_unreachable) && (__cpp_lib_unreachable >= 202202L)
> ```
> 
> See also https://godbolt.org/z/5Ys77oM9j
>

Thank you, will update the CPP version to this.
 
[...]

PR #792, opened 2023-02-01
Fix `XXH_UNREACHABLE` feature test for C++23 and remove for C23 #791

goldsteinn:
gcc11/12 don't support `std::unreachable()` in there `-std=c++23` build.

A more reliable check is:
```
```
so use that.

A similiar issue may occur for C23 `unreachable()` so just remove it on the basis that if a compiler doesn't currently support some builtin for unreachable/assume its very likely not using it in analysis so we are probably not missing much (if anything) leaving it out.

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -716,8 +716,15 @@ XXH_PUBLIC_API XXH_PUREF XXH32_hash_t XXH32_hashFromCanonical(const XXH32_canoni
 # define XXH_HAS_ATTRIBUTE(x) 0
 #endif
 
+/*
+ * C23 __STDC_VERSION__ number hasn't been specified yet. For now
+ * leave as `201711L` (C17 + 1).
+ * TODO: Update to correct value when its been specified.
+ */
+#define XXH_C23_VN 201711L
+
 /* C-language Attributes are added in C23. */
-#if defined(__STDC_VERSION__) && (__STDC_VERSION__ > 201710L) && defined(__has_c_attribute)
+#if defined(__STDC_VERSION__) && (__STDC_VERSION__ >= XXH_C23_VN) && defined(__has_c_attribute)
 # define XXH_HAS_C_ATTRIBUTE(x) __has_c_attribute(x)
 #else
 # define XXH_HAS_C_ATTRIBUTE(x) 0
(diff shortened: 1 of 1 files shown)


xxhash.h
```diff
@@ -2041,12 +2041,20 @@ static int XXH_isLittleEndian(void)
...
-#  define XXH_UNREACHABLE() unreachable()
 
-#elif defined(__cplusplus) && (__cplusplus > 202002L)
+/*
+ * C23 and future versions have standard "unreachable()".
+ * Once it has been implemented reliably we can add it as an
+ * additional case:
+ *
+ * ```
+ * #if defined(__STDC_VERSION__) && (__STDC_VERSION__ > 201710L)
```
Cyan4973: I presume you meant `__STDC_VERSION__ > 202310L` in this example

goldsteinn: Truthfully not sure. The code was based on what @t-mat suggested [here](https://github.com/Cyan4973/xxHash/pull/754#issuecomment-1312366653).

I can drop the comment entirely if its liable to make confusion.

t-mat: Since C23 didn't set actual `__STDC_VERSION__` yet (right?), I think we can express C23 as "beyond/greater than C17 (`> 201710L`)" for now.
It must be fixed after the comittee and compilers set official `__STDC_VERSION__` though.
So, adding `TODO` or `note` comment would be nice to reminder.

t-mat:
Thank you @goldsteinn for quick fix!

----

Since this PR drops support for C23, below is not a review.  Just a note for (possible) future issues/improvements.

According to "7.21 Common definitions <stddef.h>" in [n3054.pdf](https://open-std.org/JTC1/SC22/WG14/www/docs/n3054.pdf), `unreachable()` is a macro.
So more "safer" definition may have check for it.

```C
#if defined(__STDC_VERSION__) && (__STDC_VERSION__ >= XXH_C23_VN)
   /* C23 and future versions have unreachable() macro */
#  include <stddef.h>
#  if defined(unreachable)
#    define XXH_UNREACHABLE() unreachable()
#  endif
#endif
```

If we can allow cumbersome definition, it may look like this:

```C
/* C/C++ standard definitions */
#if !defined(XXH_UNREACHABLE)
#  if defined(__STDC_VERSION__) && (__STDC_VERSION__ >= XXH_C23_VN)
     /* C23 and future versions have unreachable() macro */
#    include <stddef.h>      /* unreachable() */
#    if defined(unreachable)
#      define XXH_UNREACHABLE() unreachable()
#    endif
#  elif defined(__cpp_lib_unreachable) && (__cpp_lib_unreachable >= 202202L)
     /* C++23 and future versions have std::unreachable() */
#    include <utility>     /* std::unreachable() */
#    define XXH_UNREACHABLE() std::unreachable()
#  endif
#endif

/* Compiler specific definitions */
#if !defined(XXH_UNREACHABLE)
#  if XXH_HAS_BUILTIN(__builtin_unreachable)
#    define XXH_UNREACHABLE() __builtin_unreachable()
#  elif defined(_MSC_VER)
#    define XXH_UNREACHABLE() __assume(0)
#  endif
#endif

/* Fallback */
#if !defined(XXH_UNREACHABLE)
#  define XXH_UNREACHABLE()
#endif
```

eseiler:
Thank you for the fix!

One more thing:
`__cpp_lib_unreachable` is defined in `<version>`.

`xxhash.h` currently includes
```cpp
#include <stddef.h>
#include <stdlib.h>
#include <string.h>
#include <limits.h>
```
None of which include `<version>`. So, `__cpp_lib_unreachable` will always be undefined. I also checked that just including `xxhash.h` does indeed not define the macro.

I would propose adding a 
```cpp
#if __has_include(<version>)
#include <version>
#endif
```
Before checking for `__cpp_lib_unreachable`.

The `<version>` header only exists since gcc 9, clang 9, and msvc 19.22, hence the check with `__has_include`.

goldsteinn:
> Thank you for the fix!
> 
> One more thing: `__cpp_lib_unreachable` is defined in `<version>`.
> 
> `xxhash.h` currently includes
> 
> ```c++
> #include <stddef.h>
> #include <stdlib.h>
> #include <string.h>
> #include <limits.h>
> ```
> 
> None of which include `<version>`. So, `__cpp_lib_unreachable` will always be undefined. I also checked that just including `xxhash.h` does indeed not define the macro.
> 
> I would propose adding a
> 
> ```c++
> #if __has_include(<version>)
> #include <version>
> #endif
> ```
> 
> Before checking for `__cpp_lib_unreachable`.
> 
> The `<version>` header only exists since gcc 9, clang 9, and msvc 19.22, hence the check with `__has_include`.

Hmm, so once its actually working there seems to be an issue with the `extern "C"` b.c `<utility>` contains templates.
see: https://godbolt.org/z/v773nWeYP

Not too hard to fix by closing the `extern "C"` before include / re-enabling after, but seems like the kind of thing that may cause an link error in the future (if std::unreachable() ever gets implemented with templates for example).
Think I'm just going to drop `std::unreachable()` for now.

Cyan4973 (approved):

PR #804, opened 2023-02-27
Fix GCC 12 -Og

easyaspi314:
The issue actually appears to be GCC disliking that a function pointer is SOMETIMES forceinline, and withSecret was marked force inline instead of noinline (despite what the comments suggested)

This marks `XXH3_hashLong_{64,128}b_withSecret` as `XXH_NO_INLINE` instead of `XXH_FORCE_INLINE` like the rest of the `hashLong` functions.

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -1628,6 +1628,23 @@ XXH3_128bits_reset_withSecretandSeed(XXH_NOESCAPE XXH3_state_t* statePtr,
 #  define XXH_NO_INLINE_HINTS 0
 
 /*!
+ * @def XXH3_INLINE_SECRET
+ * @brief Determines whether to inline the XXH3 withSecret code.
+ *
+ * When the secret size is known, the compiler can improve the performance
+ * of XXH3_64bits_withSecret() and XXH3_128bits_withSecret().
+ *
+ * However, if the secret size is not known, it doesn't have any benefit. This
+ * happens when xxHash is compiled into a global symbol. Therefore, if
+ * @ref XXH_INLINE_ALL is *not* defined, this will be defined to 0.
+ *
+ * Additionally, this defaults to 0 on GCC 12+, which has an issue with function pointers
+ * that are *sometimes* force inline on -Og, and it is impossible to automatically
+ * detect this optimization level.
+ */
+#  define XXH3_INLINE_SECRET 0
+
+/*!
  * @def XXH32_ENDJMP
  * @brief Whether to use a jump for `XXH32_finalize`.
  *
@@ -1698,6 +1715,15 @@ XXH3_128bits_reset_withSecretandSeed(XXH_NOESCAPE XXH3_state_t* statePtr,
 #    define XXH_NO_INLINE_HINTS 1
 #  else
 #    define XXH_NO_INLINE_HINTS 0
+#  endif
+#endif
+
+#ifndef XXH3_INLINE_SECRET
+#  if (defined(__GNUC__) && !defined(__clang__) && __GNUC__ >= 12) \
+     || !defined(XXH_INLINE_ALL)
+#    define XXH3_INLINE_SECRET 0
+#  else
+#    define XXH3_INLINE_SECRET 1
 #  endif
 #endif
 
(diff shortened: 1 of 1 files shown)


easyaspi314:
It might be a good idea to add `-Og` and GCC 12 to CI.

easyaspi314:
Fixes #800 #720 
improved version of fix in #802

Cyan4973:
> It might be a good idea to add `-Og` and GCC 12 to CI.

Agreed

easyaspi314:
I misread the comment. Perhaps this?
```c
#if __GNUC__ >= 12 || !defined(XXH_INLINE_ALL)
XXH_NO_INLINE
#else
XXH_FORCE_INLINE
#endif
```

easyaspi314:
I now made it force inline if and only if XXH_INLINE_ALL is defined and it isn't GCC 12

xxhash.h
```diff
@@ -5146,8 +5146,15 @@ XXH3_hashLong_64b_internal(const void* XXH_RESTRICT input, size_t len,
  * It's important for performance to transmit secret's size (when it's static)
  * so that the compiler can properly optimize the vectorized loop.
  * This makes a big performance difference for "medium" keys (<1 KB) when using AVX instruction set.
+ * When the secret size is unknown, or on GCC 12 where the mix of NO_INLINE and FORCE_INLINE
+ * breaks -Og, this is XXH_NO_INLINE.
  */
-XXH_FORCE_INLINE XXH64_hash_t
+#if (defined(__GNUC__) && !defined(__clang__) && __GNUC__ >= 12) || !defined(XXH_INLINE_ALL)
```
Cyan4973: This should probably be abstracted behind a dedicated macro, that can then be shared by the 2 places where it's invoked.

xxhash.h
```diff
@@ -5146,8 +5146,15 @@ XXH3_hashLong_64b_internal(const void* XXH_RESTRICT input, size_t len,
  * It's important for performance to transmit secret's size (when it's static)
  * so that the compiler can properly optimize the vectorized loop.
  * This makes a big performance difference for "medium" keys (<1 KB) when using AVX instruction set.
+ * When the secret size is unknown, or on GCC 12 where the mix of NO_INLINE and FORCE_INLINE
```
Cyan4973, line 5180: The comment explains that, when secret's size is `static`,
then it's advantageous for performance to pass that to the function.
This suggests that the constant size is propagated inside `XXH3_hashLong_64b_withSecret()` for further optimization.

Of course, if we use `XXH_NO_INLINE`, this property is lost.

Now, this could be nonetheless an acceptable trade-off if the goal is to survive the specific `gcc-12 -Og` compilation scenario.

But my understanding of below macro test is that it's more general : 
1) It's applicable to any current and future version of `gcc`, >= 12
2) `XXH_NO_INLINE` is _always_ set (not just when `-Og` is set), with the exception when inlining is expressly required with `XXH_INLINE_ALL`.

So, effectively, this will impact performance on a broader range of scenarios.

And I wonder if this is the right trade off.

easyaspi314: When XXH3 is not inlined, there isn't much of a benefit to inlining XXH3_hashLong_withSecret, as the secret size would be entirely unknown.

Cyan4973: That's a good point.
For some reason, I expected this variant to be within the call chain of the "default" variant using the default `kSecret`,
but it's not, "default" delegates directly to `XXH3_hashLong_64b_internal()`.

Cyan4973 (approved):

Cyan4973:
I'm a little lost about this `gcc-12 -Og` issue.
It seems that #814 fixes it, using a different initialization strategy.

This PR seems to address the same issue, but in a very different way (`inline`).
Is it still useful despite #814 ?

easyaspi314:
This specific issue comes from the "failed to inline" errors, the other issue is unrelated.

Basically, the issue is that GCC 12 on -Og does not like a function pointer that can either point to something marked `always_inline` or `noinline`, and it errors.

tristan957:
The Meson workaround should be fixed too if anyone gets around to it.

Issue #806, opened 2023-02-28
[security] Set minimum permissions for workflows


gabibguti:
Setting minimum permissions to your workflows is important to keep your repository safe against supply-chain attacks. By default, GitHub gives higher permissions then necessary to the GITHUB_TOKEN generated for workflows. I see `ci.yml` does not need many permissions to run the tests. The actions used by `ci.yml`, `msys2/setup-msys2` and `actions/checkout`, only need access to read the repository code. Thus, we can reduce the permissions to `contents: read` and should be good to go. This change should not impact any tests for any environments :) If you agree with the changes, I can open a PR!

This is considered good-practice and recommended by [GitHub itself](https://docs.github.com/en/actions/security-guides/automatic-token-authentication) and by other security tools, such as [Scorecards](https://github.com/ossf/scorecard) and [StepSecurity](https://github.com/step-security).

##### Additional Context
I'm Gabriela and I work on behalf of Google and the OpenSSF suggesting supply-chain security changes :)

t-mat:
Hi @gabibguti,

Thank you for letting us know.  I suppose that you're suggesting the following actions to us.

1. Set "Read repository contents and packages permissions" to the "Workflow permission" in the [Settings tab of the repo](https://github.com/Cyan4973/xxHash/settings/actions).
  Reference: [Setting the permissions of the GITHUB_TOKEN for your repository - GitHub Docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#setting-the-permissions-of-the-github_token-for-your-repository)

2. As you described, add the following lines to [`ci.yml`](https://github.com/Cyan4973/xxHash/blob/2b328a10983d232364ceda15df1d33531b5f0eb5/.github/workflows/ci.yml#L55).  Similar (same?) to above settings, but since this form is written in code, it may propagate to other forks.
  Reference: [Configuring the default GITHUB_TOKEN permissions](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#configuring-the-default-github_token-permissions)

```YAML
        on: [push, pull_request]

+       permissions:
+         contents: read

        jobs:
```

I think we need to do both of them.  Because 1. helps all branches which don't have `permissions:` in `ci.yml`.
[...]

PR #807, opened 2023-03-01
added read-all permission status to Github Actions Workflow

Cyan4973:
Following recommendations from @gabibguti and @t-mat  at #806

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,12 @@ name: xxHash CI tests
 
 on: [push, pull_request]
 
+concurrency:
+  group: fast-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions: read-all
+
 jobs:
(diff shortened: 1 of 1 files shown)


gabibguti:
Just FYI `permissions: read-all` is different than `permissions: contents: read`.

`read-all` grants `read` access to `actions`, `checks`, `contents` and other scopes, while `contents: read` gives `read` access only to `contents` scope and `none` access to all other scopes. The second option is more similar to the setting "Read repository content".

That said, I don't see any problems granting `read-all` in this case. 👍 

https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

gabibguti (approved):

Cyan4973:
Thanks for explaining the distinction @gabibguti , it's very clear !

There's no reason to give more rights than necessary, 
let's change it to `content: read` as initially suggested by @t-mat .

Issue #810, opened 2023-03-03
GH-Actions (todo) : Add `-Og` and `gcc-12`


t-mat:
- https://github.com/Cyan4973/xxHash/pull/804#issuecomment-1447244510
- https://github.com/Cyan4973/xxHash/issues/800

t-mat:
`v0.8.1` and [`HEAD`](https://github.com/Cyan4973/xxHash/tree/2b328a10983d232364ceda15df1d33531b5f0eb5) may have issue which is not related to `inline` problems.  But `-Og` helps us to detect it.

```
git clone https://github.com/Cyan4973/xxHash.git xxHash-v0.8.1 --branch v0.8.1 --single-branch --depth 1
cd xxHash-v0.8.1

make clean all
# OK

MOREFLAGS="-Og" make clean all
# NG (warnings)
```

<details><summary>Here's actual warning messages.</summary>

```
cc -O3 -Wall -Wextra -Wconversion -Wcast-qual -Wcast-align -Wshadow -Wstrict-aliasing=1 -Wswitch-enum -Wdeclaration-after-statement -Wstrict-prototypes -Wundef -Wpointer-arith -Wformat-security -Wvla -Wformat=2 -Winit-self -Wfloat-equal -Wwrite-strings -Wredundant-decls -Wstrict-overflow=2 -Og -DXXH_INLINE_ALL cli/xxhsum.c -o xxhsum_inlinedXXH
In file included from cli/xsum_sanity_check.c:31,
                 from cli/xxhsum.c:41:
In function ‘XXH_INLINE_XXH3_64bits_reset_withSeed’,
    inlined from ‘XXH_INLINE_XXH3_128bits_reset_withSeed’ at cli/../xxhash.h:5402:12,
    inlined from ‘localXXH128_stream_seeded’ at cli/xsum_bench.c:176:5:
cli/../xxhash.h:4733:46: warning: ‘state.extSecret’ may be used uninitialized [-Wmaybe-uninitialized]
 4733 |     if ((seed != statePtr->seed) || (statePtr->extSecret != NULL))
      |                                      ~~~~~~~~^~~~~~~~~~~
In file included from cli/xxhsum.c:42:
cli/xsum_bench.c: In function ‘localXXH128_stream_seeded’:
cli/xsum_bench.c:174:18: note: ‘state’ declared here
  174 |     XXH3_state_t state;
      |                  ^~~~~
In file included from cli/xsum_sanity_check.c:31,
[...]

PR #814, opened 2023-03-06
Fix: add gcc -Og check and fix its false positive warnings

t-mat:
Before adding `gcc-12` and `-Og` test to our CI, I'd like to fix `gcc -Og` issue.
See also #810

Actually, this is a false positve of `gcc -Og`.  So there must be another ways to avoid this warning.
But since it seems there's easy / safe / explicit fix for it with negilible const, I'd like to it to `XXH3_INITSTATE()`.
( See advice https://github.com/Cyan4973/xxHash/issues/810#issuecomment-1455245974 from  @easyaspi314 )

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -316,6 +316,11 @@ clangtest: clean
 	@echo ---- test clang compilation ----
 	CC=clang MOREFLAGS="-Werror -Wconversion -Wno-sign-conversion" $(MAKE) all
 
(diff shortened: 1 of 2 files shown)


t-mat:
note: Test badge shows red ❌ , but it actually means "some tests are skipped".  #817 fixes this issue.

Cyan4973 (approved):

Issue #821, opened 2023-03-09
Add dedicated sanity test program


t-mat:
To detect a bug like #816 , we need more complete and dedicated sanity test program.

easyaspi314:
In addition, this could be an opportunity to make the quick bench more lightweight by decreasing the initial number of iterations to something low like 10 MB/s which can greatly improve performance if the target cannot reach high speeds, and automatically adjusts on faster targets. This is easy to notice when using `qemu`, `-O0`, or sanitizers.

As an extreme example, using my pretty crappy 1.5 GHz Cortex-A53 tablet with `s390x` in QEMU with `-O0 -mzvector`, `./xxhsum -b99i0` took 38 minutes, as the combination of the unoptimized code and the emulation layer brought it to 3 MB/s, meaning each iteration was about 100 seconds. 

![Screenshot_20230308-125702.png](https://user-images.githubusercontent.com/6258309/224182413-930968fd-5d6a-4f5c-9bf3-9611c319f026.png)

For the more accurate and thorough benchmarks, we have benchHash, and the same should be for the self tests.

t-mat:
I updated test vector generator and sanity test program.

We can test them with the following commands:

```sh
git clone https://github.com/Cyan4973/xxHash.git xxHash-issue-821-2023-0310
cd xxHash-issue-821-2023-0310
cd tests

curl -JOL https://gist.githubusercontent.com/t-mat/168b85d35af11c17ad5953317ff3f2d9/raw/sanity_test_vectors_generator.c
curl -JOL https://gist.githubusercontent.com/t-mat/6db94988ff5dcb4a979c1291144f7154/raw/sanity_test.c

# Generate sanity_test_vectors.h (HUGE! 😮 )
cc sanity_test_vectors_generator.c && ./a.out

# Actual sanity test
cc sanity_test.c && ./a.out
```

If it seems okay as a version 0, TODOs are
- [x] Write test vector generator and sanity test function for secret generator.
- [x] Make sure it tests and provides good test vectors to prevent similar error.
- [x] Add target for `make check`.
- [ ] Tweak compatibility with `xsum_sanity_check.c` (should we care about it?)
- [x] Tweak compatibility with QEMU platforms. (Definitely we need it)

Cyan4973:
Yes, this is pretty good @t-mat !

Cyan4973:
> Tweak compatibility with xsum_sanity_check.c (should we care about it?) 

I'm not sure if that's necessary.
[...]

PR #822, opened 2023-03-10
benchmark : adjust initial speed target to 10 MB/s

Cyan4973:
Additionally, the initial speed target can be selected at compilation time, by setting new optional build macro `XXH_1ST_SPEED_TARGET`.

Solves @easyaspi314 comment in https://github.com/Cyan4973/xxHash/issues/821#issuecomment-1462961579, in order to support "1.5 GHz Cortex-A53 tablet with `s390x` in QEMU with `-O0 -mzvector`" which is very slow.


easyaspi314:
This should improve the performance on the qemu CI builds too, making sure the `./xxhsum -bi0` in make check doesn't take long

PR #823, opened 2023-03-11
Dedicated sanity test program

t-mat:
This PR adds dedicated sanity test program.  See #821 for details.

## `make check`
- `make check` invokes `make test_sanity`
- `make test_sanity` invokes `make -C tests test_sanity`
- `make -C tests test_sanity` creates and executes `test/sanity_test`
  - Note that `test/sanity_test` depends `test/sanity_test_vectors.h`

## `make -C tests sanity_test_vectors.h`
When we need to update `tests/sanity_test_vectors.h`, we should invoke the following command

```
make -C tests sanity_test_vectors.h
```

This command creates and executes `tests/sanity_test_vectors_generator`.
And it updates `tests/sanity_test_vectors.h`

## TODO

Since large part of `sanity_test.c` is copied from `xsum_sanity_check.c`, we should tweak then to avoid to copy same code to multiple places.
But as a "version 0" of dedicated sanity test program, I leave them as is to detect possible issue in `sanity_test.c`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -305,7 +305,7 @@ jobs:
       matrix:
         include: [
           { name: 'ARM',             xcc_pkg: gcc-arm-linux-gnueabi,        xcc: arm-linux-gnueabi-gcc,        xemu_pkg: qemu-system-arm,   xemu: qemu-arm-static,     os: ubuntu-latest, },
-          { name: 'ARM64',           xcc_pkg: gcc-aarch64-linux-gnu,        xcc: aarch64-linux-gnu-gcc,        xemu_pkg: qemu-system-arm,   xemu: qemu-aarch64-static, os: ubuntu-latest, },
+          { name: 'AARCH64',         xcc_pkg: gcc-aarch64-linux-gnu,        xcc: aarch64-linux-gnu-gcc,        xemu_pkg: qemu-system-arm,   xemu: qemu-aarch64-static, os: ubuntu-latest, },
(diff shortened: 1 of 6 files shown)


t-mat:
I've fixed issue for older verion of QEMUs.

- Replace matrix.name checking condition operator `==` with `startsWith()`

But since `gcc-8-aarch64-linux-gnu` has linker issue (it doesn't work with `LDFLAGS=-static`), I commented out it for now.

## Reproducing and workaround for `gcc-8-aarch64-linux-gnu` issue

```
sudo apt install -y gcc-8-aarch64-linux-gnu qemu-system-arm qemu-user-static

CPPFLAGS="-DXXH_VECTOR=XXH_SCALAR" LDFLAGS="-static" CC=aarch64-linux-gnu-gcc-8 RUN_ENV=qemu-aarch64-static make clean check
# Linker error

CPPFLAGS="-DXXH_VECTOR=XXH_SCALAR" CC=aarch64-linux-gnu-gcc-8 RUN_ENV="qemu-aarch64 -L /usr/aarch64-linux-gnu" make clean check
# Works.  But it uses `qemu-aarch64`
```

Cyan4973:
Regarding the "TODO" :
is it planned to be added as a later part of this PR,
or will it be a separate PR ?

t-mat:
oops.  Sorry for my ambiguous message.
I mean I'll post separate PR for "TODO".

Cyan4973 (approved):
Great work @t-mat !

PR #833, opened 2023-05-12
Reference actions by commit SHA

### LINKED ISSUE #832: Reference actions by commit SHA
Referencing actions by commit SHA in GitHub workflows, guarantees you are using an immutable version. In contrast, actions referenced by tags and branches are vulnerable to attacks, such as the tag being moved to a malicious commit, a malicious commit being pushed to the branch or typosquatting.

Although there are pros and cons for each reference, GitHub acknowledges that [using commit SHAs is more reliable](https://docs.github.com/en/actions/learn-github-actions/finding-and-customizing-actions#using-shas), as does [Scorecard](https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies) security tool.

Currently, in this repository, we use the GitHub action `actions/checkout@v3` and the third-party action `msys2/setup-msys2@v2`. Both referenced by tags. If you agree, I can open a PR to change these references to commit SHAs.

##### Additional Context

Hi again! I'm Gabriela and I work on behalf of Google and the OpenSSF suggesting supply-chain security changes :)

gabibguti:
Resolves https://github.com/Cyan4973/xxHash/issues/832

It's important to make sure the SHA's are from the original repositories and not forks. I have manually verified them and added references in the commit description.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,11 +27,11 @@
 # - You can investigate various information at the right pane of GitHub
 #   Actions report page.
 #
-#   | Item                      | Section in the right pane             |
-#   | ------------------------- | ------------------------------------- |
-#   | OS, VM                    | Set up job                            |
-#   | git repo, commit hash     | Run actions/checkout@v3               |
-#   | gcc, tools                | Environment info                      |
+#   | Item                      | Section in the right pane                                              |
+#   | ------------------------- | ---------------------------------------------------------------------- |
+#   | OS, VM                    | Set up job                                                             |
+#   | git repo, commit hash     | Run actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2 |
+#   | gcc, tools                | Environment info                                                       |
 #
 # - To fail earlier, order of tests in the same job are roughly sorted by
 #   elapsed time.
@@ -104,7 +104,7 @@ jobs:
       CC: ${{ matrix.cc }}
       CXX: ${{ matrix.cxx }}
     steps:
-    - uses: actions/checkout@v3 # https://github.com/actions/checkout
+    - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
 
     - name: apt-get install
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -27,11 +27,11 @@
...
 #
-#   | Item                      | Section in the right pane             |
-#   | ------------------------- | ------------------------------------- |
-#   | OS, VM                    | Set up job                            |
-#   | git repo, commit hash     | Run actions/checkout@v3               |
-#   | gcc, tools                | Environment info                      |
+#   | Item                      | Section in the right pane                                              |
+#   | ------------------------- | ---------------------------------------------------------------------- |
+#   | OS, VM                    | Set up job                                                             |
+#   | git repo, commit hash     | Run actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2 |
```
Cyan4973, line 33: Is it useful to also update the comment ?

Cyan4973 (approved):

.github/workflows/ci.yml
gabibguti: Up to you! We can revert to `Run actions/checkout@v3` or change to `Run actions/checkout@v3.5.2`, whatever feels easier to understand here.

Cyan4973: Indeed, it's just an overview,
I would assume `Run actions/checkout@v3` to be significant enough here.

I'm just concerned about some potential future automated code changes, like possibly dependabot, that could miss the comment nature of this line.

gabibguti: I have tested renovatebot on this case. The bot updates only the real action line and not the table comment line. So, the table comment and the action may get unsynced over time.

gabibguti: Complementing, I have tested comment lines such as:

- `# | Run actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2 |`
- `# actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab`
- `# actions/checkout@v3`

and none get updated by renovatebot.

Cyan4973: Yes, that's why I was more in favor of keeping it simple and generic, since it won't get updated with the rest of the code.
But don't worry, I'll do it directly.

PR #838, opened 2023-06-17
Feature: Support # comment lines in --check files

### LINKED ISSUE #836: Support # comment lines in --check files
`md5sum` seems to support `#` comments in `--check` files.

For better compatibility, `xxhsum` should preferably support the same capability.

t-mat:
This PR resolves #836.

Basically, we just follow the following logic in the `digest.c` of GNU coreutils.
https://github.com/coreutils/coreutils/blob/d53190ed46a55f599800ebb2d8ddfe38205dbd24/src/digest.c#L1180-L1182

This PR also contains a trivial test shell script, `Makefile` entries and `ci.yml` entry for it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -251,6 +251,10 @@ jobs:
       run: |
         make clean test-filename-escape
 
+    - name: test-cli-comment-line
+      run: |
+        make clean test-cli-comment-line
+
   ubuntu-cmake-unofficial:
     name: Linux x64 cmake unofficial build test
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


t-mat:
tipi.build fails for some reason.  I have no idea though 😿 

image

Cyan4973 (approved):
This looks all great to me !

Cyan4973:
The `tipi` CI test has unfortunately proven unstable several times in the past, so I don't think your code is at stake here.

That's mostly because this test downloads a container with a changing content, occasionally breaking established behavior, a symptom of an evolving experimental project.

It's probably best to remove this test from the `xxhash` CI suite.

PR #749, opened 2022-10-19
ci: fix tipi build error on github CI workflow

hzhuang1:
[err] Dynamic exception type: std::runtime_error
std::exception::what: can not create git repository [-3] - config value
  'safe.directory' was not found
[err] Something went wrong, for help on your issue, we could enjoy getting
  your report at https://tipi.build/#contact
Error: Process completed with exit code 1.

Apply the workaround from Yannic (https://github.com/Cyan4973/xxHash/pull/744) on github CI workflow

Signed-off-by: Haojian Zhuang <haojian.zhuang@linaro.org>
Cc: Yannic Staudt <yannic@tipi.build>

PR #744, opened 2022-10-09
Full acc loop

hzhuang1:
It's a replacement patch set of (https://github.com/Cyan4973/xxHash/pull/692).

With this patch set, empty parameter is totally avoided.

PR #841, opened 2023-06-19
removed tipi test from CI suite because it's unstable

Cyan4973:
I wonder if I should also remove the files in `.tipi/` directory.

These files are essentially empty, so they don't seem to add much value.

On the other hand, I'm not sure if such an action would break _something_, like an external build system depending on this repository.


t-mat:
Here are some logs about tipi.build for the record.

- https://github.com/Cyan4973/xxHash/pull/744#issuecomment-1274389957
- https://github.com/Cyan4973/xxHash/pull/749
- https://github.com/Cyan4973/xxHash/pull/759
- https://github.com/Cyan4973/xxHash/pull/838

It's sad but it seems that tipi.build team keep introducing breaking changes to their environment without prior notices (from our point of view).

Since tipi.build is unstable and we don't how to test/stabilize it in our side, I think we should remove `.tipi/` directory for now.    
If someone is using it, they should fork xxHash and add their own "stable" `.tipi/` settings.  If we don't change `.tipi/` directory anymore, they'll be able to `git rebase` easily.

Though I hope that we'll be able to re-introduce it when it'll be stabilized.

PR #845, opened 2023-06-25
added a cmake pkgconfig generation test

Cyan4973:
ensure proper path after advanced example provided by @ilya-fedin : https://github.com/Cyan4973/xxHash/pull/723#issuecomment-1605755702

note : is expected to fail before integration of #723

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -271,12 +271,22 @@ jobs:
     - name: cmake
       run: |
         cd cmake_unofficial
-        mkdir build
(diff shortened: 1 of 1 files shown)


Cyan4973:
The new test fails as expected.
After merge of #723, it's expected to pass.

Cyan4973:
And as expected, following #723 merge, the new test "cmake pkgconfig generation" pass correctly. So I think we are good. This is a property of the `cmake` build system that will be protected vs future code changes.

PR #849, opened 2023-07-02
Detect and display SVE when active

Cyan4973:
display was mistakenly mentioning `NEON` instead.

Codemod inspired by @hzhuang1 's #762.

`SVE` detection should be visible in the `QEMU AARCH64` test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -338,7 +338,8 @@ jobs:
           { name: 'MIPS, gcc-10',    xcc_pkg: gcc-10-mips-linux-gnu,        xcc: mips-linux-gnu-gcc-10,        xemu_pkg: qemu-system-mips,  xemu: qemu-mips-static,    os: ubuntu-20.04, },
 
           { name: 'ARM, gcc-9',      xcc_pkg: gcc-9-arm-linux-gnueabi,      xcc: arm-linux-gnueabi-gcc-9,      xemu_pkg: qemu-system-arm,   xemu: qemu-arm-static,     os: ubuntu-20.04, },
-          { name: 'AARCH64, gcc-9',  xcc_pkg: gcc-9-aarch64-linux-gnu,      xcc: aarch64-linux-gnu-gcc-9,      xemu_pkg: qemu-system-arm,   xemu: qemu-aarch64-static, os: ubuntu-20.04, },
+          # SVE tests require at least gcc-10.1
+          # { name: 'AARCH64, gcc-9',  xcc_pkg: gcc-9-aarch64-linux-gnu,      xcc: aarch64-linux-gnu-gcc-9,      xemu_pkg: qemu-system-arm,   xemu: qemu-aarch64-static, os: ubuntu-20.04, },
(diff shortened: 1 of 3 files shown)


Cyan4973:
Well, it seems that CI SVE tests actually do not trigger or do not detect SVE : 
https://github.com/Cyan4973/xxHash/actions/runs/5439169192/jobs/9890844332?pr=849#step:8:593
> #warning "__ARM_FEATURE_SVE isn't supported. Use SCALAR instead."

The build system then use the `SCALAR` backup, which then succeeds the tests, so there is no failure signal.

Well, that's embarrassing.

Cyan4973:
Well, this little exercise made it possible to control and fix SVE tests in CI, so I guess that's good.

.github/workflows/ci.yml
```diff
@@ -338,7 +338,8 @@ jobs:
           { name: 'MIPS, gcc-10',    xcc_pkg: gcc-10-mips-linux-gnu,        xcc: mips-linux-gnu-gcc-10,        xemu_pkg: qemu-system-mips,  xemu: qemu-mips-static,    os: ubuntu-20.04, },
 
           { name: 'ARM, gcc-9',      xcc_pkg: gcc-9-arm-linux-gnueabi,      xcc: arm-linux-gnueabi-gcc-9,      xemu_pkg: qemu-system-arm,   xemu: qemu-arm-static,     os: ubuntu-20.04, },
-          { name: 'AARCH64, gcc-9',  xcc_pkg: gcc-9-aarch64-linux-gnu,      xcc: aarch64-linux-gnu-gcc-9,      xemu_pkg: qemu-system-arm,   xemu: qemu-aarch64-static, os: ubuntu-20.04, },
+          # SVE tests require at least gcc-10.1
+          # { name: 'AARCH64, gcc-9',  xcc_pkg: gcc-9-aarch64-linux-gnu,      xcc: aarch64-linux-gnu-gcc-9,      xemu_pkg: qemu-system-arm,   xemu: qemu-aarch64-static, os: ubuntu-20.04, },
```
t-mat, line 342: It's 100% okay but there's minor issue.  Since this line also tests Scalar and NEON code path for `gcc-9-aarch64`, disabling this line may miss potential issue for the compiler.
If this will be a real issue, I'll add special conditions for this line.

t-mat:
> Well, that's embarrassing.

🤦 

Right.  We should introduce `-Werror` for all QEMU tests.

Actually, old `armtest` had it though.
https://github.com/Cyan4973/xxHash/blob/0d0e823b385c16f9b6eaa6fb661ef0e5ba6cf316/Makefile#L318-L320

.github/workflows/ci.yml
Cyan4973: Unless we believe something important needs to be tested specifically with `gcc-9-aarch64`, I'm fine losing one target. We already have so many ! And 2 of them are testing `gcc` + `aarch64` already.

Cyan4973:
I just added `-Werror`, as suggested.
Unfortunately, it also results in new errors, that were previously warnings.

Hopefully, fixing them will not push this PR too far from its original goal.

PR #855, opened 2023-07-10
Add missing doxygen style comments to XXH64 family

t-mat:
All comments are simply copied from correspond XXH32 functions.

Now all entries of "XXH64 family" ( `doxygen/html/group___x_x_h64__family.html` ) have their own documents.

image


Cyan4973 (approved):
Great work @t-mat !

Issue #831, opened 2023-05-01
Can documentation be clearer, please?


donaastor:
This technically barely falls under "issues", but I am having an _issue_ trying to figure out who the actual designer of this algorithm is and trying to find the specifications of it in case I want to implement the algorithm myself. Also, the list of functions and how they are used is hard to find too (I didn't find it). So, please, will someone be kind to make it all more accessible?

t-mat:
> who the actual designer of this algorithm

See https://github.com/Cyan4973/xxHash/blob/dev/LICENSE#L1C1-L3

> the specifications of it

See https://github.com/Cyan4973/xxHash/tree/dev/doc

> the list of functions and how they are used

See https://github.com/Cyan4973/xxHash/blob/dev/Doxyfile

For example,

```sh
git clone https://github.com/Cyan4973/xxHash.git
cd xxHash
doxygen
open doxygen/html/index.html
```

Actually, we should maintain it though.

t-mat:
@Cyan4973, to maintain the document, should we include `doxygen` as a part of our CI (and treat its warning as error) ?

So far, I've got the following warnings from `doxygen` command.

```
$ doxygen
xxhash.h:1320: warning: unable to resolve reference to 'XXH3_128bits_reset_withSecret()' for \ref command
xxhash.h:1543: warning: End of list marker found without any preceding list items
xxhash.h:3231: warning: unable to resolve reference to 'XXH_X86DISPATCH' for \ref command
xxhash.h:1140: warning: unable to resolve reference to 'XXH_STATIC_LINKING_ONLY' for \ref command
xxhash.h:1141: warning: unable to resolve reference to 'XXH_IMPLEMENTATION' for \ref command
xxhash.h:1228: warning: unable to resolve reference to 'XXH_STATIC_LINKING_ONLY' for \ref command
xxhash.h:1229: warning: unable to resolve reference to 'XXH_IMPLEMENTATION' for \ref command
xxhash.h:1237: warning: unable to resolve reference to 'XXH3_createState()' for \ref command
xxhash.h:1237: warning: unable to resolve reference to 'XXH3_freeState()' for \ref command
xxhash.h:1164: warning: unable to resolve reference to 'XXH_STATIC_LINKING_ONLY' for \ref command
xxhash.h:1165: warning: unable to resolve reference to 'XXH_IMPLEMENTATION' for \ref command
```

donaastor:
[...]

PR #856, opened 2023-07-10
Add missing doxygen style comments to XXH3 family

t-mat:
Trivial PR which adds basic Doxygen style comment to XXH3 family funcions.

Hopefully we'll be able to introduce [new CI test for documentation](https://github.com/Cyan4973/xxHash/issues/831#issuecomment-1529668262
).

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -1022,20 +1022,50 @@ XXH_PUBLIC_API XXH_PUREF XXH64_hash_t XXH3_64bits_withSecret(XXH_NOESCAPE const 
 typedef struct XXH3_state_s XXH3_state_t;
 XXH_PUBLIC_API XXH_MALLOCF XXH3_state_t* XXH3_createState(void);
 XXH_PUBLIC_API XXH_errorcode XXH3_freeState(XXH3_state_t* statePtr);
+
+/*!
+ * @brief Copies one @ref XXH3_state_t to another.
+ *
+ * @param dst_state The state to copy to.
+ * @param src_state The state to copy from.
+ * @pre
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):

t-mat:
When I change the following line in the `Doxyfile`, `doxygen` still give me a several hundred warnings :cry:

```
        # Until we document everything
-       WARN_IF_UNDOCUMENTED   = NO
+       WARN_IF_UNDOCUMENTED   = YES
```

It seems it'll need more PRs.  So, I think we won't achieve this goal at `0.8.2` timely.

Cyan4973:
Understandable, and fully acceptable outcome.

Cyan4973:
I was wondering, regarding `doxygen` documentation,
what's the better policy ?
Should it be pre-generated inside the repository, typically before a release,
or is it better to let repository users generate it on their side ?
(another option could be to provide it as a release artifact)

t-mat:
I think almost all open source project doesn't provide output of doxygen (or other document generators) as a part of release (or their repository).
Instead of it, they provides the document as web page.

For example, [FLAC](https://xiph.org/flac/) project releases its doxygen-generated document as https://xiph.org/flac/api/index.html

Another example is [SDL](https://github.com/libsdl-org/SDL).  They have [`doxyfile`](https://github.com/libsdl-org/SDL/blob/main/docs/doxyfile).  But they don't provide its output as part of [their release](https://github.com/libsdl-org/SDL/releases/tag/release-2.28.1).

Cyan4973:
Thanks @t-mat , these are very good references.

PR #859, opened 2023-07-12
cmake: bump minimum required to version 3.5.0

scribam:
This fixes the following warning when configuring with CMake 3.27+:
```
CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -341,26 +341,26 @@ jobs:
         cat build/libxxhash.pc | grep "libdir=/usr/lib"
         cat build/libxxhash.pc | grep "includedir=/usr/include"
 
-    - name: cmake minimum version v2.8.12 test
+    - name: cmake minimum version v3.5 test
       run: |
         mkdir -p cmake_bins
         cd cmake_bins
-        wget https://cmake.org/files/v2.8/cmake-2.8.12.2-Linux-i386.tar.gz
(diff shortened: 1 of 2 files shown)


Cyan4973:
Moving the minimum version to `v3.5` is fine by me,
as it's the version supported by Ubuntu Xenial, which is the oldest linux distribution worth supporting in 2023 imho.

But we don't update a minimum version just for the sake of it, since it mechanically reduces the addressable scope of the cmake script. We must "gain" something from this change.
It seems the gain in this case is a higher guarantee to remain supported in some future versions of `cmake`, which may drop support for versions older than `v3.5` in some unspecified future.
That's a good enough reason I believe.

The PR is also well done, updating the cmake min version test introduced by @t-mat, and exploiting the increased minimum version to slightly reduce complexity of the script.

The only remaining question is if it's worth updating it "right now", or can it wait after the imminent release. 

I would tend to lean towards the latter, as there is no immediate risk (the only downside is a warning message for some really recent versions of `cmake`) but increasing the minimum version may prevent some older systems from building `xxhash` with `cmake`. While it's probably not major, it still is a negative.

PR #860, opened 2023-07-12
added a compilation warning message for XXH_OLD_NAMES

Cyan4973:
as suggested by @t-mat,
warning about the upcoming deprecation of this feature.


Cyan4973:
uh oh ... I expected `XXH_OLD_NAMES` to be disabled by default,
but maybe it's not ?...

_edit_ : arf, it was just a formatting error ...

PR #867, opened 2023-07-15
fix: Intel compiler's warning for XXH_ASSERT()

t-mat:
Intel compiler expects `unsigned char` as an argument of `__builtin_assume()`.
Without this patch, it reports the following warnings to all `XXH_ASSERT()`s.

```
In file included from xxhash.c(43):
xxhash.h(6737): warning #2259: non-pointer conversion from "int" to "unsigned char" may lose significant bits
      XXH_ASSERT(secretBuffer != NULL);
      ^
```

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -2138,7 +2138,11 @@ static void* XXH_memcpy(void* dest, const void* src, size_t size)
 #  include <assert.h>   /* note: can still be disabled with NDEBUG */
 #  define XXH_ASSERT(c)   assert(c)
 #else
-#  define XXH_ASSERT(c)   XXH_ASSUME(c)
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):

Cyan4973:
This makes me think :
do we have an `icc` test in CI ?

t-mat:
> do we have an icc test in CI ?

Yeah.  Sometimes we have `icc` related issue, it should be in the CI.

Since there's official blog article, it seems it can be done relatively easily.

[Adding Intel oneAPI toolkits to GitHub actions | Intel DevMesh](https://devmesh.intel.com/posts/991253/adding-intel-oneapi-toolkits-to-github-actions)

PR #868, opened 2023-07-16
fix: hide internal macros from doxygen

t-mat:
This trivial PR hides the following internal macros from the document

https://xxhash.com/doc/v0.8.2/group___x_x_h32__family.html

image


t-mat:
In another PR, I'll hide x86 dispatcher intenals from doxygen
https://xxhash.com/doc/v0.8.2/group__dispatch.html

Cyan4973 (approved):

PR #875, opened 2023-07-22
Improve performance of the collision test filter

Cyan4973:


Cyan4973:
`mingw` compilation doesn't seem to complain when requested to use the `c11` standard,
but then doesn't provide `aligned_alloc()` which is [part of `C11` standard](https://en.cppreference.com/w/c/memory/aligned_alloc) ?

Am I missing something ?

t-mat:
Perhaps they still have this error?
https://github.com/ebassi/graphene/issues/83

I put the error log for the record.

https://github.com/Cyan4973/xxHash/actions/runs/5633492455/job/15262446792#step:5:640

```
cc -Werror -Wall -Wextra -Wconversion -std=c11 -g3 -O0 -DDEBUG -I ./ -I ../../ -I allcodecs/  -c -o main.o main.c
                                                                      
main.c: In function 'create_Filter':
main.c:337:24: error: implicit declaration of function 'aligned_alloc' [-Werror=implicit-function-declaration]
  337 |     Filter* const bf = aligned_alloc(cacheline_size, bfsize);  // requires C11
      |                        ^~~~~~~~~~~~~
main.c:55:1: note: include '<stdlib.h>' or provide a declaration of 'aligned_alloc'
   54 | #include "sort.hh"    /* sort64 */
  +++ |+#include <stdlib.h>
   55 | 
main.c:337:24: error: incompatible implicit declaration of built-in function 'aligned_alloc' [-Werror=builtin-declaration-mismatch]
  337 |     Filter* const bf = aligned_alloc(cacheline_size, bfsize);  // requires C11
      |                        ^~~~~~~~~~~~~
main.c:337:24: note: include '<stdlib.h>' or provide a declaration of 'aligned_alloc'
cc1.exe: all warnings being treated as errors
mingw32-make[1]: *** [<builtin>: main.o] Error 1
```

Cyan4973:
Indeed, 
the compiler seems to believe that `aligned_alloc()` is defined within `<stdlib.h>`,
and therefore recommends including it,
but `#include <stdlib.h>` has already already done.
Moreover, even when manually declaring `aligned_alloc()`, link stage fails afterwards,
it points to an _absence_ of the symbol.

The compiler proper is probably `C11` ready, but the bundled standard libs are not.

PR #878, opened 2023-08-15
fix build with Apple clang 4.2 on Mac OS X Lion

thestr4ng3r:
Discovered here: https://trac.macports.org/ticket/67963
Related: https://github.com/macports/macports-ports/pull/19899

Seems like the old clang can't handle the function-like macro definition to `__has_include`.


Cyan4973 (approved):

Cyan4973:
This is a surprising limitation,
but well, the fix is simple enough and doesn't seem to degrade other versions,
so that's fine.

PR #880, opened 2023-08-24
xxhsum: use pledge(2) on OpenBSD

klemensn:
xxhsum(1) only ever reads files or standard input, i.e. it never
creates or writes any;  the "rpath" promise could be dropped at a
later point (when no files are given), but that needs more patching.

sanity_test does not even read files, just standard input.

https://man.openbsd.org/pledge.2

### DIFF
--- a/cli/xsum_os_specific.c
+++ b/cli/xsum_os_specific.c
@@ -69,6 +69,11 @@ static int XSUM_IS_CONSOLE(FILE* stdStream)
  || defined(__DJGPP__) \
  || defined(__MSYS__) \
  || defined(__HAIKU__)
+#  ifdef __OpenBSD__
+#    include <errno.h>       /* errno */
+#    include <string.h>      /* strerror */
+#    include "xsum_output.h" /* XSUM_log */
+#  endif
 #  include <unistd.h>   /* isatty */
 #  define XSUM_IS_CONSOLE(stdStream) isatty(fileno(stdStream))
 #elif defined(MSDOS) || defined(OS2)
@@ -135,6 +140,16 @@ static int XSUM_stat(const char* infilename, XSUM_stat_t* statbuf)
 #ifndef XSUM_NO_MAIN
 int main(int argc, const char* argv[])
 {
+#ifdef __OpenBSD__
+    /*
+     * xxhsum(1) does not create or write files, permit reading only.
+     */
+    if (pledge("stdio rpath", NULL) == -1) {
+        XSUM_log("pledge: %s\n", strerror(errno));
+        return 1;
+    }
+#endif
+
     return XSUM_main(argc, argv);
 }
 #endif


klemensn:
These patches are already applied to OpenBSD's official port of xxHash, upstreaming would reduce local maintenance.

They try to be consistent with the existing coding style, i.e. `XSUM_log()` is used in regular fashion.
A previous version used just [err(3)](https://man.openbsd.org/err.3) to simplify code.

t-mat:
Hi @klemensn,  thanks for the patch.

I have two questions:

(1) Could you elaborate when we need to add `pledge()` or not?

Specifically, I don't understand why we should add `pledge()` for `sanity_test.c` (and possibly other test executables: `tests/*.c`).
Since `pledge()` mainly provides benefits to actual artifact of the project and end users, it seems we don't need to use it for development/test/helper programs.
Please let us know why you and OpenBSD team needed to add `pledge()` to `sanity_test.c`, and didn't need to add other test programs.

(2) We need decent CI test for OpenBSD environment.
Because we may make mistake.  And without CI test, we can't notice it.
So, it'll be found in downstream/OpenBSD side.  Therefore, it may bring extra burden for OpenBSD team.

Since we don't have CI test for OpenBSD environment, this patch will be fragile.
Do you have any idea to add OpenBSD environment to our CI?

klemensn:
> Hi @klemensn, thanks for the patch.
> 
> I have two questions:
> 
> (1) Could you elaborate when we need to add `pledge()` or not?

Unless xxhsum(1) gains the ability to, e.g. create/write files, open a socket or do anything else described in [pledge(2)](https://man.openbsd.org/pledge.2), nothing will have to be added or changed.

pledge(2) is an intrinsic enforcement mechanism for authors to specify the intended behaviour of programs towards the kernel, such that they get killed when their "promises" get broken (read: bugs) rather than keep running.

> Specifically, I don't understand why we should add `pledge()` for `sanity_test.c` (and possibly other test executables: `tests/*.c`). Since `pledge()` mainly provides benefits to actual artifact of the project and end users, it seems we don't need to use it for development/test/helper programs. Please let us know why you and OpenBSD team needed to add `pledge()` to `sanity_test.c`, and didn't need to add other test programs.

I added it there as means of verification, since that is what a test suite does;  that hunk obviously has no effect on xxhsum(1), so we might as well drop it.

> (2) We need decent CI test for OpenBSD environment. Because we may make mistake. And without CI test, we can't notice it. So, it'll be found in downstream/OpenBSD side. Therefore, it may bring extra burden for OpenBSD team.

One downstream patch less is usually a plus.  Upon release, ports maintainers/developers already check and test new code anyway, so in case any future change should break on OpenBSD, or rather due to pledge, I'm happy to step in and fix it.
 
> Since we don't have CI test for OpenBSD environment, this patch will be fragile. Do you have any idea to add OpenBSD environment to our CI?

I am not aware of OpenBSD support in CI.

Cyan4973:
This looks fine and well done,
just a little detail: please document why these lines are desirable,
as it's not necessarily obvious for a non-openBSD contributor.
This will be helpful for maintenance.

Essentially provide the same argument as you already did in your comment above,
in a (summarized) code comment version.

cli/xsum_os_specific.c
```diff
@@ -135,6 +140,13 @@ static int XSUM_stat(const char* infilename, XSUM_stat_t* statbuf)
 #ifndef XSUM_NO_MAIN
 int main(int argc, const char* argv[])
 {
+#ifdef __OpenBSD__
```
Cyan4973, line 143: Please explain why `pledge()` is desirable and why it's only for OpenBSD.

tests/sanity_test.c
```diff
@@ -639,6 +639,13 @@ int main(int argc, const char* argv[])
     (void) argc;
     (void) argv;
 
+#ifdef __OpenBSD__
```
Cyan4973: I understand from your comment that this is just meant for testing the following code snippet,
as a mirror of the other instance in `xxhsum`'s `main()`.
However, since we already test `xxhsum` as part of CI and from `make check`,
what do we gain by checking this code snippet a second time in `sanity_test` ?

Consider for example a scenario of a future modification of these lines by some future contributor.
It seems it implies that both instances of the code snippet should be kept in sync manually.

klemensn: Sure, I'll just drop this hunk.

cli/xsum_os_specific.c
klemensn: A comment now explains the promises, but I've omitted further reasoning how pledge(2) is an OpenBSD-only syscall.

Cyan4973 (approved):

PR #881, opened 2023-08-24
chore: Improve document

t-mat:
Apply same format for all doxygen documents.
Now, every document has the following structure.

```
Signature of the function

@brief   description
@params  parameters
@pre     preconditions
@returns possible return values
@note	 notes
@see-also
```

### Old style
image

### New style
image


Cyan4973 (approved):
Thanks @t-mat, it looks great !

PR #906, opened 2023-12-23
Add fuzzing harness

nathaniel-brough:


Cyan4973:
I'm rather favorable to this effort, given the increased security coverage offered by oss-fuzz,

note though that the current proposal (this PR) fails CI tests.
This should be fixed first.

Makefile
```diff
@@ -315,6 +316,12 @@ test-xxhsum-c: xxhsum
 	echo "00000000  test-expects-file-not-found" | ./xxhsum -c -; test $$? -eq 1
 	@$(RM) .test.*
 
+LIB_FUZZING_ENGINE?="-fsanitize=fuzzer"
+ifeq '' '$(findstring clang++,$(CXX))'
```
Cyan4973: Is this `ifeq` correct ?
I thought it would require parenthesis

nathaniel-brough:
> I'm rather favorable to this effort, given the increased security coverage offered by oss-fuzz,
> 
> note though that the current proposal (this PR) fails CI tests. This should be fixed first.

Great! Working on getting everything green, it's been a while since I've been working with Makefiles so I'm a little rusty.

Makefile
nathaniel-brough: Your probably right. Just fiddling with this to get it working.

Cyan4973 (approved):

PR #917, opened 2024-01-30
add sparc emulation to Github Actions CI

Cyan4973:


Cyan4973:
Works fine :
https://github.com/Cyan4973/xxHash/actions/runs/7704716779/job/20997501954?pr=917

though the CLI is unable to determine the cpu type, so it's called "unknown big endian".
Something to improve later on if it ever matters.

PR #931, opened 2024-04-01
Correct check for Windows compilation

pps83:
`_WIN32` is the proper check for compilation for Windows. `WIN32` is defined when `Windows.h` is included or could be defined by some projects

### DIFF
--- a/cli/xsum_os_specific.c
+++ b/cli/xsum_os_specific.c
@@ -79,7 +79,7 @@ static int XSUM_IS_CONSOLE(FILE* stdStream)
 #elif defined(MSDOS) || defined(OS2)
 #  include <io.h>       /* _isatty */
 #  define XSUM_IS_CONSOLE(stdStream) _isatty(_fileno(stdStream))
-#elif defined(WIN32) || defined(_WIN32)
+#elif defined(_WIN32)
 #  include <io.h>      /* _isatty */
 #  include <windows.h> /* DeviceIoControl, HANDLE, FSCTL_SET_SPARSE */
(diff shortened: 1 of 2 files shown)


pps83:
some refs: https://stackoverflow.com/a/74084728/468725

Cyan4973 (approved):

pps83:
similar fix should be merged into zstd (plus all defines for `WIN32` from projects should be removed, but that's irrelevant).
as a side note, what's the reason xxhash in lz4 is stuck using version form 6 years ago, why isn't it updated?

Cyan4973:
> what's the reason xxhash in lz4 is stuck using version form 6 years ago, why isn't it updated?

`liblz4` proper only requires `XXH32`, which hasn't evolved.

We _can_ still update `libxxhash` though. Just not critical.

pps83:
> `liblz4` proper only requires `XXH32`, which hasn't evolved.

`zstd` stays more or less up to date with xxhash because it uses `XXH64`, is that correct? From xxhash the fastest one is `XXH3`, what percentage roughly is hash compute of zstd? I'm mainly interested in decompression speed, so I'm ok to break zstd compat as it's used only in my own code and to modify it to use `XXH3` or even avoid hashing completely. Is there any issues to update zstd hashing to XXH3 or to a null hash?

Cyan4973:
You can check the decompression speed with and without the checksum, on both the library and the CLI sides.
This will tell you how much the checksum costs in your case.

The result is very dependent on the data and the hardware system, I would expect something in the 10% range but variance is high.

PR #932, opened 2024-04-15
underscore attributes to not collide with macros

crosdahl:
Attributes without underscores expand improperly when already defined as "language keywords"

```c
#define noinline __attribute__((__noinline__))

// ok
noinline void f() {
        ...
}

// ok
__attribute__((__noinline__)) void g(void) {
        ...
}```

// error
__attribute__((noinline)) void h(void) {
        ...
}```

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -333,7 +333,7 @@ extern "C" {
    /* make all functions private */
 #  undef XXH_PUBLIC_API
 #  if defined(__GNUC__)
-#    define XXH_PUBLIC_API static __inline __attribute__((unused))
+#    define XXH_PUBLIC_API static __inline __attribute__((__unused__))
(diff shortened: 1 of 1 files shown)


Cyan4973:
Do you imply some of these attributes were effectively inactive before this patch ?

crosdahl:
No, the non-underscored attribute names (e.g. `noinline`), clash with my usage of `#define noinline __attribute__((__noinline__))`. Alias attribute names with underscores exist so that you can use them even if they have been defined to something else. Currently the code depends on noinline,packed,aligned,noescape,etc not being a macro, forbidding code that uses xxhash to use those identifiers

Cyan4973 (approved):
lgtm

PR #949, opened 2024-06-07
Update xxhash_spec.md

davesnx:
Fix typo in xxhash_spec


Cyan4973:
I don't think it's a typo.

Since, in the previous loop, the code consumed input 8 bytes at a time,
at the end of this loop, only `[0-7]` bytes remain.
Consequently the `>=4` may happen once at a maximum.

davesnx:
Totally, silly me.

PR #976, opened 2024-10-30
add GCC >=15 __builtin_stdc_rotate_left suppor to rotl

crrodriguez:
GCC does not support the clang builtins but does have type generic, type-checking __builtin_stdc_rotate_left.

The only advantage of this is that the builtin evaluates the arguments only once.


Cyan4973 (approved):
Looks good to me!

PR #982, opened 2024-12-04
removed ARM-on-Visual and macos-12 tests

Cyan4973:
Visual 2022 does no longer support ARM target.
`macos-12` runners are disabled by Github.


Cyan4973:
And of course, as the Visual+ARM CI test get fixed (aka removed),
another thing breaks, the `macos-12` test...

Cyan4973:
> GitHub Actions is starting the deprecation process for macOS 12. While the image is being deprecated, You may experience longer queue times during peak usage hours. Deprecation will begin on 10/7/24 and the image will be fully unsupported by 12/3/24

PR #994, opened 2025-01-04
Update MANDIR for FreeBSD

sunpoet:
FreeBSD uses share/man/ rather than man/ now.

Reference:	https://cgit.freebsd.org/ports/tree/devel/xxhash/Makefile#n17

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -565,7 +565,7 @@ else
 PKGCONFIGDIR ?= $(LIBDIR)/pkgconfig
 endif
 
-ifneq (,$(filter $(UNAME),OpenBSD FreeBSD NetBSD DragonFly SunOS))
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):

Cyan4973:
> FreeBSD uses share/man/ rather than man/ now.

When you state "now",
does that mean it's rather recent,
and would it actually depend on the FreeBSD version ?

sunpoet:
share/man became a valid path for manpage since [Jan 2020](https://cgit.freebsd.org/ports/commit/?id=003a571d1d6585196545295efc181514f171c4c4). And we converted the whole ports tree to share/man around last March. And you don't have to check FreeBSD version since the change here does not affect the ports tree. I should do it earlier so that we could drop the patch along with the 0.8.3 update of devel/xxhash.

sunpoet:
Thanks!

PR #1000, opened 2025-01-21
Use _M_X64 only without mixing with _M_AMD64

pps83:


Cyan4973:
Note: CI errors seem unrelated to this PR, and should disappear after rebasing on top of latest `dev` branch.

Cyan4973:
`_M_X64` seems indeed redundant with `_M_AMD64`,
so I agree with the simplification keeping only one.

Is there a specific reason you would prefer `_M_X64` ?
I was under the impression that `_M_AMD64` was more common.

pps83:
> I was under the impression that `_M_AMD64` was more common.

No, `_M_X64` is more common.

t-mat:
Some notes for future contributors.

As for MSVC

> ## MACRO Revisited
> https://learn.microsoft.com/en-us/archive/blogs/reiley/macro-revisited  
> _M_AMD64 and _M_X64 are equivalent.

Not sure about `icc` though.

It proves nothing, but according to sourcegraph.com:

- [_M_AMD64](https://sourcegraph.com/search?q=context%3Aglobal+_M_AMD64+count%3Aall&patternType=keyword&sm=0&groupBy=repo) -> `67.7k results`
- [_M_X64](https://sourcegraph.com/search?q=context:global+_M_X64+count:all&patternType=keyword&sm=0) -> `152.1k results`

Also, it seems `microsoft/STL` uses `_M_X64` as a canonical symbol for x86-64.

- https://github.com/microsoft/STL/blob/4119eb8436e2a084d8364509fcec3ee643e6d220/stl/inc/atomic#L214

```C++
#elif defined(_M_IX86) || defined(_M_X64) // ^^^ ARM64/ARM64EC/HYBRID_X86_ARM64 / x86/x64 vvv
```

pps83:
> Not sure about `icc` though.

if that's old intel compiler, then likely they would blacklist anything mentioning amd :)
new intel compiler uses clang front-end

Cyan4973 (approved):

pps83:
ping @Cyan4973 seems like this is ok to merge as well

PR #1004, opened 2025-01-21
attempt to silence a recent ubsan warning with code reorganization

Cyan4973:
apparently, `UBSAN` uses the prototype of the inner inlined function instead of the outer shell of the function.
Let's see if one indirection level is enough to fix this.


Cyan4973:
OK, so `clang`'s `ubsan` has become sensible to `noescape` attribute and now complains when there is any difference.

Cyan4973:
OK, all tests fixed, CI signal is green.
Let's fix the other PRs by rebasing on top of this one.

Issue #383, opened 2020-06-01
strict aliasing violations lead to bad hash results


anholt:
In Mesa we just started using xxhash (inlined in many callsites), and the compiler in some circumstances produces different hash results that appear to be due to the strict aliasing violations.

https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/5271

Cyan4973:
Thanks for notification @anholt .

I was looking at the commit message, but was unable to determine with certainty if the problem comes from aliasing or from unaligned reads.
The commit message talks about aliasing, but the attached fix [effectively impacts unaligned reads](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/5271/diffs?commit_id=ea35dde0d3e4a940d27bd3d30ec0070c0c044877) and nothing else.

anholt:
I needed both of the defines to avoid u32 access without getting deeper into xxh changes:

- make the unaligned path use memcpy
- always take the unaligned path, since the aligned fastpath uses a u32 deref instead of the read32 memory accessor and didn't have a memcpy wrapper knob to tweak.

My actual failing data is 32 bit aligned in pointer address.

anholt:
(as a side note, the platform affected was arm64)

Cyan4973:
The unaligned path is supposed to use `memcpy()` by default. So I'm surprised you needed to set `XXH_FORCE_MEMORY_ACCESS == 0`, I would expect this value to be already set at `0`.

As for the aligned data shortcut, well that's an interesting (and unexpected) development.

The aligned data shortcut will read data directly as `xxh_u32*` on the ground that the pointer address is 32-bit aligned, which has been previously verified.

However, it's true that, if the buffer was previously filled using another pointer type, the compiler is free to consider that both actions are unrelated, and even order them differently.

The thing that was supposed to protect such a scenario from happening is that there are several layers of invocations using `const void*` pointers before reaching the function that actually reads the data, `XXH_readLE32_align()`, which uses `xxh_u32*`.
And if do remember correctly, `void*` pointers and `char*` pointers are supposed immune to strict aliasing rule, meaning that the compiler shall never blindly consider that the pointer doesn't alias anything.

Now, if the compiler just rips through the layers of indirection due to inlining, and goes directly to the 32-bit read instruction, _then_ it could separate production and ingestion actions, and re-order them. 

[...]

PR #1013, opened 2025-02-13
Use may_alias with unaligned reads to fix miscompilation on GCC

lassipulkkinen:
When `XXH_FORCE_MEMORY_ACCESS==1`, which is the default on supported compilers unlike stated in the README, a strict aliasing violation occurs in `XXH_read64`, resulting in miscompilation on GCC (but not Clang) in some oddly specific circumstances.

The following code reproduces the problem on x86_64 GCC 14.2.1 when compiled with -O3:

```c
#define XXH_INLINE_ALL

#include <inttypes.h>
#include <stdio.h>
#include <xxhash.h>

int main() {
	// it seems this has to be exactly 24 bytes.
	union {
		char x[24];
		// force 8-byte alignment without making
		// aliasable with uint64_t.
		void *y[3];
	} data = {.x = "garblegarblegarblegarble"};
	uint64_t hash = XXH64(&data, sizeof(data), 0);
	printf("%016"PRIx64"\n", hash);
	return 0;
}
```

A bogus `-Wuninitialized` warning is produced if enabled, and the resulting program outputs an incorrect hash.

While this definitely looks like a compiler bug in some ways, I see no reason to assume `aligned(1)` alone should excempt the type from aliasing restrictions regardless, and adding `may_alias` does fix the problem.

The separate strict aliasing bug when `XXH_FORCE_ALIGN_CHECK==1`, discussed in #383, still remains.

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -2661,7 +2661,7 @@ typedef union { xxh_u32 u32; } __attribute__((__packed__)) unalign;
 #endif
 static xxh_u32 XXH_read32(const void* ptr)
 {
-    typedef __attribute__((__aligned__(1))) xxh_u32 xxh_unalign32;
+    typedef __attribute__((__aligned__(1))) __attribute__((__may_alias__)) xxh_u32 xxh_unalign32;
     return *((const xxh_unalign32*)ptr);
 }
 
@@ -3374,7 +3374,7 @@ typedef union { xxh_u32 u32; xxh_u64 u64; } __attribute__((__packed__)) unalign6
 #endif
 static xxh_u64 XXH_read64(const void* ptr)
 {
-    typedef __attribute__((__aligned__(1))) xxh_u64 xxh_unalign64;
+    typedef __attribute__((__aligned__(1))) __attribute__((__may_alias__)) xxh_u64 xxh_unalign64;
     return *((const xxh_unalign64*)ptr);
 }
 


Cyan4973:
`XXH64()` exposes a `void*` pointer as input argument,
so it should be enough to allow aliasing.

However, with inlining via `XXH_INLINE_ALL`, I guess the compiler can pass through this interface layer, and access directly the code below it, skipping the `void*` intermediate stage. 
I don't know C well enough to tell if the compiler is allow to do that (i.e. ignore `void*` contract), or is well within its right.
Either way, I can see how this can be an immediate problem.

The proposed fix is simple enough, and it should do exactly as advertised, i.e. allow aliasing, even when inlining.

That being said, this issue has not been exposed in CI so far, showing that we probably need to amend the test suite to produce the problem, observe its fix, and ensure the property remains guaranteed over time.

Since you also provide a test case in the issue description, I'll try to use that as a basis for the new test.

thesamesam:
The compiler is allowed to do this. You have to cast back to the original type (or a compatible one) for it to be legal, i.e. `void*` isn't a firewall for aliasing.

Using `may_alias` is fine as a workaround, just not all compilers support it.

Cyan4973:
I've been trying to reproduce the problem, by using exactly the example provided in this PR.

It's still not clear to me if this is a compilation issue (warning produced during compilation) or a runtime issue (strict aliasing violated at runtime, hopefully detectable with the UB sanitizer). 
I tried both, but with no reproduction so far.

Here is my exact compilation line :

```
cc -g -fsanitize=undefined -O3 -Wall -Wuninitialized  -DXXH_FORCE_MEMORY_ACCESS=1   test_alias.c   -o test_alias
```

I've been using both gcc-9 and gcc-10.

The system I'm testing on doesn't have access to `gcc-14`.
It could be that this version of `gcc` is required to reproduce the issue.
In which case, I will need to test this scenario on another system.

_edit_: another test with `mingw64`, using `gcc-15.1` compiler,
still unable to reproduce the mentioned warning.

thesamesam:
UBSAN doesn't detect aliasing problems, though it will detect unaligned access. Clang has a new sanitizer for this (`-fsanitize=type`) but it's very primitive for now.

I think the aliasing violation is clear here though, a `char*` gets read as `xxh_unalign64*`.

Cyan4973:
Finally, I could reproduce the warning using `gcc-13.3.0`.
Let's now try to create a new CI test based on this warning.

Cyan4973:
OK, we now have a failing test in #1047, 
so if all goes according to expectations, 
rebasing the failing test on top of this patch will fix the test.

PR #1023, opened 2025-04-28
Prefer NEON over SVE

Nicoshev:
We've observed the NEON implementation perform 25% to 40% better than the SVE one.

aarch64 builds will now prefer the NEON variation instead of the SVE one.

The SVE version can still be used by explicitly defining XXH_SVE

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -3995,9 +3995,7 @@ XXH_PUBLIC_API XXH64_hash_t XXH64_hashFromCanonical(XXH_NOESCAPE const XXH64_can
 #endif
 
 #ifndef XXH_VECTOR    /* can be defined on command line */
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks @Nicoshev ,
it's indeed a good reason to prefer `NEON` by default.

I presume a logical consequence is that `SVE` code can now only be reached when explicitly requested at compile time with `-DXXH_VECTOR=XXH_SVE` ?

Nicoshev:
@Cyan4973 Yes, correct. Having SVE support implies having NEON support, thus XXH_NEON will be selected by default in this case.

If tuning performance is desired for a CPU with a very wide SVE register, XXH_SVE can be supplied to override the default choice.

Cyan4973 (approved):

PR #1047, opened 2025-06-18
added test_alias

Cyan4973:
expected to fail due to #1013


Cyan4973:
as expected, the new test fails.

Rebasing the test on top of #1013 fix should fix the test

Cyan4973:
And indeed, tests are now passing. So this validates #1013.

PR #1056, opened 2025-07-28
fix DISPATCH=1 on non-x86 target

Cyan4973:
complete #1055:
the `DISPATCH` value would remain set to 1 when set on the right side of `make`.

Also: displays a warning, so that users understand what is going on.


eworm-de:
Ups, missed that it was not sufficient... Did non have anything non-`x86_64` around.

Thanks a lot for the follow-up (db36ccbeea58bd6ee7dd55e77130f1655d5ff6db)!

PR #1068, opened 2025-09-08
Update XXHASH_DIR path retrieval in CMakeLists.txt

stefan-sedlak:
Make compiler flag `-I` shorter / cleaner. Change from `/path/xxHash/build/cmake/../..` to `/path/xxHash`

### DIFF
--- a/build/cmake/CMakeLists.txt
+++ b/build/cmake/CMakeLists.txt
@@ -14,10 +14,8 @@ cmake_minimum_required (VERSION 3.10 FATAL_ERROR)
 # absolute, cleaner ones. For example:
 #   /path/xxHash/build/cmake/../..   ->   /path/xxHash
 #
-# We use get_filename_component() twice to step up two levels and store
-# the result in XXHASH_DIR.
-get_filename_component(XXHASH_DIR "${CMAKE_CURRENT_SOURCE_DIR}" DIRECTORY)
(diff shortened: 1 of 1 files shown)


build/cmake/CMakeLists.txt
```diff
@@ -7,7 +7,8 @@
 
 cmake_minimum_required (VERSION 3.10 FATAL_ERROR)
 
-set(XXHASH_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../..")
+get_filename_component(XXHASH_DIR "${CMAKE_CURRENT_SOURCE_DIR}" DIRECTORY)
+get_filename_component(XXHASH_DIR "${XXHASH_DIR}" DIRECTORY)
```
Cyan4973: I understand the goal,
and agrees that this construction do the work.

But it also feels a bit hackish, because understanding what the code is doing is less obvious.

To be fair, I have no better solution to offer, given `cmake` limitations,
so I think it deserves some inline comment, explaining what is going on here,
so that future reviewers easily understand what these 2 lines are trying to achieve.

stefan-sedlak: As I am not native speaker I used LLM to tidy the comment. It came with advise to make it in one line (-:

Cyan4973 (approved):
Yes, that's much better

PR #1073, opened 2025-10-01
Use CMAKE_SYSTEM_PROCESSOR to detect the target architecture

### LINKED ISSUE #1072: CMake platform detection uses the build machine to decide on architecture
When cross compiling, the cmake build looks at the machine running cmake to decide whether or not build the dispatch code depending on that here:

```
# detect architecture for DISPATCH mode
CMAKE_HOST_SYSTEM_INFORMATION(RESULT PLATFORM QUERY OS_PLATFORM)
message(STATUS "Architecture: ${PLATFORM}")

# libxxhash
if((DEFINED DISPATCH) AND (DEFINED PLATFORM))
  # Only support DISPATCH option on x86_64.
  if(("${PLATFORM}" STREQUAL "x86_64") OR ("${PLATFORM}" STREQUAL "AMD64"))
```

Which obviously does not build the dispatch codepath when cross compiled on an arm64 machine for an x86_64 machine, and tries to build the x86_64 code when cross compiling for arm64 on an x86_64 machine.

I think the cmake code should look at `CMAKE_SYSTEM_PROCESSOR` to make this decision, and should be a quick and easy PR but I want to make sure I'm not missing anything.

FatihBAKIR:
Fixes https://github.com/Cyan4973/xxHash/issues/1072

### DIFF
--- a/build/cmake/CMakeLists.txt
+++ b/build/cmake/CMakeLists.txt
@@ -80,7 +80,7 @@ include(CMakeDependentOption)
 CMAKE_DEPENDENT_OPTION(BUILD_SHARED_LIBS "Build shared libraries" ON "NOT XXHASH_BUNDLED_MODE" OFF)
 
 # detect architecture for DISPATCH mode
-CMAKE_HOST_SYSTEM_INFORMATION(RESULT PLATFORM QUERY OS_PLATFORM)
+set(PLATFORM ${CMAKE_SYSTEM_PROCESSOR})
 message(STATUS "Architecture: ${PLATFORM}")
 
 # libxxhash


Cyan4973:
This is likely a good idea,
but how do we prove that it works as intended?

Ideally, we should introduce a new test scenario in CI that demonstrates the failure prior to this patch and its resolution afterward.
More importantly, the purpose of CI is to ensure that this behavior remains correct in the future.

Cyan4973:
#1084 implements the cmake cross-compilation scenario, 
and the test fails without the fix, as expected.

And this fix, fixes it, so I guess we are good to go now.

Cyan4973 (approved):

PR #1084, opened 2025-12-02
add a cmake cross-compilation test

Cyan4973:
according to #1073,
this should fail,
due to confusion in host architecture.

If it does fail as expected, then we'll have a test to prove that #1073 works


Cyan4973:
The test fails as expected

Cyan4973:
And with the fix, the test passes. So I guess that proves it.

Cyan4973:
rebasing the test on top of `dev` which now includes #1073,
so it should work fine now.

PR #1088, opened 2025-12-15
cli: add seed parameter

d06i:


d06i:
I don't understand why the job is failing. Can you clarify the issue? Thank you.

Cyan4973:
It mostly complains about:
```
cli/xxhsum.c:
make: *** [Makefile:531: trailingWhitespace] Error 1
```

This test is meant to fail when a trailing whitespace happens to be present somewhere in the source code.

Cyan4973:
In #1089, the test `trailingWhitespace` is updated, to also provide the line number.

It gives the following result on this PR:

```
cli/xxhsum.c:1639:
Error: trailing whitespace detected
make: *** [trailingWhitespace] Error 1
```

I checked and indeed this line has white spaces.
This was not visible on the github interface, it's necessary to look directly into the file with an editor.

d06i:
Thanks for finding the CI issue. I can finally sleep now.

cli/xxhsum.c
```diff
@@ -210,8 +210,8 @@ static int XSUM_algoBitmask_Accepts(XSUM_U32 algoBitmask, AlgoSelected parsedLin
 *  File Hashing
 **********************************************************/
 
-#define XXHSUM32_DEFAULT_SEED 0                   /* Default seed for algo_xxh32 */
-#define XXHSUM64_DEFAULT_SEED 0                   /* Default seed for algo_xxh64 */
+XSUM_U32 XXHSUM32_DEFAULT_SEED = 0;                   /* Default seed for algo_xxh32 */
```
Cyan4973: We typically reserve `ALL_CAPS` for macros, eventually constants.

Since this is a mutable variable, I would rather use lowercase here.
Also, I use the convention to name global variables with ` g_` prefix.
You still can remap it to a macro afterwards, which would be appropriate here.

Even though these variables are global, there is no reason to expose them outside of this unit.
So we should make them `static`.

Finally, since currently your patch only allows 32-bit seeds, we may not need more than one variable.

Which gives :

```
static XSUM_U32 g_default_seed_u32 = 0;
#define XXHSUM32_DEFAULT_SEED g_default_seed_u32
#define XXHSUM64_DEFAULT_SEED g_default_seed_u32
```

cli/xxhsum.c
```diff
@@ -1636,6 +1637,18 @@ XSUM_API int XSUM_main(int argc, const char* argv[])
                 XSUM_logLevel--;
                 break;
 
+            /* Modify seed */
+            case 's': argument++;
+                switch( algo ){
```
Cyan4973: If you follow above change, there is no need for a `switch()` here.
Just:
```
g_default_seed_u32 = XSUM_readU32FromChar(&argument);
```

cli/xxhsum.c
Cyan4973: Note: another option: since the global variable is only used (mutated and read) within `XSUM_main()`,
maybe there is no need to make it global.
It could be a local variable of `XSUM_main()`,
thus limiting its scope
which is always preferable for maintenance.

Cyan4973 (changes requested):

cli/xxhsum.c
d06i: Should I add a `strtoull` based helper function for `u64` seed or should I keep using `XSUM_readU32FromChar` only?

cli/xxhsum.c
d06i: > Note: another option: since the global variable is only used (mutated and read) within `XSUM_main()`, maybe there is no need to make it global. It could be a local variable of `XSUM_main()`, thus limiting its scope which is always preferable for maintenance.

You're right but the seed is used by `XSUM_hashStream` and so I kept it global for now. 

```c
static Multihash
XSUM_hashStream(FILE* inFile,
                AlgoSelected hashType,
                void* buffer, size_t blockSize)
{
    XXH32_state_t state32;
    XXH64_state_t state64;
    XXH3_state_t  state3;

    /* Init */
    (void)XXH32_reset(&state32, XXHSUM32_DEFAULT_SEED);
    (void)XXH64_reset(&state64, XXHSUM64_DEFAULT_SEED);
    (void)XXH3_128bits_reset(&state3);
```

cli/xxhsum.c
Cyan4973: It depends if you want to offer 64-bit seed too.
But you could start simpler (32-bit only), and expand later, when the need arises.

cli/xxhsum.c
Cyan4973: Another option could be to pass that variant as a parameter of `XSUM_hashStream()`,
but the simplicity argument is also valid.

cli/xxhsum.c
```diff
@@ -251,9 +251,10 @@ XSUM_hashStream(FILE* inFile,
     XXH3_state_t  state3;
 
     /* Init */
-    (void)XXH32_reset(&state32, XXHSUM32_DEFAULT_SEED);
-    (void)XXH64_reset(&state64, XXHSUM64_DEFAULT_SEED);
+    (void)XXH32_reset(&state32, g_default_seed_u32);
+    (void)XXH64_reset(&state64, g_default_seed_u64);
     (void)XXH3_128bits_reset(&state3);
```
Cyan4973: nit: this initialization is no longer necessary since it's overwritten by the following one.

Cyan4973 (approved):

cli/xxhsum.c
d06i: Are you referring only to the XXH3_128bits_reset() or all the reset calls? For state32 and state64, don't we need the reset calls to the initialize the seed values?

Cyan4973: It's about `state3`.
It's being initialized twice.
The first time is unnecessary since it's being overwritten by the second one.

sergeevabc:
Arrgghh! Please, roll out `--seed`.

d06i:
> Arrgghh! Please, roll out `--seed`.

You're right. I used "-s" since the parser was already in place but I can change it. Thank you.

Cyan4973 (approved):

sergeevabc:
Dear developers, there are no releases since December 2024. Maybe it's time to roll out a new version? I tried to find an artifact for Windows in the Actions, but failed to do so.

PR #1101, opened 2026-03-17
use AVX-512F setzero instead of set1_epi8

NexusXe:
For zero-initializing a 512-bit vector, it is better to use `_mm512_setzero_si512()`. This shouldn't change anything in compiled binaries, but could avoid issues under unique optimization scenarios.

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -5076,7 +5076,7 @@ XXH3_initCustomSecret_avx512(void* XXH_RESTRICT customSecret, xxh_u64 seed64)
     (void)(&XXH_writeLE64);
     {   int const nbRounds = XXH_SECRET_DEFAULT_SIZE / sizeof(__m512i);
         __m512i const seed_pos = _mm512_set1_epi64((xxh_i64)seed64);
-        __m512i const seed     = _mm512_mask_sub_epi64(seed_pos, 0xAA, _mm512_set1_epi8(0), seed_pos);
+        __m512i const seed     = _mm512_mask_sub_epi64(seed_pos, 0xAA, _mm512_setzero_si512(), seed_pos);
 
         const __m512i* const src  = (const __m512i*) ((const void*) XXH3_kSecret);
               __m512i* const dest = (      __m512i*) customSecret;


gzm55:
from intel docs: [_mm512_set1_epi8](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm512_set1_epi8&ig_expand=5807) and [_mm512_setzero_si512](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm512_setzero_si512&ig_expand=5871) are both AVX512F.

But the latency and throughput of _mm512_setzero_si512 are better.

NexusXe:
> from intel docs: [_mm512_set1_epi8](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm512_set1_epi8&ig_expand=5807) and [_mm512_setzero_si512](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm512_setzero_si512&ig_expand=5871) are both AVX512F.
> 
> But the latency and throughput of _mm512_setzero_si512 are better.

Note how it says "Sequence" - if only compiled targeting AVX-512F (which it is in this case) it may generate a functionally equivalent but strictly inferior sequence of instructions. `vpbroadcastb` (which `_mm512_set1_epi8` prefers to compile to) is [only available with AVX-512BW](https://www.felixcloutier.com/x86/vpbroadcastb:vpbroadcastw:vpbroadcastd:vpbroadcastq#:~:text=VPBROADCASTB%20zmm1%20%7Bk1,AVX512BW).

gzm55:
> > from intel docs: [_mm512_set1_epi8](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm512_set1_epi8&ig_expand=5807) and [_mm512_setzero_si512](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm512_setzero_si512&ig_expand=5871) are both AVX512F.
> > But the latency and throughput of _mm512_setzero_si512 are better.
> 
> Note how it says "Sequence" - if only compiled targeting AVX-512F (which it is in this case) it may generate a functionally equivalent but strictly inferior sequence of instructions. `vpbroadcastb` (which `_mm512_set1_epi8` prefers to compile to) is [only available with AVX-512BW](https://www.felixcloutier.com/x86/vpbroadcastb:vpbroadcastw:vpbroadcastd:vpbroadcastq#:~:text=VPBROADCASTB%20zmm1%20%7Bk1,AVX512BW).

a quick test https://godbolt.org/z/de4aEdjas

`_mm512_set1_epi8` as a function, it should generate 512F-only instructions when found no 512bw.

see functions in the above test:

```C
__attribute__((target("avx512f,no-avx512bw"), optimize("O0")))
__m512i test_f_set1_epi8_O0() {
    // NOTE: vpbroadcastb ymm0  (AVX2)
    return _mm512_set1_epi8(0xAB);
}
__attribute__((target("avx512f,avx512bw"), optimize("O0")))
__m512i test_bw_set1_epi8_O0() {
    // NOTE: vpbroadcastb zmm0  (512BW)
    return _mm512_set1_epi8(0xAB);
}
```

`vpbroadcastb ymm`  should be part of AVX2.

the previous [document](https://www.felixcloutier.com/x86/vpbroadcastb:vpbroadcastw:vpbroadcastd:vpbroadcastq#:~:text=VPBROADCASTB%20zmm1%20%7Bk1,AVX512BW)
only describe the broadcast _with mask_ version of vpbroadcastb.

in another document https://www.felixcloutier.com/x86/vpbroadcast
we can find `VPBROADCASTB ymm` without mask is from AVX2.

NexusXe:
> `vpbroadcastb ymm` should be part of AVX2.
> 
> the previous [document](https://www.felixcloutier.com/x86/vpbroadcastb:vpbroadcastw:vpbroadcastd:vpbroadcastq#:~:text=VPBROADCASTB%20zmm1%20%7Bk1,AVX512BW) only describe the broadcast _with mask_ version of vpbroadcastb.
> 
> in another document https://www.felixcloutier.com/x86/vpbroadcast we can find `VPBROADCASTB ymm` without mask is from AVX2.

But `vpbroadcastb zmm` is only available with AVX-512BW, which `_mm512_set1_epi8` is equivalent to. Either way, using the `setzero` intrinsic is the correct way to do it.

gzm55:
> 
> But `vpbroadcastb zmm` is only available with AVX-512BW, which `_mm512_set1_epi8` is equivalent to. Either way, using the `setzero` intrinsic is the correct way to do it.

Most likely, I think _mm512_setzero is exactly the right choice for performance. (better also post some perf test results)

Both intrinsics functions are correct in terms of semantics and the generated instruction sequences. The intrinsics functions is not mapped 1-to-1 instruction, but implement in different instruction sequence for different constrains.

In particular, for `_mm512_set1_epi8(0)`, it could be interpreted as `vpbroadcastb ymm (AVX2) --> vinserti64x4 zmm (512F)` or `vpxor xmm0, xmm0, xmm0 (AVX)` based on `-O?` and `-m???` constrains.

And _mm512_setzero_si512(), always generate ``vpxor xmm0, xmm0, xmm0` which is an AVX instruction, shorter, faster then `vpxor zmm0 (AVX512F)`.

Cyan4973 (approved):
It's unlikely to be measurable,
but that's still the right thing to do,
cleaner, clearer,
and yeah, slightly more optimized at micro-op level.

Issue #1064, opened 2025-08-26
XSUM_hashStream exits on file read errors, breaks multiple input processing


ZoomRmc:
https://github.com/Cyan4973/xxHash/blob/bab7e27f4c6ae4efbb83dd99ae8a554423571635/cli/xxhsum.c#L279-L281

`XSUM_hashStream` exits on `ferror` with the `Error: a failure occurred reading the input file.` message. 
This prevents processing multiple input files.

In my case, finding FS/device errors was exactly the point of running xxhsum. When processing files in batches and parsing the output, if the process exited prematurely, the whole batch was marked as failed which resulted in great number of false negatives.

Cyan4973:
It requires updating `XSUM_hashStream()` so that it can return an error. Right now, it can only return successfully, and therefore it `exit()` when it can't.

ZoomRmc:
Sure. 
If fixing this is not planned, the docs should mention that outputting any result for every input file is not guaranteed.

PR #1110, opened 2026-05-10
cli: don't exit on read error during streaming hash

eyupcanakman:
XSUM_hashStream called exit(1) on ferror. A read failure on any one input aborted the whole xxhsum run mid-batch, even though XSUM_hashFiles is already designed to continue past per-file errors. The error now flows back via an out-param, mapping to a new LineStatus_failedToRead status so xxhsum keeps processing remaining files. References #1064.

### DIFF
--- a/cli/xxhsum.c
+++ b/cli/xxhsum.c
@@ -240,11 +240,13 @@ typedef union {
  * XSUM_hashStream:
  * Reads data from `inFile`, generating an incremental hash of type hashType,
  * using `buffer` of size `blockSize` for temporary storage.
+ * On read failure, sets `*readError` to 1 if `readError` is non-NULL.
  */
 static Multihash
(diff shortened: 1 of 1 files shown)


Cyan4973:
I agree with the objective,
but first we need a test, that produces the problem and controls the fix.

_Edit_: test at #1130, it triggers an error signal, as intended

Cyan4973:
Thanks for the PR.
It's directionally correct, and works,
though I also believe that the "correct" fix must go deeper,
and change a few function signatures
that were previously designed as if they cannot fail (i.e. they `abort`, they do not error out)
and now must rather return an operation status.
This leads to a much different code, now written in fda0452c9df98439fc0e0a38d385d4c983608f67.

superseded by #1130

PR #1125, opened 2026-08-16
fix: include system headers outside extern "C"

### LINKED ISSUE #1122: <xxhash.h> includes system headers inside extern "C" block
Including system headers inside a linkage spec is a no-no.  It has caused
this error:

/usr/lib64/gcc/aarch64-suse-linux/17/include/arm_neon.h:5286:1: error:
conflicting declaration of ‘int64_t vaddd_s64(int64_t, int64_t)’ with ‘C’
linkage
 5286 | vaddd_s64 (int64_t __a, int64_t __b)
      | ^~~~~~~~~
/usr/lib64/gcc/aarch64-suse-linux/17/include/arm_neon.h:37:21: note: previous
declaration with ‘C++’ linkage
   37 | #pragma GCC aarch64 "arm_neon.h"
      |                     ^~~~~~~~~~~~

ameliabarnabyhub:
Fixes #1122

## Summary

- System headers must not be included inside an `extern "C"` region (can force C linkage onto C++ declarations, e.g. `arm_neon.h` on aarch64 GCC 17).
- Add `XXH_EXTERNC_PAUSE` / `XXH_EXTERNC_RESUME` helpers and wrap all system `#include`s so they are processed outside `extern "C"`.

## Test plan

- [ ] C++ compile that previously failed with `arm_neon.h` conflicting C vs C++ linkage now succeeds.
- [ ] Existing xxHash C and C++ builds / tests still pass.

### DIFF
--- a/xxhash.h
+++ b/xxhash.h
@@ -245,6 +245,15 @@
 extern "C" {
 #endif
 
+/* Temporarily leave extern "C" around system #includes (issue #1122). */
+#if defined(__cplusplus) && !defined(XXH_NO_EXTERNC_GUARD)
+#  define XXH_EXTERNC_PAUSE  } /* leave extern "C" for system headers */
+#  define XXH_EXTERNC_RESUME extern "C" {
+#else
+#  define XXH_EXTERNC_PAUSE
+#  define XXH_EXTERNC_RESUME
+#endif
+
 /* ****************************
  *  INLINE mode
  ******************************/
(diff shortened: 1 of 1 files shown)


Cyan4973:
We should first have a CI test that makes the problem visible.
Let me see what I can do.

_edit_ : see #1129

Cyan4973:
Thanks for the PR,
the proposed solution works, but it also introduces preprocessor complexity, detrimental to future maintenance.

Superseded by #1129, more specifically 959ca516f356d0255e7166d3f8fe049b0c532837, which fixes the same issue in a simpler way.

PR #1128, opened 2026-09-07
cli: make --quiet work no matter where it's placed

### LINKED ISSUE #1098: --quiet is not quiet if passed as last argument
Hi,

Just noticed some weird behavior. The _--quiet_ option is ignored if it is passed after the _--check_ option

Below I ran _--quiet -c checksums/chk.xxhash_ first
Output was as expected (quiet with no _OK_ lines

Then I ran _-c --quiet_ and it was not quiet

> user@fedora ~> xxhsum --version
xxhsum 0.8.3 by Yann Collet 
compiled as 64-bit x86 autoVec (AVX2 detected) little endian with GCC 15.1.1 20250521 (Red Hat 15.1.1-2) 
user@fedora ~> xxhsum --quiet -c checksums/chk.xxhash
^C
user@fedora ~> xxhsum -c checksums/chk.xxhash --quiet
_many lines ending with OK_

System is Fedora latest with xxhash from official repo

afonsojanu:
Fixes #1098.

The option-parsing loop in `xxhsum` stops looking for flags as soon as it hits the first filename argument, since `-c`/`--check` treats everything after it as a continuous list of filenames rather than more options mixed in. That means `xxhsum -c list.xxhash --quiet` swallows `--quiet` as if it were just another (nonexistent) filename, while `xxhsum --quiet -c list.xxhash` works as expected. Same root cause explains why other flags placed after a filename get ignored too, but `--quiet` is the one that's been reported.

This adds an earlier pass over `argv` that pulls every `--quiet` out before the main loop runs (stopping at an explicit `--`, since anything past that is meant to be a literal filename), so it takes effect regardless of position.

Verified by reproducing the exact case from the issue against unmodified `dev` first (trailing `--quiet` gets treated as a bogus filename and the check stays noisy), then confirming it's silent both ways with the fix applied. Added two lines to the existing `test-xxhsum-c` Makefile target covering `--quiet` on both sides of `-c`, and ran the full target clean.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -254,6 +254,9 @@ test-xxhsum-c: xxhsum
 	! ./xxhsum -q $(TEST_FILES) 2>&1 | grep Loading
 	# xxhsum does not display "Loading" message into stderr either
 	! ./xxhsum $(TEST_FILES) 2>&1 | grep Loading
+	# --quiet takes effect no matter where it's placed, not just before -c (#1098)
+	test -z "$$(./xxhsum --quiet -c .test.README.md.xxh < README.md)"
+	test -z "$$(./xxhsum -c .test.README.md.xxh --quiet < README.md)"
 	# Check that xxhsum do display filename that it failed to open.
 	LC_ALL=C ./xxhsum nonexistent 2>&1 | grep "Error: Could not open 'nonexistent'"
 	# xxhsum to/from file, shell redirection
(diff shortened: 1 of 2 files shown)


Cyan4973:
Thanks, 
I think that's a real issue,
and the tests provided are very good.

I also believe it's an opportunity to revamp the CLI argument parsing,
so that it allows any argument in any order,
hence a new parsing code is proposed in #1131 .

Superseded by #1131.

PR #1130, opened 2026-09-14
test: reproduce read error aborting file list

### LINKED ISSUE #1064: XSUM_hashStream exits on file read errors, breaks multiple input processing

Cyan4973:
test and fix #1064 

1st commit fails (by design) before the fix,
then green after the fix


ZoomRmc:
Thanks a lot!

